PR #47, opened 2016-11-29
Generate protobufs.

gclitheroe:
I've flip-flopped on how and where to do some of this more than I would like.

This commit adds protobuf definitions (just enough for drawing maps at the moment).  There are test additions for creating protobuf files from the CSV in `networks`.  On test success protobuf files are pushed to S3 (by Travis).  My plan would then be to generate GeoJSON, KML, etc in the `haz/geonet-rest` code (so it ends up in api.geonet.org.nz).

Any opinions or suggestions?

There is always a chance I've missconfiged the S3 deploy (which won't show up until this is merged to master).  If so I'll fix it.


quiffman:
S3 and a bucket for protobufs seems like a fairly legit option, for occasionally updated data to me.

@ozym do you have any thoughts you'd like to share on this?

ozym:
Looks good, I see this as the way forward for other outputs I.e. stationxml. If it's in the one repo you don't need pipelining.

The deploy script is just a placeholder at the moment.

PR #50, opened 2016-12-04
updated tests

ozym:
The switch to go 1.7 allows using the Run mechanism for tests. I've also made some of the flows a little more easier to follow (i.e. swapped switches for if continue blocks).


ozym:
May revisit this a little further down the track.

PR #51, opened 2016-12-05
Consistent receive manufacturer.

gclitheroe:
Make the receiver manufacturer consistent between receviers.csv
and firware.csv  Then is can be used as part of a key.


ozym:
Good call, I'll update the oracle join for the regular updates as well.

PR #58, opened 2016-12-08
Add outputs for Site / Sensor type services.

gclitheroe:
Adds outputs for seismic (and other sensor types) that are a Sensor at a Location at a Site at a Station.

It is a little of a departure from the way we may think of maps of the network now, this is more like "a network of sensors by type".  If this tests out ok with users then I would also plan to break marks.geojson into gnss.geojson and metsensor.geojson.

For geojson I've only output permanent Stations in the NZ network the are not restricted (this is easy to adjust).  The new geojson files are classified by Sensor type.  There are a small number of old Sites where there is no SensorType in `resp/auto.go` so they don't make it into a geojson file.

File sizes at the moment (stations.pb has all Stations, Sites, and Installed Sensors).  If `stations.json` is to large in a browser I might also need to do a json file per station.

```
157K Dec  8 20:56 accelerometer.geojson
5.0K Dec  8 20:56 barometer.geojson
36K Dec  8 20:56 broadbandseismometer.geojson
41K Dec  8 20:56 buildingarraysensor.geojson
958 Dec  8 20:56 hydrophone.geojson
56K Dec  8 20:55 marks.geojson
447K Dec  8 20:55 marks.json
181K Dec  8 20:55 marks.pb
3.7K Dec  8 20:56 microphone.geojson
16K Dec  8 20:56 pressuresensor.geojson
5.1K Dec  8 20:56 shortperiodboreholeseismometer.geojson
118K Dec  8 20:56 shortperiodseismometer.geojson
1.7M Dec  8 20:56 stations.json
729K Dec  8 20:56 stations.pb
111K Dec  8 20:56 strongmotionsensor.geojson
```


ozym:
Only comment is the "SENSOR" being appended is a hang over from the current DB. We can fix this up once we're no longer syncing, or I can adjust the results of the query to remove the SENSOR suffix.

PR #65, opened 2016-12-17
Update stationxml and response building

ozym:


ozym:
Trying to simplify the stationxml code and pushing some helpers into the resp package.

PR #67, opened 2016-12-21
Generate config as a protobuf for shakenz

gclitheroe:


tests/shakenz_test.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	if err := os.MkdirAll(apiDir + "/config", 0777); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := ioutil.WriteFile(apiDir + "/config/shakenz-slink.pb", b, 0644); err != nil {
+		t.Fatal(err)
+	}
+}
```
crleblanc: That's a pretty long and deeply nested test but I can see why.  The comments make it easy to see the intent and I don't see much point in splitting it up into different functions since it's only used here.

crleblanc:
After discussion with @gclitheroe we'll merge this PR and try to fix the dependency issues later.

PR #71, opened 2017-01-08
proof of concept for fdsnws-station service

ozym:
***Not to be merged***

@gclitheroe here's a quick look at a fdsnws-station service ... I've only done a couple of the parameters and haven't put the endpoints where they should be, or anything serious regarding a webservice.

It basically loads a stationxml file into memory, and then when a request comes in it works through the structure building a results copy ....

The alternative (traditional) approach would be to load the xml into a DB schema and make the queries against that. 

Downsides with this approach is the start-up cost (loading the stationxml) and the mechanism for updating  the contents is to stop/start. The benefit would be not having to have a database back-end.  Not sure about speed though, the slow part is the scanning and the use of wildcards. Pretty sure there would be a better technique to load the parts (i.e. Network, Station...) into maps and then look them up rather than walking the tree ....

I've also played with reading this directly from the delta repo (rather than stationxml) which works as well. Although I think this would be better as a protobuf with the final conversion to xml happening just prior to returning the results.

The stationxml tool in the repo can build a stationxml file which can be used to test this service (134Mbytes at the moment).

### DIFF
--- a/tools/fdsnws-station/main.go
+++ b/tools/fdsnws-station/main.go
@@ -1,0 +1,228 @@
+package main
+
+import (
+	"encoding/xml"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/ozym/fdsn/stationxml"
+)
+
+func matchString(str string, patterns []string) bool {
+	if !(len(patterns) > 0) {
+		return true
+	}
+	for _, pattern := range patterns {
+		if hit, err := filepath.Match(pattern, str); err == nil && hit {
+			return true
+		}
+	}
+
+	return false
+}
+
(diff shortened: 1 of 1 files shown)


gclitheroe:
Hi Mark,

thanks for this, I finally had a chance to look through it.  I was able to build and run the stationxml tool (I just had to `go get github.com/hashicorp/go-memdb`) and output an XML file (~175MB).  

@junghao is going to have a look at implementing the FDSN WS station service, using the station xml as the data source.

It also seems like we should use the FDSN webservice as a data source for configuring shakenz-slink  unless there is some additional information that is missing from the station xml for this purpose?  Then move FDSN along to some appropriate high tier level.

It makes the code I added to this repo to produce protobufs surplus and I would suggest we then delete that code (and the protobuf artifacts).

Probably this repo should produce a small number of well defined formats (well defined XML is fine) that are well tested and versioned.  

Will let you know how we get on with and FDSN station service.

/cc @crleblanc @quiffman

PR #73, opened 2017-02-24
db update

ozym:
@elidana Only query I have is the name of the Mark

> Lake Tennyson temporary

Both its name and the capitalisation.


elidana:
post-Kaikoura cgps site names now capitalized on Delta Oracle db

PR #75, opened 2017-03-01
update from db

ozym:


ozym:
Good idea to add another field review, they will match what's in delta (db) but another set of eyes can't be bad.

PR #80, opened 2017-04-07
add legacy xml config generation

ozym:
Start moving some config generation.


quiffman:
@ozym `+18,078 −4,861` is a scarily large PR to review.

Is there any way you could break this down?

ozym:
I'll see if the first PR can work on its own.

PR #81, opened 2017-04-11
metadb and unvendor go-memdb

ozym:
This should hopefully make it a little less convoluted for tools recovering data.

Some of the tools will be broken until the next pull release goes through.


ozym:
@quiffman have added the missing tidal helper functions. This will mean the tools compile correctly.

PR #94, opened 2017-05-24
update from db

ozym:


elidana:
@ozym,  I had a better look at the database. That "Mark - Description" field is not imported at the moment (for what I can see from the code and the tables).
FWIW, the content of this column might be divided in subcategories, that will need to be redistributed into different tables of the new delta, such as
"Mark name change" : we probably don't need to keep this info (to check with LINZ)
"enabling of full-GNSS": this will fit into install/sessions.csv table. 
"antenna mount description" : will go in the network/monuments.csv table
"events possible affecting the time series": not sure where the best place is for that. Under marks??? 

All of the above can be done with a "one off" manual pr/review once we switch to the new delta

peanutty:
looks good, you'll get another one for changes made at PYZ in a bit :)

PR #95, opened 2017-05-24
added notes to mark table

ozym:
@elidana Here's an example of added in "Notes" field to Marks.

The problem is that the description field entry has been used for lots of different types of information. e.g. site details, access details.

This might be something we can discuss, i.e. what types of information goes here (for reference to the data collected and its processing, not so much operational details).

I don't intend to merge this but it wouldn't be hard when we have a plan.

You'll need to expand the diff as it changes each line.


ozym:
@elidana just saw your comments on the other pull request. I agree in that these could be updated after we switch.

ozym:
Will wait until we see what happens with #97

PR #96, opened 2017-06-01
update from db

ozym:


ozym:
@elidana fair point. It looks like I'm using the Mark dates rather than the Monument dates for the extraction (most likely unintended). I'll make a new pull request following this one.

PR #97, opened 2017-06-01
now use monument dates rather than associated mark dates

ozym:
@elidana I have taken the monument start / end times from the db. I've also added a new set of tests to make sure a monument exists before a mark is placed, and to check if a mark exists after a monument has been removed.

This pull request will fail due to:

```
mark continues after monument removed CLSG: mark=9999-01-01 00:00:00 +0000 UTC monument=2011-06-21 00:00:00 +0000 UTC
mark established before monument LYTT: mark=1999-11-15 00:00:00 +0000 UTC monument=2000-03-13 00:00:00 +0000 UTC
mark established before monument PAWA: mark=2005-03-23 00:00:00 +0000 UTC monument=2005-05-23 00:00:00 +0000 UTC
mark established before monument RDLV: mark=2009-01-13 00:00:00 +0000 UTC monument=2009-01-14 00:00:00 +0000 UTC
mark continues after monument removed RGCR: mark=9999-01-01 00:00:00 +0000 UTC monument=2011-08-29 00:00:00 +0000 UTC
```

 Which is part of the checking ...

These will need to be fixed in the oracle database and then I will update the PR.


elidana:
thanks heaps @ozym !

I'll need to spend some time to review why there are so many differences, to try to keep these dates correct.  I will start with the 5 sites that are not passing the check.

elidana:
Reply from Paula Gentle, LINZ:

> Good question, and I don’t think we have a clear answer.  I think in terms of your records they will be the same (fro the most part).  In terms of the LINZ records our monuments also include things like beacons which may be installed and removed at different points in time and it is useful to know if it is possible to observe to it or how old the beacon is (which may indicate the likely hood if is still standing and observable).
> 
> In terms of a CORS site a monument can be a pillar or braced monument which could have more than one mark on it, for example the warkworth monument has two marks (WARK and the vertical offset mark WAWH) – where WARH was installed sometime after the monument (confusing), while WARK was installed at the same time.  We don’t currently identify WARH as having a monument though.
> 
> LINZ splits our mark and site information into three different groups,  mark, beacon and protection structure.   Mark is the physical point that has a position/coordinate.
> 
> Hope this helps, it is a philosophical question and will likely need further discussion. 
>

elidana:
I also had a chat with Neville, GNS Science. 
he also thinks that there's no much sense to have marks and monuments distinguished, for a continuous GNSS site. In case a mark is destroyed , the site will be considered a "new one" anyway.

ozym:
My only thought was that the two tables have different (but related) information. They could be merged into one file if that make readability easier, or kept as is but with the Date/Time dropped from the monument file.

Currently

    Mark,Network,Name,Latitude,Longitude,Elevation,Datum,Start Date,End Date

and

    Mark,Domes Number,Mark Type,Type,Ground Relationship,Foundation Type,Foundation Depth,Start Date,End Date,Bedrock,Geology

The alternative is a much wider file ...

    Mark,Network,Name,Latitude,Longitude,Elevation,Datum,Domes Number,Mark Type,Type,Ground Relationship,Foundation Type,Foundation Depth,Bedrock,Geology

And perhaps a Notes field to add as well ....

elidana:
ok, I see the "trick" here.

the `marks.csv` table in reality is a table that mainly contains info about the location (`Site Location` in GNSS-metadata terminology)
the `monuments.csv` table contains info about how this site is built (`Site Identification`)
So it's good to keep the two tables separated. 

In the GeodesyML, the date is only associated with the Site Identification  (so the `monuments.csv` file), and is the date the monument/mark was built. 
For our internal logic, the date in `marks.csv` is also used to figure out if a site is operational or not. Is that correct? 
You can have a site that is not operational, but the monument hasn't been removed, so the date would still be 9999-..., but the site would be considered as "closed" for our systems.

I'm going around in loops here but I think I'll do is review all the dates (particularly if there are differences in marks and monuments' start-date, and why). 
Then , depending on how that date info in marks.csv is used by dependent systems, we can decide how to handle it (I'd prefer to avoid unnecessary duplicates)

PR #102, opened 2017-06-27
update from db

ozym:


install/firmware.csv
```diff
@@ -316,7 +321,7 @@ Trimble Navigation Ltd.,TRIMBLE NETR9,5035K69877,4.17,2011-01-14T00:00:00Z,2012-
 Trimble Navigation Ltd.,TRIMBLE NETR9,5035K69877,4.46,2012-04-10T20:31:00Z,2016-03-29T22:40:32Z,
 Trimble Navigation Ltd.,TRIMBLE NETR9,5035K69877,5.10,2016-03-29T22:40:33Z,2016-08-16T04:10:00Z,
 Trimble Navigation Ltd.,TRIMBLE NETR9,5035K69877,5.14,2016-08-16T04:10:01Z,2016-12-20T03:32:06Z,
-Trimble Navigation Ltd.,TRIMBLE NETR9,5035K69877,5.15,2016-12-20T03:32:07Z,9999-01-01T00:00:00Z,
+Trimble Navigation Ltd.,TRIMBLE NETR9,5035K69877,5.15,2016-12-20T03:32:07Z,2017-06-26T03:02:24Z,
```
elidana: this is a wrong entry in the DB I think, I need to check that one

elidana:
@ozym , that entry for PAKI in the install/firmware.csv is wrong (that is an error in the Oracle-DB I think, will check that now). 
It made me think of a further check/hook we can add, if that's not already existing.
Is there anything that checks that the _current firmware end-date_ on an operational site is 9999-01-01T00:00:00Z ?

If not, I can open a ticket on the delta-db as a further implementation.

install/firmware.csv
elidana: now fixed on db

ozym:
@elidana I've added a test and pushed it to this PR. It should cause an error, if this looks right I'll reload from the DB and hopefully it should pass.

PR #110, opened 2017-07-13
add trillium 360 resp

ozym:
@nbalfour just adding the response information prior to these being deployed. I haven't made the full responses as it will depend on the datalogger and/or settings.


ozym:
moved to #121

PR #114, opened 2017-07-25
added extra stationxml constraints for logger/sensor and operational state

ozym:
@quiffman starting to look at tailoring the stationxml extracts, a couple of constraints from the old system were missing, these were being specific about loggers and sensors to be extracted. I've also added for real-time related work whether the site is currently, or recently, operational.

e.g. this is the original mapping for the IRIS dataless SEED files.

```
$pod->streams(
        network_remap => {'CH' => 'NZ'},
        networks => ['NZ', 'CH'],
        stations => ['KHZ', 'QRZ', 'OUZ', 'HIZ', 'BKZ', 'ODZ', 'BFZ', 'CTZ', 'URZ', 'RPZ', 'WPVZ'],
        sensors => ['STS-2', 'CMG-3TB', 'CMG-40T-60S', 'FBA-ES-T'],
        loggers => ['Q330HR/6', 'Q4120/6', 'Q330/3'],
        channels => ['HHZ', 'HHN', 'HHE', 'HH1', 'HH2', 'LHZ', 'LHN', 'LHE', 'LH1', 'LH2', 'VHZ', 'VHN', 'VHE', 'VH1', 'VH2', 'HNZ', 'HNN', 'HNE', 'BNZ', 'BNN', 'BNE', 'LNZ', 'LNN',
'LNE']

```

I've included these in the `build.sh` script as an example at this stage, together with an operational sc3 output. I'm also building the pod headers but have left the next step to another outside process for now.

The only thing at the back of my mind is that for times when the filters are used should stations with no channels (after filtering) be ignored.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,5 +1,7 @@
 #!/bin/bash -x
 
+set -x
+
 errcount=0
 
 error_handler () {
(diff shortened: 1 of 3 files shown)


quiffman:
Aside from what I think is a small oopsie with the channel regex's, it looks good @ozym.

I'm assuming:
> I've also added for real-time related work whether the site is currently, or recently, operational.

is so we can say, include operational stations for the last month-year, in `scp.xml` so they're available for review, for example...?

ozym:
@quiffman have updated the typos, yes the operational flag is so that we can have a real-time "going" set of stations (for e.g. sc3). There's the option to define what "recent" is so it will handle the within the last month etc.

PR #116, opened 2017-07-31
rework build script to generate pod tar files

ozym:
@quiffman turns out the aws sync of the pod directories (large number of small files) was taking too long for travis and may not have been a good idea in any case.

I've made the build script now generate compressed tar files which should avoid the problem. This will also make checking for changes much easier for the downstream processors as well.


quiffman:
Ah, yeah. That's not ideal. ~22,000 files there!

PR #119, opened 2017-08-10
add trillium compact response

ozym:
@nbalfour adding some extra response information for deployed portables.


ozym:
It was quite painful to do the updates and the checks. I'm thinking of stepping back a bit and finding a more easier way to manage the responses.

ozym:
moved to #121

PR #121, opened 2017-08-12
flatten response files and add trillium compact response

ozym:
@nbalfour this is basically the other pull requests for the updated trillium details plus a change in configuration file layouts to make it easier to follow what is changing and where.

Changes will also need to be made once the trillium models have been updated.


ozym:
Was thinking about tests too. One approach I thought of for sensors was to assign a frequency band over which the response should be flat (depending on type of sensor) then the poles/zeros can be compared with this assumption. May be useful to find major errors, e.g missing pole or zero or wrong units.

PR #130, opened 2017-08-29
comment out resourceid

ozym:
An alternative to filling in resources ids.


ozym:
still needs some thinking as the schema has some that look like they are not optional.

PR #131, opened 2017-08-31
update from db

ozym:
@dc735 changes from db updated as well as network changes.

~~@elidana my only comment was that I recall we were using the IGS? model descriptions for GNSS equipment, i.e. TRM5971.00 etc.  Do the new "Zephyr 3 ..." match this?~~


dc735:
@ozym Does this help answer your question about IGS model descriptions
https://github.com/GeoNet/gps-scripts/blob/master/netr9/config_files/antenna_radome_pairs.txt

PR #143, opened 2017-09-19
Add download and distribution messages to Gloria protobuf

gclitheroe:
* Adds Download (rate and priority) and Distribution messages to the Gloria protobufs.  
* Download order will be IGS, LINZ, everything else.
* I haven't set a download rate yet - it will be left unrestricted.  This can be set per site.  The rate is controlled here https://github.com/GeoNet/gloria/blob/master/internal/netr9-download/netr9_download.go#L60


elidana:
nice!!! FWIW, when I was highrate downloading data during the Kaikoura response, we agreed to have a rate limit set to 25kb/s

PR #155, opened 2017-10-08
update tests to use t.Run

ozym:
@dc735 an update to the testing code. There's not much new here, rather a switch to using the `t.Run` mechanism that became available in `go 1.7`. 
It basically allows some form of isolation between tests (one part can have a fatal result and stop but other parts will continue to be tested if possible).

I've also tried to make them a little more readable, e.g. being explicit about lists of 

> `if ( something ) { continue }` 

rather than use a switch which is more compact but isn't as obvious what it's doing.

I've also brought all the `csv` file loading parts into a single function which is called with the test variable so that it can have a fatal error.

There are still a couple of tests that only make a warning message (need `go test -v` to see them), I'll think of a way to bring these online but it may need some adjustments to the db or to the extraction code - mostly related to station/site stop/start times not matching when sensors are installed.

### DIFF
--- a/tests/antennas_test.go
+++ b/tests/antennas_test.go
@@ -1,7 +1,6 @@
 package delta_test
 
 import (
-	"sort"
 	"testing"
 
 	"github.com/GeoNet/delta/meta"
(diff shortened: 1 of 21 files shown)


ozym:
@dc735 any chance you can have a look at this pull request?

PR #164, opened 2017-10-25
site classes

ozym:
Just a proof of concept.


nbalfour:
Are these based on both the concise and extensive site class tables? It looks that way.

ozym:
This was some early work I did on the concise file. It turns out that this is pretty much all vs30 information, the actual class information is in the extensive files. However the extensive files have very little vs30 information in them.

I'll update this with the new set of files (in a geology directory). I'm not anticipating this to be merged.

nbalfour:
After also looking at https://github.com/GeoNet/delta/pull/165 I have been thinking. 
It would be great to combine the geology and sky information and call this the environment folder perhaps?
We could add the geology information from the mark table to the geology table here. Then we could add a sky visibility table, which will need to be cleaned up eventually. 
Let me know what you think?

ozym:
@nbalfour That's a good idea, I'll move these to environment.  Create a new file "visibility.csv" (or something), and change the lookup from "bedrock.csv" etc to use "Code" rather than Station so that we can match either a Mark or a Seismic site. 

We can merge the bedrock/geology info from the monuments file in after we've setup these files (merging from delta and csv is a bit more work at the moment until we turn one off).

ozym:
@nbalfour the column labels are what comes out of the raw csv files (usually a join of a couple of cells).

ozym:
The tests are failing because the subsurface files reference a site "RCS2" which doesn't exist in the rest of delta. I suspect this was simply renamed ...

ozym:
@nbalfour still just a proof of concept

nbalfour:
I talked to Anna K. on Friday and she suggested merging a few of the tables. There is also a lot of cleaning and review that should be done because this is a mix of her concise spreadsheet and Jim Cousins extensive one. 

There might also need to be a bit of a description about some of the quality parameters as there is not community standard for quality.

I think we might have to discuss this in person as it's a bit tricky.

ozym:
Sure thing. I also think we can make a more complete file as a product (essentially a join of the various tables) which can be tailored. This is to avoid storing large sparse files.

salichon:
@JonoHanson  FYI  I think It s blocked 
I just updated the branch to maintain consistency

PR #165, opened 2017-10-25
add sky visibility environment file

ozym:
Proof of concept based to add extra fields to the marks table.


nbalfour:
As I mentioned in https://github.com/GeoNet/delta/pull/164 I think sky visibility would be good to have in the geology/environment folder. 

After discussions with @elidana, it seems like the other parameters (protection, plaque and plate) are not really used for anything (as we discussed). We are required to communicate to LINZ what they are so that they can put the information in their database. We are not required as part of the LINZ specification to keep this information. In my mind, this is site information that is not required to be in the delta repo.

ozym:
@nbalfour I've added a visibility file, a useful side effect is that we can add a time window so that if the visibility changes over time we can track this (i.e. a building is added or removed etc.).

elidana:
thanks @nbalfour and @ozym !!
The "side effect" of having a time window on the sky visibility info is quite useful to keep track of "changes  affecting the data"

PR #171, opened 2017-10-31
update and add readme files

ozym:
@nbalfour @quiffman a first go at bringing the base README up to date and adding new ones in the environment and install directories.

This is still pretty much a skeleton with a few formatting inconsistencies still in there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,134 +2,50 @@
 
 Meta information for the GeoNet sensor network.
 
-## changes
-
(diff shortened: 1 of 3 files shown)


ozym:
one thought was to have a `docs` directory rather than placing the README files near the csv files. They could then be named after the directory.

elidana:
this new version of the readme files is so useful, thanks @ozym !

nbalfour:
This looks good thanks @ozym 
I don't have very strong thoughts about a docs folder but my initial reaction is to keep the README files near the csv files for now. Once people start using the repo for the change control of metadata there might be some good reasons to change things.

PR #172, opened 2017-10-31
move tidal files into environment

ozym:
I think the tidal parameters (as provided by LINZ) better fit the new `environment` grouping than the existing `network` location.


ozym:
@nbalfour now we can have sky, sea, and land.

PR #181, opened 2017-11-15
use session times based on deployment times

ozym:
Related to complexities in the oracle sessions for CHAT. Can now use the installed times to build the session start/end times and cleanup the repo to fix the CHAT issues.


ozym:
That's how I found CHAT, as it was the only existing one changed.

elidana:
metadata now fixed in https://github.com/GeoNet/delta/pull/182

elidana:
bump! :-) (pleeease!)

quiffman:
@ozym - this looks to me like code trying to clean up mis-aligned meta-data from Oracle. Is that a fair assessment?

It seems to be trying to pick the earliest/latest start/stop data of session receiver or antenna.
Where, presumably the session should win-out, right?

ozym:
Yup, the code  being uncommented was how it was supposed to work. However, we had to use the original non-overlapping sessions files so that we could do like-for-like comparison with the oracle derived results. This is no longer a requirement so we should revert back to the intended mechanism.

elidana:
thanks to both!

elidana:
awesome that this is now merged!!! Thanks both!
The code is now working as expected and sessions are correctly split in the xml output
...but I've found another (smaller) "bug" unfortunately. it's reported in https://github.com/GeoNet/tickets/issues/1705

PR #198, opened 2017-12-07
added go fmt and go vet tests

ozym:
seems like a good idea.

mostly fixes `"uses unkeyed fields"` vetting warnings and some fmt changes in sit and gloria pb code.

am not  using the "-s" tag to gofmt, will add this once I update my vim macros.

you have to jump through hoops to avoid ./vendor. currently the yaml code isn't passing go vet.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -10,10 +10,10 @@ error_handler () {
 trap error_handler ERR
 
 echo "go formatting" 1>&2
-test -z "$(find ./meta ./resp ./tests ./tools -name "*.go" -exec gofmt -l {} \; | tee /dev/stderr)"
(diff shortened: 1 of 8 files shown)


test.sh
```diff
@@ -9,6 +9,13 @@ error_handler () {
 
 trap error_handler ERR
 
+echo "go formatting" 1>&2
+test -z "$(find ./meta ./resp ./internal ./tests ./tools -name "*.go" -exec gofmt -l {} \; | tee /dev/stderr)"
+
+echo "go vetting" 1>&2
+go vet ./meta ./resp ./internal/... ./tests ./tools/...
```
quiffman: Ah, I see. `go vet ./...` behaves differently between `1.8` and `1.9`.

Is there any reason we need to keep this project using `1.7` rather than bumping it to `1.9`?

PR #212, opened 2018-01-15
Add location for sit equipments.

junghao:


ozym:
@junghao I have merged in the go vet / go fmt #201 so this may have some impact on this PR.

PR #213, opened 2018-01-17
add internal network regexp for stationxml

ozym:
@quiffman it looked like a lot of the stationxml extraction problems have been related to stations being included when they shouldn't be.

To help this I have moved the `stationxml` tool option `networks` to now match the "internal" network code and I've added a new option `external` (and related `external-list`) to match the external code (which at the moment is only "NZ").

A set of network codes has been added to the `build.sh` script to exclude the obvious candidates.

/cc @nbalfour @salichon


salichon:
Hello There!
A suggestion, a thought.
Should we consider some filters regarding the location codes, for instance 10,11,20,21 (for production)?
Because i m thinking about the existing and future small arrays, building arrays with a collection  of 22..25...  for instance.

ozym:
@salichon building arrays are in the network "SB" which will not be loaded after this pull request is merged.  However, there is an option "locations" which will do just that and it could be added as you suggest. The tricky bit is remembering to add a new location to the list when/if a new one is used.

salichon:
@ozym okay thanks - i went through the networks.csv, now it s getting clearer amongst the "local network". I will need one day to get a short story about all these and their historical difference :)

@nbalfour in this incoming version os station xml for seiscomp, the XX temporary stations won't be included in the scp volume. It s okay for me and specially while no reponse time.
 
Meanwhile, should we envision, at least discussing an "emergency response" station xml build option (either XX or else DX "temporary deployment" ...)? 
NB: theoretically, such an inventory could even be ran for manual reviews without  having to update immediatly  the whole automatic  system (using its "simple" XML version).

quiffman:
@salichon think of these "internal" networks as an attempt at logical sub-grouping, but kind of a force-fit with the existing tech at the time.

@ozym I can't really see any other way around this currently... but the long regex doesn't exactly give me warm fuzzies. Could we use a negated form, instead? Something like `(^SB|.X)`?

ozym:
@quiffman do you mean I rewrite the check so that it expects a negative match, i.e. skip these networks?

quiffman:
@ozym yeah, that one. i.e. a blacklist, rather than a whitelist.

I feel like that should be shorter and require less frequent changes. Does that make logic?

ozym:
I think I will use a convention of a "!" as a prefix to indicate that the regexp should not match, the tricky bit is the list from a file (which is a bunch joined together), but the solution here is to drop that functionality and simplify the code as it isn't being used.

PR #217, opened 2018-01-18
sensor lennartz MKIII lite set up

salichon:
NB: PAZ to be confirmed
and/or checked for the MKIIlite


salichon:
@ozym I don t understand the conflict  since the resp/auto.go file was comitted with the changes.

salichon:
@nbalfour wrt the PAZ for the LE-3Dlite isntruments 
Shall we correct the Poles and Zeroes of all  the L3Dlite with regards to that updated document ?  
 http://www.lennartz-electronic.de/index.php/downloads-en/category/1-manuals?download=5:seismometers-manual

ozym:
The problem is related to gofmt with the "-v" flag. I've fixed the conflict but need to see why it was there in the first place.

salichon:
Sweet thanks

nbalfour:
@ozym yes, i think we should correct the responses for the other sensors.

ozym:
@salichon I suspect it may be related to your version of the go compiler when building the auto.go file.

salichon:
okay Merci Mark , 
which version is recommended ? 
I  have a 1.7 linux/amd64 on the laptop

salichon:
@nbalfour @ozym @quiffman 
Good day ! 
 i m going to merge that branch 
though, double checking on a long WE tuesday -> "Is it safe?"
In addition:  
Which go version + config Should I have to prevent from formating errors (see mark s commits) ? 
Thanks,
Cheers
Jerome

ozym:
@salichon in the `.travis.yml` there is an entry saying what version of go will be used to do the testing and building.

salichon:
FYI I installed go1.9.2.linux-amd64.tar.gz

As i realized this might be relevant to run go from a docker container perhaps. ..

PR #218, opened 2018-01-22
GDS1, MUL1 undeployment

staylorofford:


staylorofford:
Changes have been made. How does it look now?

PR #219, opened 2018-01-22
fix empty auto.go file issue

ozym:
@quiffman sometimes the "go generate" goes wrong (e.g yaml file or coding error) and in that case it currently drops an empty `auto.go` file. The next time `go generate` is run the `go` program baulks at the empty file.

The change is to simply check if the newly generated auto.go file has content, if not it will be removed.


resp/response.go
```diff
@@ -1,6 +1,6 @@
 package resp
 
-//go:generate bash -c "go run generate/*.go | gofmt -s > auto.go"
+//go:generate bash -c "go run generate/*.go | gofmt -s > auto.go; test -s auto.go || rm auto.go"
```
quiffman: This sounds picky, but rather than testing after the fact and removing, why not simply remove before the fact or just set bash to clobber the file, instead?

I suspect that will ultimately be easier to understand for a wider audience.

quiffman: Ah, yeah, ignore that - I just figured out why :+1:

PR #220, opened 2018-01-22
updated lennartz 3DLite poles/zeros

ozym:
@nbalfour @salichon I've consolidated the 3DLite Poles/Zeros into one entry that is consistent with the latest documentation. The auto.go file reflects these changes, mostly 4.444 to 4.44 ...

/cc @quiffman


salichon:
Thaks @ozym 
Does this mean that there only one generic name in the  delta instrument reposne files now 
LE-3Dlite (no more MKII or MKIII), doesn't it ?

ozym:
Only for the PAZ lookup.

salichon:
Okay, Thanks!

PR #227, opened 2018-01-28
updated etna-2 resp

ozym:
@quiffman @nbalfour 

I've made a bare bones update to the response files to build meta data for Etna2 installations.

The  FIR filter information I've taken from (a) the Etna2 Manual and the IRIS NRL library. The PAZ and other details are taken from the Manual and compared with the IRIS NRL library for consistency.

We can review and work through this together.


nbalfour:
Thanks @ozym. I think it would be a good idea to work through this together. A bit later in the week if it can wait?

ozym:
@nbalfour yup, that's what I thought too.

salichon:
thanks, just refreshing my mind about a few tickets

PR #228, opened 2018-01-29
Updated README for recorder fields.

ozym:
@salichon good spotting. There are a few inconsistencies related to naming conventions which also need to updated at some point as well (e.g. Start/Stop vs Start Date, End Date etc).


salichon:
Sweet, right for the  naming conventions - even easier to interpret.

PR #238, opened 2018-02-07
fixed depth gauge 1hz response typo

ozym:
@quiffman @nbalfour while tracking down some missing traces (which turned out to be an unrelated problem) I noticed a typo in one of the 1Hz FIR filters being applied to the tsunami gauges (was a 10Hz version rather than the expected 1Hz).


ozym:
Have added a simple sample rate check ... found:

```
go generate
2018/02/08 13:22:19 invalid sample rate "Quanterra Dataloggers Connected to Fresh Water Depth Sensors" [LT]: found 10, expected 1
exit status 1

```

PR #240, opened 2018-02-14
WFES strong motion: time fix and station closure

salichon:
WFES closure and housekeeping


salichon:
FYI @ozym

salichon:
Thanks, therefore i ll proceed too merge

PR #248, opened 2018-02-23
Update streams.csv WTES

ozym:
Switch WTES to Axial layout (i.e. Z12 rather than ZNE).


salichon:
I ll proceed to the merging 
thanks

PR #253, opened 2018-03-13
Update firmware to 5.22 on HANA, TEMA NetR9s.

staylorofford:


elidana:
I think that based on that: https://github.com/GeoNet/tickets/blob/master/change_request_readme.md
you need to add a reviewer from the Network Team too...

PR #203, opened 2017-12-14
DRZ seismic station Closure 

salichon:
DRZ station Endtime changed to last known recorded data time
https://github.com/GeoNet/tickets/issues/1735
-> station and site are closed

PR #254, opened 2018-03-16
Closing DRZ dome shelter datalogger

salichon:
While closing DRZ, its datalogger Q330/3 528
wasn't. 
See (https://github.com/GeoNet/delta/pull/203)


salichon:
Thanks @samgeonet 
I ll proceed

PR #259, opened 2018-03-22
PXZ.20 to PWZ.20

salichon:
PXZ 20 strong motion Slate replacement  with Obsidian, lead to 
a new site that is more relevant to be named PWZ 20
See ticket for discussion related to the name


network/sites.csv
```diff
@@ -1510,7 +1510,7 @@ PUZ,20,-38.071547867,178.257209049,430,WGS84,Internal GPS Clock,2003-08-20T00:00
 PVCS,20,-41.22474608,174.87391737,0,WGS84,Unknown,2011-07-01T00:00:10Z,9999-01-01T00:00:00Z
 PWES,20,-41.12746,174.82586,9999,WGS84,External GPS Device,2012-06-28T23:00:07Z,9999-01-01T00:00:00Z
 PWZ,10,-40.029684442,176.861995223,66,WGS84,Internal GPS Clock,2001-01-30T00:23:00Z,2005-08-15T06:00:00Z
-PWZ,20,-40.029684442,176.861995223,66,WGS84,Internal GPS Clock,2001-03-21T00:00:00Z,2005-08-15T06:00:00Z
+PWZ,20,-40.029679,176.862007,59,WGS84,Internal GPS Clock,2001-03-21T00:00:00Z,9999-01-01T00:00:00Z
```
ozym: Should there be a 21 PWZ? If we stick with 20 then the location lat/lon shouldn't change.

salichon: Hmm right I thought this was close enough to be the same,
although after discussion with Mark, i understood....
So back to  PWZ 21 
I ll proceed accordingly in an updated branch

salichon:
This PR is outdated by https://github.com/GeoNet/delta/pull/260

PR #264, opened 2018-04-03
updated axial setting on strong motion streams

ozym:
Switching a number of strong motion sites to have axial channel naming, i.e. Z,1,2.


salichon:
SMTC is axial and does not need a change

ozym:
@salichon merge when you're happy.

salichon:
Hi there thanks for this ..mark !
It looks good.

PR #265, opened 2018-04-03
remove unnecessary duplicate GNSS sessions

elidana:
Hi,

in this pull request I've done some long overdue tidy up on the GNSS install/sessions file.
The file was showing duplicate GNSS session entries that are not necessary now that we stopped importing from Oracle-DB.
Apologise if there are lot of entries to review.
 
@staylorofford please take extra care in reviewing the entries

@ozym I've tried to create rinexml files from this pr and compared with what is currently present on S3. For some sites there are differences, I think mostly due to the introduction of the "full-GNSS" session .

@danieldooley I'm including you here to check if this is going to cause issues with gloria protobuf generation, please reassign if not for you. As far as I know there are no other dependencies on that file.

Thanks!

Elisabetta

/cc @nbalfour


danieldooley:
Shouldn't affect anything from Gloria 😄

PR #272, opened 2018-04-17
RGKA GNSS receiver and antenna swap

elidana:
RGKA NetRS to NetR9 and Zephyr1 to Zephyr2 swap.

@peanutty please reassign if necessary


elidana:
ha ha thanks Dan!! I've fixed that now

PR #269, opened 2018-04-09
LPOC closure

mnaguit:
For review LPOC closure

PR #273, opened 2018-04-17
add triggered field in streams file

ozym:
@quiffman it was a little tricky to use the response (recorder / sensor) pair on its own to determine whether a stream was triggered or not.

I've added an extra Boolean column (field) in the Streams file to indicate a triggered channel, or not. I've also updated the current streams file to match what would be given from the existing response files. This will allow us to retrospectively turn some trial streams into continuous, which we can't do now.

I've built the full existing stations.xml file before and after these changes and they agree apart from the creation date as expected. But it would be good that this can be verified, the stationxml tool is the only one that uses this information.

/cc @nbalfour


ozym:
I see one PR which will cause a conflict #269 so it may pay to wait for that to be merged, fixup the conflict here (new review).

salichon:
LPOC PR done :)

quiffman:
@nbalfour @salichon - I know we've spoken about this already, but can you just confirm that you're happy with this approach, please?

Unfortunately, this PR fell off my radar and is now in conflict...
As I mentioned earlier, this really needs to be a "stop-everything" merge. Because it changes _every line_ in the `streams.csv` file and git/github struggles to manage that automatically.

ozym:
@quiffman I'm also thinking there may need to be some adjustments needed if the datalogger location code (for SOH) needs to be inserted into the recorders / dataloggers file. This would also be a stop everything PR as well.

nbalfour:
@ozym @quiffman I'm very happy with this 👍

salichon:
:+1: hi there
@quiffman 
 I m happy with the triggered column.
Default is continuous (false) ->  all good 
thanks

ozym:
@quiffman had to add a NL to end of the streams.csv file. - hopefully it will pass muster

quiffman:
Ah, my bad. I had dropped that - it appeared as an incoming change, for some reason that I didn't fully understand.

quiffman:
I'll go and help @salichon resolve #287 now...

PR #291, opened 2018-05-09
HAAS NetR9 swap

staylorofford:
Also added new NetR9s to assets/receivers.csv


elidana:
@peanutty bump
...we have a few pr waiting for that one in the pipeline...thaanks! :-)

PR #295, opened 2018-05-13
KAIT new sensor Tsunami gauges

salichon:
Like for like change +  new asset entry


mnaguit:
Hi @nbalfour. @salichon requested me to do the review for you. Hope it's ok.

PR #304, opened 2018-05-21
NRRD NetR9 swap.

staylorofford:


staylorofford:
^updated PR with ground relationship data for NRRD (sorry @elidana)

PR #305, opened 2018-05-21
Include SOH response details.

ozym:
@quiffman @nbalfour Recent work by @francovm has highlighted the need to define what the SOH streams are and what their equivalent response information looks like (e.g.  find the power channel for a Q330 - VEP - and convert it into Volts, e.g.  multiply counts by 0.15). I am working on a document which describes each recorded channel which will help to address the _what is_ it question.

Unfortunately this information wasn't included in the original scope so I've had to re-purpose the "Role" field for datalogger connections (which does essentially the same thing) and to add one to the `recorders.csv` file, as no connection was required for this geometry.

Where given  the _Role_ should match the location code of the miniSeed streams from the particular datalogger or recorder. This will allow matching the actual datalogger or recorder to the recorded SOH streams.

For response information I have initially coded the SOH channels for the _Enta 2_ and the _Q330_ family of loggers. I've utilised the A2D response stage for these to handle the gain scaling. This has required some adjustment to the response building code (to list the SOH responses) and the StationXML building code to include these changes.

This is a breaking change for the `recorders.csv` file. The change is quite simple, add a column header, and then add an empty row after the Station Code (expect for those configured), but it impacts every row.

For reference I've also updated the CAW and the Etna 2 installations (TENS, GLOS, LOKS).

If we use this we'll need to work through and populate the `recorders.csv`, `dataloggers.csv`, and `connections.csv` files as required. No changes are needed for instruments that don't produce SOH files.


ozym:
@junghao would it be possible for you to checkout this PR into a local branch and then build and test that the generated stationxml file will load okay into FDSN.  (cd tools/stationxml; go build; ./stationxml > all.xml).

junghao:
Tested with local FDSN loaded with 2199 stations.

ozym:
Thanks @junghao super

/cc @quiffman

salichon:
Hi  there!  this might be re-assigned  to @JonoHanson in place of @nbalfour :).
i ll add a question mention for the DMC meeting

salichon:
have assigned @JonoHanson   in place of Natalie for future discussion and dispatch

JonoHanson:
was there a reason why this wasn't merged in 2018? and what were the hanging issues that were unresolved? as far as I can tell it looks ok. 

This is mostly going to touch Delta, so if it needs resurrecting I'd say it comes up there and gets a recommendation from that group.

ozym:
In the passage of time the goal is probably the same, just the mechanism may be a little different now. May think about this too with regards low rate datalogger and FITS info.

ozym:
I think the aim here will be to make "SOH" channels, simply new data feeds and work on it from that direction.

PR #306, opened 2018-05-22
add channel naming documentation

ozym:
@nbalfour a first attempt at some documentation about the seed stream naming conventions we use.

/cc @quiffman


quiffman:
Awesome.

I bet former site-selection folks like @labland and @carolinelittle would be happy to see this too!

PR #315, opened 2018-06-04
RGTA NetRS to NetR9 swap.

staylorofford:


timjmcdougall:
Check serials etc. Looks good.

PR #324, opened 2018-06-19
Close LPCC station

mnaguit:
Both streams (BN* and HN*) are closed


ozym:
@mnaguit the sample rates in the streams.csv file need to be sorted in numerical order, e.g. 50 comes before 200.

PR #325, opened 2018-06-20
Added polarity information for BFZ, BKZ, HIZ, KHZ, ODZ, OUZ, QRZ.

jristau:
I have updated the file 'streams.csv' with polarity information for the stations in the subject line. I changed the 'Reversed' field to 'true' for when the station has a reversed polarity, and put in the appropriate dates.


ozym:
@jristau looks like you've used an old version of the streams file to update. This means that there are monster conflicts. You will have to level up on your git foo to fix this, but I suspect it may be easier to simply make the same changes on the latest version of the file.  (then do a force push onto this same branch).

PR #326, opened 2018-06-20
Added polarity information for BFZ, BKZ, HIZ, KHZ, ODZ, OUZ, QRZ.

jristau:
I have updated the file 'streams.csv' in 'install' directory with polarity information for the stations listed in the subject line. I changed the field 'Reversed' to 'true' with the appropriate dates for when polarity is reversed.


jristau:
Thanks Mark - I didn't know I had to change the 1 Hz and 0.1 Hz streams as well. I'll do that.

ozym:
Yup. We may have avoided it by having a separate file with only polarity reversals in it (based on station and site code) but that adds more complexity behind the scenes rather than pain up front

jristau:
I've noticed that several of the stations I changed also have a 50 Hz stream, but that stream ends before the other streams start. Should I leave that one as is?

ozym:
These will most likely be the original EARSS installations. If the triggered column is true then skip them.

jristau:
I've updated the polarity information in the streams.csv file for the 0.1 Hz and 1 Hz streams. 50 Hz all have the triggered columns set to 'true', so I left them alone.

salichon:
@ozym @jristau @mnaguit @nbalfour 
Should we merge all the pending Pull request that are updating the streams.csv file,
before merging this polarity PR ? (not to break things ;) )

ozym:
@jristau small sets of changes like these should be okay to merge anytime, only problem would be if someone else changed the streams for these exact sites.  The previous issues were related to having every line in a file being changed so the smarts of git couldn't work out what should be where.

nbalfour:
@jristau, since this pull request is assigned to you it is your responsibility to merge it. Note you will have to update your branch first.

elidana:
@jristau can you please merge this one?

PR #327, opened 2018-06-20
HCDS equipment change

mnaguit:
replaced CUSP-3D


salichon:
replaced cusp3D  seems good to me
former one did not produce data in archive since 2018.061  - (March 2018)

PR #342, opened 2018-07-09
ETAZ changedatalogger

mnaguit:
change from Q330/3 to Q330S/6


mnaguit:
Hi @peanutty.
This is the last of the four AZ stations with datalogger changes. Does this look ok? :)

PR #358, opened 2018-07-26
OHWZ changelogger

mnaguit:
change datalogger from Q330/6 to Q330S/6


mnaguit:
There were conflicts related to the csv files being modified.  I tried to resolve the conflicts (new entries came in which should have come before the line for OHWZ), maybe due the branch not being updated?

PR #367, opened 2018-08-08
NCPT replace sensors

mnaguit:
replace Druck PTX-1830 with Druck PTX-1835


ozym:
@mnaguit note PTX-1835 is a new sensor model so extra changes will be needed, to avoid problems displaying RT data.

ozym:
@peanutty do these behave the same as the PX-1830 models?

mnaguit:
Thanks @ozym 
If there are other steps I need to do, please let me.
Basically, what I did is register the PTX-1835 in assets/sensors.csv and define the closing & starting times in install/sensors.csv

ozym:
@mnaguit you've done all you need to do, which is correct. Ordinarilly this would just be merged, but I've blocked it as we (e.g. me) needs to add the response information for this sensor model. It may be very simple (e.g. it behaves the same) or it may need some extra work.  The danger is that the tsunami gauge plotting routines will ignore these sites as they will have no response information, whereas now they have the wrong information, but in reality it is probably very close, if not correct, to what it should be.

/cc @quiffman @JonoHanson

ozym:
It is also possible that the PTX-1830 is a family of sensors and this is just one model variation, but is essentially the same (from a meta-data point of view).

mnaguit:
Okidoki ... many thanks @ozym 👍 
Will be extra careful next time if it's a new sensor.

On the other hand, I have created recent pull requests wherein I have registered a new sensor too (for Lennartz LE-3Dlite MKIII), are the delta and responses ready for this sensor? Thanks!

ozym:
@mnaguit yes, see e.g. #220

mnaguit:
Great 👍 
Thank you Mark!
@ozym

peanutty:
@ozym @mnaguit they are indeed the same! you can just call it ptx-1830 series if it makes it easier...

ozym:
Thanks @peanutty, I suspected as much. @mnaguit could you rename the models from 1835 to 1830.

mnaguit:
Thanks Mark & Dan. I have already rename the models from PTX-1835 to PTX-1830

@salichon @peanutty @ozym

mnaguit:
Thanks Mark! 👍

PR #383, opened 2018-09-12
VRZ replace Q330 and Basalt

mnaguit:
replacement:
from Q330/3 to Q330S/6
from Basalt to Obsidian


JonoHanson:
@salichon following on from our chat about this, as this is data without meta-data, and of short duration, there is no need to keep this orphan (a nice word for meta-data-less data) data.

PR #386, opened 2018-09-18
TSZ change logger sensor

mnaguit:
Change datalogger (from Q330S/3 to Q330S/6)
Change sensor (from CMG-3ESP to CMG-3ESPC)


mnaguit:
Many thanks @samgeonet 👍

PR #408, opened 2018-10-12
adjustments to station dates

ozym:
@salichon @mnaguit These are a number of automatic adjustments to the `station` start and/or end times to match the `site` start and/or end times.

The station start/end times tended to be done to the nearest day, whereas the site times tended to be more related to the recorded data.

I've just extended the `station` start/end times slightly so that the station times encompass the site times.

The only oddity I saw was --

`135A,SX,Maui B Offshore Tower (motor control)`

The site was still open but the station had a end date. I suggest this is another PR.

This is the first of a number of post oracle delta cleanup steps.


ozym:
@salichon I'm not sure what you mean by "locality" name?  Each station (e.g. "AIRO") has a "Station Name" which is usually related to where it is. This shouldn't change as it will have been allocated when the Station Code was registered.

There are also a lot of stations in the lists which have no data in our archives (mostly temporary XX) - this is a holdover from when the database was also used to track where the equipment was, not just the recorded data.

salichon:
@ozym oh sorry for the confusion  i meant "place"  in the 
 install/dataloggers.csv for instance  

for the place "Airedale" which corresponds to the AIRO station 
the datalogger is closed after the station  (2015-04-30T00:00:00Z)

ozym:
@salichon the "place" in that context doesn't really have any meaning other than that it joins the datalogger and collections tables. So we should make sure the stream/connection dates are correct for the closed stations (next tests).

PR #411, opened 2018-10-18
TIWZ Logger Sensor change

mnaguit:
from Q330/3 to Q330S/6
from Sercel L4C-3D to Lennartz MKIII


mnaguit:
Hi @ozym . Good morning.
Does 'complex' tickets like this need review from 3 teams (data, network & platform?). Thanks.

ozym:
@mnaguit a change like this should still be fine for the data/network teams. It's swapping a known configuration for another known configuration. The platform (or a wide group) may be needed when one of the configurations is new (i.e. a sensor that has never been used with a model of datalogger).

PR #425, opened 2018-11-08
MSWZ EH Sensor orientation correction

salichon:
As mentioned in the ticket https://github.com/GeoNet/tickets/issues/2927
The previous MSWZ EH  sensor orientation is corrected a posteriori

FYI @ozym @peanutty


salichon:
thanx this will done

PR #434, opened 2018-11-15
Historical WEL station Split WEL WL and WN

salichon:
The WEL station covers historically  several sites 1884 to Now
This has been addressed by splitting into 3 sites and names


JonoHanson:
are there are any corresponding changes that need to be made to parts of the archive? and do we do anything about sensor meta data for these really old ones?

salichon:
> are there are any corresponding changes that need to be made to parts of the archive? and do we do anything about sensor meta data for these really old ones?

If we are talking about the digital data , the first ever WEL one was: 
NZ.WEL.11.SHZ.D.1986.265

The related WEL Metadata aren't affected buy the Change. ( -> it started with this 1986 data)

Here I think we are more on the Historical meaning and representation sides of the Wellington station over the past which were actually several stations ...

(no FDSN/IAsPEI  standards to define Wellington then, more the good old teacher philosophe turning modern newly seismologist and having his seismometer in his cottage or in the brand new museum :) ). 

(more or less like this raspberry shake now ;) https://shop.raspberryshake.org/)

salichon:
This has been superseded by https://github.com/GeoNet/delta/pull/435 ticket

PR #444, opened 2018-12-04
LREZ sensor change

mnaguit:
close Sercel and open Lennartz MKIII


mnaguit:
Many thanks @samgeonet 👍

PR #445, opened 2018-12-05
CKHZ sensor change

mnaguit:
replace Sercel with Lennartz MKIII


mnaguit:
thank you @samgeonet 🥇

PR #452, opened 2018-12-11
KBAZ EH new datalogger

salichon:
new Q330S+ for EH 12Z borehole sensor


salichon:
Thanks @brianf3rris I ll proceed the merge for KBAZ

PR #457, opened 2018-12-20
JCZ station Slates and connection

salichon:
Polish the metadata and the configuration
Mostly Address the Slate, the connection of the former sensor and the end/start times
see https://github.com/GeoNet/tickets/issues/3760


salichon:
@samgeonet  the "connection" corresponds to the wiring between the Sensor and the datalogger.
The strong motion sensor was removed but not the virtual "wiring" between the datlogger and the sensor.  This is done in that change.

The new recorder does not need a connection config since its a combo embedding a  (sensor +datalogger)

PR #463, opened 2019-01-16
SBRC BN stream 2016 fix update

salichon:
2016 legacy update fix BN 20 50 hz to be mentioned
the stream BN was not existing hence abscence in the metadata


salichon:
@mnaguit yes indeed
but i took the time mentioned in the other files from delta when the CUSP change likely happened :)

PR #465, opened 2019-01-17
updated kinemtrics sbepi paz

ozym:
The PAZ lookup for the borehole SBEPI sensor was set to a missing value, this should have been pointing to the existing FBA-ES-T lookup.


salichon:
Wow That was quick!
thanks Mark! 
it looks like it might solve also for  older pre- 2015  SBEPI sensor RHBS response as well ! 
by the way it also helps in the response making understanding :). 
Cheers

PR #471, opened 2019-01-24
OGWZ logger and sensor upgrade

mnaguit:
from Q330/3 to Q330S/6
from Sercel to Lennartz


mnaguit:
Noted with thanks @brianf3rris . Correction in the station code is now underway c/of @salichon

PR #476, opened 2019-01-30
EDRZ sensor change

mnaguit:
From Sercel L4C-3D to Lennartz MKIII; and 
change in sensor orientation for old sensor (Sercel)


mnaguit:
Many thanks @samgeonet 
I have corrected the orientation of Sercel. Please review again. Thanks.
@salichon @jristau @brianf3rris

PR #477, opened 2019-01-30
Gloria Third Party

danieldooley:
Tracking which sites in Gloria are third party sites so we can exclude them from download.


danieldooley:
@staylorofford that is exactly what's happening

danieldooley:
@staylorofford are you able to merge this?

PR #479, opened 2019-02-10
Missing BHHZ-Black Hill datalogger closure

salichon:
Fixing the missing datalogger closure


salichon:
thanks  ill proceed to the merging correction

PR #481, opened 2019-02-11
WHVZ logger & sensor (acoustic) upgrade

mnaguit:
from Q330/6 to Q330S/6 
from Setra 270-600/24V to 270-600/12V


salichon:
Check with this :https://github.com/GeoNet/tickets/issues/3954

PR #482, opened 2019-02-12
PPHS replace recorder

mnaguit:
replace faulty CUSP3D


salichon:
FYI @mnaguit it s all good 
Anyway We ll get a confirmation of the  PPHS recorder orientation with mr @peanutty 
It should not have changed but checking if it s still a 213 degres one.

peanutty:
i can confirm orientation is as it was previously

PR #501, opened 2019-03-21
updated sitelogs to include TRM115000.00 diagram

ozym:
Not sure if this is actually used, but I've updated the sitelog diagrams to include the TRM150000 receiver diagram.

This impacts installations at fox1 and vexa.


staylorofford:
Nice picture, what does it do?

ozym:
these sorts of things ... ftp://ftp.geonet.org.nz/gps/sitelogs/logs/bluf_20180110.log

they were traditionally managed by hand ....

staylorofford:
Definitely beyond my knowledge, but @elidana probably has some ideas on this.

elidana:
on a related note, to the best of my knowledge those are not used (yet). IGS Sitelogs production is definitively something that needs an overhaul , so it's excellent if we can keep these files updated.

PR #505, opened 2019-03-26
Update dataloggers.csv to add new Q330S/6 dataloggers prior to deployment.

merijnthornton:
Added 11 Q330S/6


salichon:
Hi, Kia ora  @merijnthornton @elidana @samgeonet @staylorofford @mnaguit 
This branch/ticket  should be publicized more and reviewed by 2 teams at least the platfporm and the data naaaahh ?
Cheers! Please edit a  reviewer list and we proceed :)
NB: We usually upadte this list while setting up  a change  on the field - but  don t mind  otherwise !
Thanks
Cheers Regards Namaste Choukran Grazie
Jerome

ozym:
@salichon this looks like a new set of dataloggers has been added to the inventory (rather than being deployed anywhere). I'll update to the title.

samgeonet:
Usually we leave this addition to the delta repo up to the data team once something is installed. Do you want to keep going down this track or do you like what Merijn has done?

salichon:
I  don t mind either ways. - perhaps it brings more clarity for the platform team so i like it.
and indeed if  it stays on the "shelf" meanwhile  all good for you to do so :).
Maybe just a FYI to data team. 
anyway we need some reviewers.   
thanks 
Cheers

ozym:
From my perspective it's best if the equipment (when purchased) is added to the inventory prior to deployments. This can be reviewed at this point (e.g. equipment is eyeballed and the asset number matched to the serial number - this means this part is checked and reviewed.

Then in the field when something is installed with a serial number that doesn't match anything we own then a flag is raised, otherwise if you add both at the same time then there is no check (they are not independently reviewed) and an error can be introduced and only discovered when a clash happens at a later time ...

ozym:
Perhaps the use of the phrase assets is causing the problems, this in reality is simply a list of equipment that can/could be installed in the field.  It's aim is to stop the same bit of gear to be installed in multiple places at the same time ...

samgeonet:
Ok i think for this PR one from Data and one from RIM team do the review. We can discuss how we want to proceed at the next Delta meeting
Thoughts?

salichon:
RIM  like Remote Infrastructure Management aka old platform aka techs
... new acronym digested ... okay :) 
NB: The rim is the "outer edge of a wheel, holding the tire"

samgeonet:
@salichon I think you can merge now

salichon:
@samgeonet mureil s done it feel free to close this ticket or @merijnthornton

PR #513, opened 2019-04-10
Update cameras.csv

JonoHanson:
all installed cameras have more information


danieldooley:
Added a commit to attempt fixing build issues

mnaguit:
Hi @JonoHanson . I can do the review for this, if there's some background details available to compare the changes with?

JonoHanson:
the information is pieced together from SIT, but mainly from Richard Johnson and this ticket
https://github.com/GeoNet/tickets/issues/576
with a slight update through personal communication.

additionally, this data makes up part of the this other ticket
https://github.com/GeoNet/tickets/issues/4475
but both the table in ticket 4475 and the change I want to make here come from me, so the data source is no more reliable.

JonoHanson:
good catch @mnaguit . the extra info about the pole the camera is on should be stored in SIT. I've added a corresponding note in the places.csv

Update place.csv #1285

mnaguit:
Noted @JonoHanson . Thanks.

PR #518, opened 2019-04-17
NTVZ Datalogger change

salichon:
Q330 -> Q330S/6


salichon:
please @samgeonet check the asset references =/= from the ticket

NB: Check ticket and bring me chocolate if i m right ;) as mentioned in the ticket

salichon:
i ll proceed - Asset reference --> assumed
i ll get this during the verification pahse

PR #521, opened 2019-05-08
URZ borehole sensor Orientation

salichon:
set URZ weak motion sensor to estimated orientation 
(305N)
FYI @timjmcdougall @JonoHanson @ozym 

Cheers
Jerome


salichon:
thanks I ll proceed

PR #522, opened 2019-05-09
CRSZ-BN Sensor orientation (2015-2016)

salichon:
CRSZ BN Orientation Corrected back in the past (2015-2018)
FYI @ozym @JonoHanson @peanutty


salichon:
thanks I ll proceed

PR #537, opened 2019-06-06
SensorNamingCorrection

salichon:
PTX series 1830 and 1835 are the same
hence their response
Default is 1830  then


salichon:
Corrected and Done As requested by @mnaguit sorry @samgeonet  :)
Cheers
jerome

salichon:
@muriel sorry a last polishing related to sorting instruments
one review should suffice

PR #543, opened 2019-06-18
MQZ recorder swap

mnaguit:
from Basalt to ETNA 2 (from 50hz BN*streaming to 200hz HN* streaming)


mnaguit:
Hi @ozym 
Should I include you in the reviewers for this PR? It is a recorder swap, from Basalt to ETNA2, with streams changing from 50Hz to 200Hz,

FYI @salichon

PR #547, opened 2019-06-19
RDCS recorder change

mnaguit:
replace faulty CUSP3D


mnaguit:
Hi everyone. Can u please do another round of checking for this PR? Azimuth is updated from 360 to 0 for both removed and installed CUSP3D. Thanks.

PR #548, opened 2019-06-19
New Zephyr 3 GNSS antennas - Paul

paulaic:


elidana:
just noticed that might be worth rebasing this before merging

PR #558, opened 2019-06-25
MOVZ replace datalogger

mnaguit:
from Q330/3 to Q330S/6


mnaguit:
Hi. Can you please do another round of checking? Much thanks. 
@samgeonet @jristau @brianf3rris

PR #563, opened 2019-07-01
added responses for VTZ data streams

ozym:
@salichon @mnaguit updating the response information for the VTZ channels which are generated on Raoul Island and at Puysegur Point to allow sending low-rate data over the BGAN satellite link.

This is essentially a clone of the 1 Hz stream (LTZ) with the rate changed from 1 Hz, to 0.1 Hz as well as the associated Q330 FIR filters.

/cc @quiffman @nbalfour the main aim is to get these channels into the station.xml file, so that they can appear in the FDSN stations channel lists, allowing them to be monitored for latency.


salichon:
thanks sorry  not to have seen this PR before

PR #564, opened 2019-07-02
NRRD NetR9 swap

staylorofford:


samgeonet:
Good point I will get Brendon to add the details of the antenna to the change ticket

samgeonet:
@staylorofford antenna details are in the change ticket now

staylorofford:
updated

elidana:
looks good, antenna sn. verified against Dropbox photo, receiver sn. and timestamp verified on receiver web interface

PR #565, opened 2019-07-04
KHZ -Kahutara -  slate and strong motion sensor change| 2013-2014 legacy

salichon:
- Slate 463 instead of 285
- only a strong motion sensor swap no slate swap (always been 463)
- Sensor swap happened in 2013 not in 2014 

-Slate 287 should be on site . perhaps further enuiry for this 
Cheers
jerome


salichon:
@ozym  FYI more than review :)

mnaguit:
same with @samgeonet , i'm not familiar with slates.go. I approved the review, just checking on the year change (from 2014 to 2013) and slate number (from 285 to 463)
FYI @salichon

PR #573, opened 2019-07-11
GNSS: RGTA antenna swap

elidana:
- time stamp from data
- antenna serial number verified on pictures


elidana:
thanks @staylorofford 
for the time stamp I've choose a few minutes before the first epoch in the rinex, but, within the data gap. 
Ashton gave me the antenna photo, I will upload it on Dropbox now

PR #578, opened 2019-07-30
CJHB new structural array

mnaguit:
with LC 21 to 29 and 2A to 2G


ozym:
As far as I am aware there is no data for this array.

PR #580, opened 2019-08-08
TROB update metadata

mnaguit:
updated _streams.csv_ to include LC 22 and 28. These channels started on 2013-08-02 and were reportedly closed on 2014-07-01.


salichon:
Okay ..this is a temporary fix  before additonal information

mnaguit:
I'll merge this PR for the closure of streams based on the available information.
Once we receive additional info, then we can start another PR to revise as needed.

PR #581, opened 2019-08-09
Updated datalogger connections mapping.

ozym:
This is a little tidying up from the initial deployment.  I've made the check for "place / role" now an error of there isn't a match between the entries in the connections.csv file and the dataloggers.csv file.

This was hiding some warnings that should have been fixed a while ago. The main one being the strong motion channels for the site at Blackbirch not being recognised.

There were a number of sites around Ruapehu that used FM telemetry back to the Chateau Observatory, I've deployed the datalogger then and updated the place / role for these sites.
(Their response information will only be approximate as they used FM analogue tecniques).

Feel free to add more reviewers as needed.


salichon:
@ozym thanks for this 

jsut FYI about the delta changes 
@brianf3rris @JonoHanson @jristau about testing  the fieds "place" and "role" when  pluging a sensor to a datalogger (csv files )
cheers
jerome

PR #583, opened 2019-08-14
MRZ continuous 200 Hz are No more Triggerred

salichon:
Continuous 200HZ streams ETNA  2 
Corrected for the change of streaming mode in the response 
for MRZ HN 20


salichon:
@ozym Hello
I guess this MRZ  triggered to Continuous stream delta  correction is eventually just for your information :) 
I deem i need one review only
cheers
Jerome

PR #592, opened 2019-09-09
TMHS 21 is 12Z Not axial ENZ

salichon:
TMHS  metadata are  ENZ instead of 12Z 
Change for correction


salichon:
thanks i ll merged it therefore

PR #594, opened 2019-09-10
GLOS metadata  fix - Legacy 2017 

salichon:
GLOS is a strong motion  with ENZ components  not 12Z since end  of 2017
- Changing the metadata instead of the data 
- I d assume configuration issue at that time 
- Used the start of the ENZ streaming to have the correct metadata now.

FYI  and or review @ozym


salichon:
Resolving conflict

PR #597, opened 2019-09-11
GNSS CLIM receiver swap

elidana:
CLIM receiver swapped back to NetR9. Also fix the session file


elidana:
@samgeonet can we please have your blessing on that one? :)

PR #598, opened 2019-09-12
StationSiteClean (IU, AU, OE networks, year 2020 stations )

salichon:
Remove stations begining in 2020  (would have been seen in FDSN i suppose)
Remove OE  network stations
Remove AU network stations
Remove IU  network  stations except SBA and SNZO


salichon:
Hi 
I dont now if it s a goos idea but  it seems relevant to keep in delat only the station we re using
and also spring clean the staion begining in 2020, before it happens

We coudl also document the historical overseas station lcoation and names into the readme part ? 

Cheers
Jerome

salichon:
@elidana Woudl you add this ticket  into the next delat agenda ? 
thanks a lot
grazie
cheers
jerome

ozym:
There may be some uses of the OE, AU, and IU site location information having been used for locating earthquakes prior to GeoNet. I'm not actually sure whether this is the case (other than SNZO), and if so, how big an issue it is. Whether this information needs to be in delta I'm not sure, it may simply need to be switched into an archive document somewhere especially if the earthquake location data (QuakeML) is self-contained. There have also been other sites associated with NZ, such as RAO, SBA, and RAR ...

ozym:
I'd be much happier if this was two pull requests, the first 2020 stations would be easy and can be merged. The second is a little trickier as it has implications on what delta is aimed at and what it is used for.

salichon:
thanks @ozym 
i ll split up into 2 tickets and PR 
and let the overseas station case pending while making the best decision.

salichon:
Will Close this bracnh and split it into 2  different issue as mark advised
- 2020 stations (easy)
- Overseas historical stations information of Delta (To be developped and discussed more in depth)

PR #599, opened 2019-09-12
add caps tool

ozym:
This is a fairly simple tool to add caps style template files to a given directory.

It has a few features that may not be used yet (e.g. unmatch) as the environment that it will run in hasn't been completely defined yet.

More constraints which stations should be included can be added (or removed) once we have a better idea of what's needed in the actual caps instance.

### DIFF
--- a/tools/caps/main.go
+++ b/tools/caps/main.go
@@ -1,5 +1,9 @@
 package main
 
+/**
+build template files for configuring a caps system based on site details
+*/
+
 import (
 	"flag"
 	"fmt"
@@ -19,6 +23,20 @@ const contents = `###
 ###
 capslink:sl4caps
 `
+
+func match(s string) string {
+	switch r := strings.Split(strings.ToUpper(s), "_"); len(r) {
+	case 2:
+		return strings.TrimSpace(r[0]) + "_" + strings.TrimSpace(r[1])
+	case 1:
+		if strings.ContainsAny(r[0], "*?\\[") {
+			return strings.TrimSpace(r[0])
+		}
+		return "*_" + strings.TrimSpace(r[0])
+	default:
+		return ""
+	}
+}
 
 type Station struct {
 	Network string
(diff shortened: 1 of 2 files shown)


tools/caps/main.go
```diff
@@ -0,0 +1,214 @@
...
+func main() {
+
+	var base string
+	flag.StringVar(&base, "base", "../..", "delta base files")
+
+	var output string
+	flag.StringVar(&output, "output", "keys", "output caps configuration directory")
+
+	var networks string
+	//TODO: this could be managed via configuration elsewhere.
```
ardrigh: What ideas do you have on getting this from another source?

tools/caps/main.go
```diff
@@ -0,0 +1,214 @@
...
+					installation.Start)
+				if stream == nil || err != nil {
+					if err != nil {
+						log.Fatalf("unable to load metadata streams %s: %v", station.Code, err)
+					}
+					continue
+				}
+
+				for range response.Channels(stream.Axial) {
+					//TODO: filter here, maybe regexp on channel, or look at equipment.
```
ardrigh: What would a filter help with here?

ozym:
good point @ardrigh will add a test case or two.

The filter may be something more specific than simply network codes, i.e. we want all instruments that are (insert query)  ... this is a little vague as we haven't gone through the entire flow.

ozym:
for testing it may be better to split it up a bit ....

tools/caps/main.go
```diff
@@ -0,0 +1,214 @@
...
+	flag.Parse()
+
+	check := make(map[string]interface{})
+	for _, n := range strings.Split(networks, ",") {
+		check[strings.TrimSpace(strings.ToUpper(n))] = true
+	}
+
+	var unmatch []string
+	for _, p := range strings.Split(exclude, ",") {
+		if s := func() string {
```
danieldooley: The anonymous function strikes again. There's a lot going on here, if a function helps simplify it break it out into a named function; and maybe with some tests; otherwise simplifying it to more linear execution would help a lot.

tools/caps/main.go
```diff
@@ -0,0 +1,214 @@
...
+			continue
+		}
+
+		installations, err := db.Installations(station.Code)
+		if err != nil {
+			log.Fatalf("unable to load metadata installations %s: %v", station.Code, err)
+		}
+
+		network, err := db.Network(station.Network)
+		if network == nil || err != nil {
```
danieldooley: See other PR comments on two seperate `if` blocks

danieldooley:
Oh also maybe some more comments 👍

PR #600, opened 2019-09-13
updated sc3 tool

ozym:
Another initial go at an automatic template builder for sc3.

### DIFF
--- a/tools/sc3/main.go
+++ b/tools/sc3/main.go
@@ -1,0 +1,201 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/GeoNet/delta/internal/metadb"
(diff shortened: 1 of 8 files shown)


tools/sc3/main.go
```diff
@@ -0,0 +1,201 @@
...
+			continue
+		}
+
+		installations, err := db.Installations(station.Code)
+		if err != nil {
+			log.Fatalf("unable to load metadata installations %s: %v", station.Code, err)
+		}
+
+		network, err := db.Network(station.Network)
+		if network == nil || err != nil {
```
danieldooley: I think this makes it a little harder to read, not a deal breaker.

```go
if err != nil {
   log.Fatalf(...)
}
if network == nil {
   continue
}
```

See a few below as well (and the one on L186 is pretty ugly).

ozym:
@danieldooley have tidied it up a bit.

PR #601, opened 2019-09-15
WAKS is orientated ENZ now

salichon:
from the new install the WAKS orientation is true North 
metadata modified


salichon:
Ata Marie
thanks for this 
merging ....

PR #602, opened 2019-09-16
HHSS 20 will be relocated

salichon:
HHSS  christchurch (CUSP) Site relocation. 
Site Closed until further notice since March  2019 
station is kept opened meanwhile


salichon:
kia Ora  @muriel 
I think the station location  will be kept the same: only the site (20) will  be updated - or even switched to  21 upon distance :). 
Cheers
jeorme

PR #604, opened 2019-09-18
HORN GNSS receiver and antenna upgrade

elidana:
please note change in station operator. 
Might need to add a network code change as well (under discussion)


elidana:
good catch @erabbott ! I have updated this pull request to correctly include session closing timestamp

PR #608, opened 2019-09-19
added ISO version of episensor to obsidian response list

ozym:
Will kick WAKS into life.


salichon:
Updated field
-- >kinemetics Strong Motion Recorders and Dataloggers Connected to Strong Motion Sensors:

PR #611, opened 2019-09-25
allow xyz as axial option

ozym:
This is a slight change to the "Axial" parameter for the `streams.csv` entry. Previously it was a boolean indicating the sensor channels were laid out like "Z12" (true) or "ZNE" (false).  However, a new layout has been added, namely "XYZ" as used in structural arrays".

The solution with the minimum impact is to convert the boolean to a string, with a small set of permitted values (e.g. "true", "false", "Z12", "ZNE", "XYZ") . The outputs and the tests have been added. 

In time the true/false could be removed, this may have impacts on outstanding pull requests so can be deferred as needed.


salichon:
great
FYI @mnaguit @JonoHanson

PR #615, opened 2019-09-26
CPFS change location code

mnaguit:
from LC 20 to LC 21


mnaguit:
Changes made @salichon . Thanks!

PR #623, opened 2019-10-01
TENS NZ station set 

salichon:
This is a 200Hz  permanent NZ strong motion now


salichon:
@ozym this may not affect platform team  process  - but i m double checking 
cheers

PR #624, opened 2019-10-01
GNSS DUND receiver and antenna swap

elidana:


elidana:
thanks @mnaguit and @samgeonet !

PR #625, opened 2019-10-01
ARAZ POIZ THQ2 are in TP SubNet Now

salichon:
Seismic Taupo Stations Sub network change to allow private station to be part of the public network
FYI @ozym


salichon:
@JonoHanson 
Hello! For your approval 
this will  turn these stations into  national stations 
cheers

PR #627, opened 2019-10-03
NTVZ Datalogger Correction

salichon:
New datalogger starts 2 days earlier
Typo correction


salichon:
one review necessary for this one :)
thanks

PR #632, opened 2019-10-04
Create CUSP3D firmware follow up

salichon:
- Aims at following up the previous CUSP3D  firmware update  (cusp3d+r400.74.1) that fixes a timing error
- Derived from the firmware.csv for GPS 
- List a datalogger cusp3d  in combination with a station (date of install of the cusp3d) and  a time frame for a firmware version
- Notes display a reason for the update or the abscence of information
- Location of this file and content are in delta/install/ and subject to change
  
FYI @peanutty @JonoHanson @MilesRyan 
cheers
Jerome


salichon:
This is a documention file proposal - or metadata for the metadata to help folloing up and correcting data in the past, or give start date for well timed data sets. (strong motion)  
Perhaps could this be described with a readme or is it self described enough ? 
Open to discussion and modification

samgeonet:
@ozym we decided at the last delta meeting that having one file for all devices would be unweildly, given history is involved and that a new file for each type would be more manageable

samgeonet:
@salichon I wonder instead of a firmware for the cusps we should make it firmware for strong?

salichon:
@samgeonet 
Kia Ora,
> make it firmware for strong ?

That s beyond my knowledge of instrumentation software and firmware and how standard amongst the providers they are.. :).   - 
A bit of discussion of the the next Δ meeting perhaps.

salichon:
@ozym hi, 
> Why didn't you just add this to the existing firmware csv file?

I did not want to mix GPS and seismic (besides strong motion) specifically if it's one-off existing file.
Tracking the firmware of the seismic instrument into delta is certainly something to be considered with regards to the Cusp case. 
cheers
jerome

ozym:
> @ozym we decided at the last delta meeting that having one file for all devices would be unweildly, given history is involved and that a new file for each type would be more manageable

I strongly disagree.

quiffman:
I disagree too. But perhaps we should move this discussion elsewhere...?

The major down side of creating a new file like this, is that it is vestigial.
It has not checks or tests wrapping it up, but it lives with the rest of the `.csv` files that do...

salichon:
Hi  there,
all right so it s way beyond the initial scope of documenting the CUSP3D then :)
I m proposing to 
- Close this PR 
- Raise a ticket about firmware for seismic data (at least) and the example of the cusps firmware

PR #644, opened 2019-10-21
Nply

staylorofford:


elidana:
good catch. I have resolved some conflicts, can you please check it looks good and squash this pr into one commit?

PR #652, opened 2019-10-25
GLWS location code & equipment change

mnaguit:
from LC 20 to 21
from recorder (ETNA) to sensor-datalogger (FBA-ES-T-ISO Obsidian 4x)


samgeonet:
@JonoHanson Which site do you want listed in sites.csv? The new one or the original?

mnaguit:
Hi @ozym . Thanks for the review. Just copied the original entry from the stations.csv. But no worries, will update it now. 👍

JonoHanson:
following on from @salichon comments, I think we should have an end-date to the original station location, and a new start up date for the new station location (presumably 2019-10-23 for both); 
or is there a reason why not?

ozym:
I think it comes down to how the information is being used (or was used). Usually the station location information doesn't change, if there is a big enough change then a new site code is made. The actual site details are stored in the site location, an example here is an array. The central point (or reference) vs the recording nodes.

mnaguit:
Hi @salichon @JonoHanson & @ozym 
Following up on your recent conversation, if I understood it correctly, are you referring to the changes made for the stations.csv? I think (please correct me if I'm wrong @ozym ), there is only one entry for every station in that file, regardless if there are a number of location codes per station. Hence, it's not possible to reflect the closing and opening time if there is a change in LC. The closing-opening will be reflected in streams.csv and sites.csv.

That is why I'm asking recently, which coordinates should be reflected in stations.csv. Because changing it will wipe out the previous coordinates, and it will be taken as if these new set of coordinates is where it was since the station was opened (this case in 2002).

salichon:
Okay thanks for the information
I would agree with what @ozym mentioned  
"_Usually the station location information doesn't change,_ etc . .  "
so perhaps say as a common practice, we keep station.csv as it was considering it a reference information indeed. (even though no more the location of the station hutt)

mnaguit:
Okidoki @salichon. Noted with thanks.
Based on conversation with @JonoHanson & @samgeonet last Friday, the preference is to reflect the new coordinates in the stations.csv file.

Kindly let me know which is the final procedure to follow and I'll change the file accordingly.

mnaguit:
Hi @JonoHanson 
Shall we keep the stations.csv as shown here in this PR or revert to the old file with the original coordinates set in 2002? Thanks.

JonoHanson:
@mnaguit based on the conversation above, I think we should leave the station.csv with the original co-ordinates (the 2002 co-ordinates). Sounds like that is the more consistent usage.

mnaguit:
Noted @JonoHanson . Thanks.
The _stations.csv_ is now updated, reflecting only the change in spelling of Glynn, otherwise, all other details remain as is.

Can someone review this again so that I can proceed with the merging? Thanks.

mnaguit:
Many thanks @samgeonet & @ozym 
I think I can now proceed with the merging, noting that previous changes other than _stations.csv_ were already reviewed.

PR #660, opened 2019-11-01
WKZ change in location code

mnaguit:
from LC 20 to LC 21


mnaguit:
Hi @ozym . Would like to request for your feedback on this.
For these set of stations that require LC change (at least 30 stations), should we reflect the date/time when the location of the equipment changed (until we are ready to update the archive in AWS) or when the location code was changed from 20 to 21?

ozym:
@mnaguit we will do three steps.

1) fix the time to when the recorder was updated.
2) do the changes to the archive to relabel the code as needed.
3) fix the time to when the recorder was actually moved.

This means that we don't suddenly lose the data for 30 stations (except for the short window between 2 and 3).

ozym:
This means we have some control on the gap between step 1 and step 2 ...

mnaguit:
Thanks a lot @ozym . 👍

mnaguit:
1 Hz closed for LC 20 on 28 June 2017, per @salichon recommendation.
Can reviewers please do another round of checking? Thanks.

mnaguit:
Hi @ozym 
With the recent changes, does this PR look fine for you? Thanks.

PR #665, opened 2019-11-04
INSS recorder change

mnaguit:
from Basalt SN 1675 to Basalt SN 2157


samgeonet:
We need to talk to Jono about this one as it has had changes in location but no LC change ever...

mnaguit:
> We need to talk to Jono about this one as it has had changes in location but no LC change ever...

Noted with thanks @samgeonet ...now tagging @JonoHanson so that we're all in the same page on this.

PR #673, opened 2019-11-05
MWDS_TKAS_TUDS are Subnet SM now

salichon:
Moved strong motion stations from private SX to public SM


salichon:
Hi @ozym 
Do you require to review the subnet changes for Delta when it happens occasionnaly ? 
I reckon there is no downstream implication although i m not very certain.
 Cheers
jerome

PR #676, opened 2019-11-07
LTZ change in location code

mnaguit:
from LC 20 to LC 21


mnaguit:
For your review @ozym

PR #677, opened 2019-11-07
NNZ change in location code

mnaguit:
from LC 20 to LC 21


mnaguit:
Hi @ozym . Does this PR look ok for you? Thanks.

PR #680, opened 2019-11-08
update 4 GNSS monument heights

elidana:
Hi, this is to update recently measured monument heights on 4 GNSS stations.


elidana:
@samgeonet can you please review this?

PR #682, opened 2019-11-11
switched default caps keys directory

ozym:
Turns out it's "key", not "keys" for caps ...

### DIFF
--- a/tools/caps/main.go
+++ b/tools/caps/main.go
@@ -74,7 +74,7 @@ func main() {
 	flag.StringVar(&base, "base", "../..", "delta base files")
 
 	var output string
-	flag.StringVar(&output, "output", "keys", "output caps configuration directory")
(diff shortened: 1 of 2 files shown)


ardrigh:
It is actually `/opt/seiscomp/etc/key` directory for both SeiscomP3 and CAPS installation on the server itself.

SeiscomP3 is not yet using these files, I thought it had been updated to do so. I'll put in a PR for it to apply this in Dev to match CAPS.

https://github.com/GeoNet/puppet-gns_sc3/blob/b80189162f6083dad5a1caa27ae3352136f7d6ff/manifests/caps.pp#L44

https://github.com/GeoNet/puppet-gns_sc3/blob/b80189162f6083dad5a1caa27ae3352136f7d6ff/manifests/station_processing.pp#L10

ardrigh:
Related PR https://github.com/GeoNet/puppet-gns_sc3/pull/79

PR #683, opened 2019-11-12
only use the lowest matching location code

ozym:
@ardrigh @salichon There is a corner case at the moment that involves a site having multiple sensors.  The output sc3 profile may use a less desired stream ..... This PR will adjust it to favour the channel set with the "lowest" location code in an alphabetical sense.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -20,6 +20,8 @@ go test ./tools/spectra
 go test ./tools/chart
 go test ./tools/impact
 go test ./tools/rinexml
+go test ./tools/sc3
+go test ./tools/caps
 
 exit $errcount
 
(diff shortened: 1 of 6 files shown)


salichon:
Hello @ozym 
Can you describe me how this go function is working briefly :)? 

In my ( now outdated) manual settings  way  (Nothing that you don t know  anyway !)
- say as long as i had a code 10 or 11 available (data+metadata) i was enabling the weak motion profile (whatever additional stream)
- for the Code 20/21 - If it was the the unique stream available a strong profile: 
   if a  BN is available a strong_low rate,  HN otherwise. 

BTW with regards to the huge number of strong motion often over some very spatially focussed places (CHCH for instance) used to be  somehow restricted but since a wee while it s completely opened (which may create sometimes a larger data request over the GUIs than actually necessary)

cheers
thanks a lot 
Jerome

PR #686, opened 2019-11-14
GVZ change in recorder

mnaguit:
from ETNA2 to Basalt


mnaguit:
**Note to reviewers:**
In _streams.csv_, the 200hz continuous for ETNA2 was turned off and then another entry was opened for 200hz, this time triggered, for Basalt along with 50hz continuous.

Kindly let me know if that looks accurate for you. Thanks.

PR #690, opened 2019-11-15
DSZ change in recorder

mnaguit:
from Obsidian to Basalt
change in LC from 20 to 21


mnaguit:
Hi @samgeonet .
Can you please do another round of check for this? Thanks.

PR #691, opened 2019-11-15
INGS Basalt swap

mnaguit:
from Basalt SN 620 to Basalt SN 619


mnaguit:
Hi @salichon & @brianf3rris 
Can you please review this PR? Thanks.

PR #697, opened 2019-11-21
Kawerau NX KQ?? Closed | default date:2017.124 (No data since)

salichon:
The data from  the Kawerau field are closed
Last date  2017 May 4 02:00 UTC  used as a default to close
 these statiosn in abscence of finer details or data.
See ticket 
cheers
jerome


salichon:
@mnaguit Morena,
I did not include formaly additional reviewers since it s a legacy from the previous delta.
No data is recorded since a while and Only the metadata are affected (no actual downstream effect)
We can always add an additional one if you deem so. 
i ll check and correct KQE  and KQH indeed  - these Kawerau stations have no data stored or are opened still  - once again I closed them for convenience clarity using the 2017-05-04 date.
Cheers
Jerome

salichon:
@mnaguit done  accordingly
cheers

PR #699, opened 2019-11-21
GNSS HIKB receiver and antenna swap

elidana:


elidana:
@staylorofford any chance you can review/approve this?

PR #701, opened 2019-11-28
TOZ location code change

mnaguit:
from LC 20 to 21


mnaguit:
Hi @ozym . Another LC change. Can you please review? Thanks.

PR #702, opened 2019-11-28
MECS recorder swap

mnaguit:
from Basalt SN 615 to Basalt SN 1678


brianf3rris:
Looks ok. Basalt 1678 installed on Nov 27, 2019, at 22:47 UTC.

PR #706, opened 2019-12-02
HAZ SM  configuration  20 to 21 change 

salichon:
HAZ strong motion gets the right location code configuration 
1 Hz strong motion stream is closed


salichon:
@ozym  sorry i  squashed and merged before your review - apologies for the HAZ 20 to 21

PR #724, opened 2019-12-16
Update antennas.csv

bglendinning:


samgeonet:
@elidana likewise this one - can we utilize your merging powers

PR #734, opened 2020-01-09
DART Metadata Test - DEMO Branch -  

salichon:
DART Stations (BPR)  in Delta (Mimimum information) - TEST - DEMO - 
1- Integrated into Delta
2- Define a Subnetwork for DART network
3- Define a Station for each DART instrumentation: Anchor location, start and stop dates
4- Define  a site for each DART instrumentation: BPR location, no site code
5- Define the BPR as a dartsensor.csv install in delta: BPR sensor details and Depth
6- Register the sensor: BPR paroscientific into the assets in a dartsensor.csv file 

No response information or type of data distributed (water depth, pressure, temperature)
No specified output format

FYI @ozym @JonoHanson


ozym:
@salichon some quick comments -- it's always easier to get the meta data sorted out at the beginning of the project so this is great!!

The sensors are just pressure sensors so they could live with the other sensors in the `assets/sensors.csv` file unless there's a compelling reason to have a new file ...

The installation file `install/dartsensors.csv` mixes a couple of concepts which we have tended to split out, e.g for seismic there is an "`install/sensors.csv`" and an "`install/dataloggers.csv`" which describe what is where, and then there is a "`install/streams.csv`" which describes how they are configured. Likewise for GNSS, there's a `marks.csv` and `monuments.csv` for where things are put, but also a "`sessions.csv`" file for how they are configured.

Likewise these are just pressure sensors, so I don't see why they can't have that aspect managed by the `install/sensors.csv` file, with the other information being handled elsewhere. I'll make a straw man at the end of this spiel.

Networks looks okay but I don't see the need for '4G', will we have to rename the network if they buy the next generation system?  Personally I would call it something like "National tsunami deep ocean network".

The `network/sites.csv` looks fine, except I would add a `location` code. This could simply be a count of the deployment, e.g.  01 - now, 02 - when they go back and replace the batteries etc. The locations will vary but they may not change the site code (i.e. NZA will stay the same).

The goal (from my point of view) is to try and keep things as similar as possible to  the existing equipment, that way no new coding is needed (or at least only a little).

I would try and feed the "`install/dartsensor.csv`" into as much of the existing systems as possible, e.g. create a "`install/dataloggers.csv`" entry (the data is digitised somehow I presume?). Then the various modes can be managed in the "`install/streams.csv`" file, e.g. non-triggered, and triggered sampling rates. At this point the mini-seed codes would just fall out for each of the streams ... (can't see why this data shouldn't be available via fdsn?).

The final parts, Thresholds, and Event modes?  Are these needed from a meta data perspective, e.g. if you have been given the data, are these values needed? [This is clearly useful for real-time use and configuration, but maybe not in 30 years time].  If the answer is that there's some DART specific information that must be there, then at this point we can create a `install/dart.csv` (or something) file. There's also the possibility of simply writing some documentation that goes with the entries.

The only nagging thing I have is how we represent the sampling rates, sometimes they just don't work (i.e. 1 sample per minute is 1/60 sps), One of the tricks the SEED people used (to avoid floats) was to use the convention that if the rate was negative then it was really the sample period.  That is -60 meant 1/60 ... this may be something we can think about it (may come up with other datasets).

/cc @JonoHanson

salichon:
This will be worked out and been superseded - Can be closed

PR #737, opened 2020-01-14
THZ Strong motion Config change 20 to 21

salichon:
THZ colocated strong motion
Configuration change to account for site location of the instrument


salichon:
needs an update on lat/lon for site 20/21

PR #741, opened 2020-01-15
WHZ strong motion loc 20 to 21 config update

salichon:
WHZ colocated strong motion configuration change to account for location code


salichon:
Hello @mnaguit Would you re confirm your review
i had to solve for a conflict while merging 
thanks

PR #760, opened 2020-01-31
update GNSS antenna skyview at MQZG

elidana:
a recent visit to the site confirmed that skyview is a bit poor at the moment due to forestry pines growing to the north. They are almost mature so will be removed in a few years most likely.


environment/visibility.csv
```diff
@@ -91,7 +91,8 @@ METH,"Cloud cover made it impossible to see mountains to the west, calculations
 MING,Excellent,2010-04-28T00:00:00Z,9999-01-01T00:00:00Z
 MKNO,Obstructions to East up to 12 degrees formed by Mokai Patea Range.,2011-05-11T00:00:00Z,9999-01-01T00:00:00Z
 MNHR,excellent,2011-02-16T00:00:00Z,9999-01-01T00:00:00Z
-MQZG,No obstructions above 10 degrees elevation. Tree/gorse cutting may be required in future,1999-09-27T00:00:00Z,9999-01-01T00:00:00Z
+MQZG,No obstructions above 10 degrees elevation. Tree/gorse cutting may be required in future,1999-09-27T00:00:00Z,2020-01-29T23:59:00Z
```
staylorofford: Do you want the end time to be with 59 seconds? Or do you want to have a minute between these entries?

elidana: time here is not really used anywhere, and the change in skyview is much more gradual. It is mainly to report time of observation

PR #768, opened 2020-02-03
GNSS GLDB antenna and receiver upgrade

elidana:


elidana:
not needed, one already exists for GLDB swap https://github.com/GeoNet/delta/pull/767

PR #775, opened 2020-02-12
GNSS MKNO receiver swap

elidana:


elidana:
@samgeonet can I please have your 👍 on this one when you're back in the office?

PR #776, opened 2020-02-14
WGTN NetR9 to Alloy, Zephyr2 to Zephyr3 swap.

staylorofford:


staylorofford:
Should be good now. Thanks for catching that, for whatever reason I must have thought it normal.

elidana:
@samgeonet can you please re-approve? (i won't be able to merge otherwise) thanks

PR #778, opened 2020-02-19
WHFS recorder swap

mnaguit:
from ETNA2 to Basalt


salichon:
@ozym - as suggested by  sam you may  want to  have a check or a comment  of this recorder swap PR.
chers

PR #783, opened 2020-03-03
Close old mount and add new mount at T due to movement of pole

staylorofford:
Assigned reviewers from each team as per equipment change process - can review selection in delta meeting if needed


ozym:
@staylorofford Currently, if you change the mount then you need to change the camera code, e.g. there can be but one `T` ...

One possibility is to use the `Installed Camera` file option and use the X,Y,Z to shift the camera slightly (which I think is fine in this case), this has a time component which will mean it can be shifted back etc.

staylorofford:
Thanks all, I have updated the PR following your suggestions

staylorofford:
@ozym, @samgeonet, could you please re-review?

PR #786, opened 2020-03-04
NOVZ new station

mnaguit:
with LC 10 EH (ENZ) for seismic and LC 30 (1 & 100hz streaming) for acoustic


mnaguit:
Thanks for reviewing @samgeonet @ozym @elidana 
Had to resolve the conflict associated with the other acoustic-related PR.
If one of you can do another round of review, that would be great!

PR #787, opened 2020-03-05
Remove superfluous mounts.csv entry for N

staylorofford:
As per delta meeting, removing the extra entry for mount `N` in the mounts.csv file

It was introduced in PR: https://github.com/GeoNet/delta/pull/530 but is not required.


mnaguit:
Hi @staylorofford .

Sorry, I'm not well-versed with the details for mounts. It might be better to add reviewers who have a better understanding of this metadata. As for CVMA, I issued the PR while discussing it with @salichon , following the procedures related to webcam change as documented in wiki.

PR #795, opened 2020-03-12
WAHU GNSS receiver swap

CBurton90:


staylorofford:
smash that squash and merge @CBurton90

PR #796, opened 2020-03-12
Moved building of the meta to ac-tool

qhua948:


qhua948:
@ozym

qhua948:
@ozym I don't have write access

PR #797, opened 2020-03-19
update microphone response

ozym:
The `auto.go` file is generated automatically from the yaml files.

I've split the Q330S/3 and Q330S/6 responses for microphones into two. Assume a single (or 3 independent microphones and no seismometer will use a Q330S/3 and will have pre-gain applied). Assume otherwise a Q330S/6 is used and no pre-gain amplifier is used (will also be the case if a setra is installed).


rumachan:
@ozym Sorry this was lost at the bottom of my email. Based on what you've previously explained to be and what I understand that sounds correct. Thanks

PR #801, opened 2020-04-04
strip decimation delays and corrections

ozym:
@salichon it's not 100% clear how the decimation delays and corrections will be used by various software (e.g. ObsPy).  I think it's safer just to set these to zero, rather than allow the possibility of either one or the other being used.  As these are handled inside the dataloggers there shouldn't be any correction needed (i.e. the delay subtract the correction will be zero).


salichon:
Thanks @ozym 
" decimation delays and corrections"  are related to the decimation filters used in the instrument response
 And they are actually accounted for only internally by the datalogger 
--> the delay subtract the correction will be zero in the response file

PR #806, opened 2020-04-08
NPLY firmware upgrade

CBurton90:


CBurton90:
GeoNet/tickets/issues/6005

PR #816, opened 2020-04-29
Change DART, TD network Decimal to 3

salichon:
Aims at simplifying the DART coordinates


ozym:
@salichon there is a problem which trailing zeros. I'm not sure which direction to go here, we would have to switch to using a string to store the numbers rather than a float, the float would need to be generated from the string at point of use.

PR #817, opened 2020-05-04
Add seismic site and channel naming convention documentation

staylorofford:
Assigned 3 reviewers, but probably just need 2 reviews: one from data and one from platform.


docs/SEISMIC_SITE_NAMING_CONVENTIONS.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Stations, Sites, Location Codes
+
+The designation practices of station, site, and location codes are described in the sister document `SEISMIC_CHANNEL_NAMING_CONVENTIONS.md`.  In short, station codes are designed to convey an idea of where the station is located, site codes denote different sensor installation locations at a station, and location codes describe the type(s) of sensor installed at a site.
+
+A station can have many sites, and a site can have many location codes. Ideally, each site will only have one location code associated with it, but this is often not the case. There may be multiple location codes at a site when:
+1. There are sensors of multiple types installed at a site, e.g. a weak motion sensor and a strong motion sensor.
+1. There are multiple sensors of the same type installed at a site, e.g. two weak motion sensors.
+1. The subtype of a sensor has changed, e.g. a short period sensor has been swapped for a broadband sensor, but both are weak motion sensors.
```
ozym, line 16: There's really no difference between a short-period and a broadband (both weak motion velocity), so no need to change location code.

docs/SEISMIC_SITE_NAMING_CONVENTIONS.md
```diff
@@ -0,0 +1,29 @@
...
+
+It should be noted that the choice of having multiple location codes at a site is arbitrary: it is equally as possible to have one site per location code as it is to have one site for all location codes. The former case has the advantage of being able to provide the coordinates of each site as those of the given sensor rather than those of one of (or an average of) the sensors at the site and correspondingly is easier to produce high quality metadata for. As such this is the preferred of the two options. However, the legacy of decisions made prior to the creation of clear conventions muddies our metadata in this respect.
+
+## Naming Conventions When Moving or Changing Sensors
+
+There are 3 cases of how site code and location code change at a station following the movement or changing of a sensor:
+1. If the sensor changes to one that is not of the same type or subtype, a new location code will be made for the sensor installation, e.g. if a short period seismometer with location code 10 was swapped for a broadband seismometer the new location code would be 11.
```
ozym, line 23: The example should be swapped to changing from short period to strong motion.

staylorofford:
Thanks @ozym, I have made those changes.

I think regarding the borehole case you are right: the installation details would differ, although if by more than 1 m I suppose the location code could change via the elevation.

Multiple sensors in a borehole should be OK, we'd just have multiple location codes as in the case of multiple sensors in a vault, no?

PR #820, opened 2020-05-20
Add new Titans

samgeonet:
I think we can remove Titan 0390 as I believe it was used as a test device in 2016 which was then sent back to the manufacturer. Also let me know if you want to rename the devices to Titan SMA which is more accurate to what they are.


ozym:
The added Titan "assets" have an extra "," on the end due to adding the asset number.

ozym:
@samgeonet just read your earlier comment. You should also change the one we looked at in WEL if that was an SMA as well.

samgeonet:
@ozym changes made as requested

install/recorders.csv
```diff
@@ -1261,7 +1261,7 @@ Kinemetrics,QDR,QDR,0189,904F,20,270,0,0,2000-08-04T00:00:00Z,2014-07-09T23:00:0
 Kinemetrics,QDR,QDR,0190,924H,20,218,0,0,2000-07-07T00:00:00Z,2014-07-15T00:00:00Z
 Kinemetrics,QDR,QDR,0191,904E,20,270,0,0,2000-08-04T00:00:00Z,2014-07-09T23:00:00Z
 Kinemetrics,QDR,QDR,0192,924J,20,128,0,0,2000-07-07T00:00:00Z,2014-07-15T00:00:00Z
-Nanometrics,Titan,Titan,0390,WEL,22,0,0,0,2016-05-06T00:05:00Z,2016-08-29T03:00:00Z
+Nanometrics,Titan SMA,Titan SMA,0390,WEL,22,0,0,0,2016-05-06T00:05:00Z,2016-08-29T03:00:00Z
```
ozym, line 1264: Could we flip this back to a "Titan" for the sensor, and a "Titan SMA" for the datalogger (impacts the response building).

samgeonet: Done

ozym: We can use it as a template for the next deploys, and if we decide we don't have any data for the evaluation box, we can remove it then.

samgeonet:
@mnaguit @salichon I cannot merge. Can either of you please do so when its all approved?

PR #829, opened 2020-06-01
PNMS open LC 21

mnaguit:
open LC 21 for 50 & 200 hz streams


mnaguit:
> Sorry to be a pain, can we change the height to 10 as that is where the previous installation was. I got the 14.2 from the gps on the CUSP and is likely to be inaccurate.

No worries at all @samgeonet . Is the azimuth correct?

samgeonet:
Yes @mnaguit I have checked with Paul and its the same

mnaguit:
Height changed to 10 @samgeonet . All good.

PR #835, opened 2020-06-08
Add COVZ, URZ borehole instrument orientations derived from compariso…

staylorofford:
…n with surface instrument records.

As per equipment change metadata guidelines, only one reviewer is required to merge.


salichon:
thanks! @staylorofford  
this is a good change - Perhaps a validation process can be enacted by @jristau 
prior/after the PR is merged over a CMT computation or a simple confirmation.

If surface seismometer orientation requires one reviewer only  (direct measure by the techs)
I may suggest borehole ones may require a little additional attention in that prospect (indirect measure)
FYI @JonoHanson

staylorofford:
Very good points @salichon.

I should note here that I did consult with John Ristau regarding the URZ orientation, and our results were in agreement. The orientation given here is in the reference as the one the CTBTO have, the difference between this and what John Ristau has in his moment tensor code is about 90 degrees.

And regarding COVZ, Art Jolly and myself have the same result derived in his case using successive windows of the microseism, and in my case using multiple teleseismic events.

I do agree that such values should be validation. I am not sure how/who would be involved in this, perhaps a good point for our next delta meeting?

PR #836, opened 2020-06-11
WIP: don't merge -- initial example of camera views and lenses

ozym:
This is an initial attempt at slotting in some new data tables related to cameras and lenses.

The current "mount" table remains the same.  This is the equivalent of the "station" table as used in seismic sites, it allows attaching a camera code (related to a place) to a geographical location and time range.

The current "camera" table has been expanded to add a "view" column, the view is defined along the same lines as a seismic site,  it represents the "slot" where a camera can be installed and has information related to where the camera would be pointing, and a description etc. It has a code which is used int the camera image name.

Finally there's a new "lens" table which matches lenses to views, it has a "type" parameter which allows attaching multiple lenses to the same view at the same time (e.g. a Day/Night set, although only one will be used at a time).

This is just a straw man, and is up for discussion (I've just made an unthinking pass through the current cameras.csv file to add a "view" to make it pass the tests and shouldn't necessarily represent real life).

/cc @JonoHanson @elidana


ozym:
I'm maybe not happy with the order of parameters in the view table, probably want the dip/azimuth at the start followed by the description as this is variable length.

install/cameras.csv
```diff
@@ -1,47 +1,47 @@
...
+Mobotix AG,M15,10.18.101.137,CVTP,01,0.5,290,-2.5,0,0,2019-07-08T00:50:00Z,2019-11-27T20:50:00Z,Tai Ping looking at East Ruapehu & South East Ngauruhoe
+Mobotix AG,M15,10.18.101.137,CVTP,01,9.5,290,-2.5,0,0,2019-11-27T21:00:00Z,2020-05-07T00:19:59Z,Tai Ping looking at East Ruapehu & South East Ngauruhoe
+Mobotix AG,M15,10.18.101.33,WIWR,01,0,0,1.5,0,0,2016-10-03T06:14:00Z,9999-01-01T00:00:00Z,White Island West Rim looking into the crater
+Mobotix AG,M15,10.18.168.61,CVWF,01,0,0,0,0,0,2016-06-01T23:40:00Z,9999-01-01T00:00:00Z,White Island Factory looking across the crater floor
+Mobotix AG,M15,10.20.210.120,CVTP,01,9.5,290,-2.5,0,0,2020-05-07T00:20:00Z,9999-01-01T00:00:00Z,Tai Ping looking at East Ruapehu & South East Ngauruhoe
+Mobotix AG,M15,10.20.245.115,TOTM,01,0,0,0,0,0,2018-05-03T00:00:00Z,9999-01-01T00:00:00Z,Karewerewa looking at TeMaari
+Mobotix AG,M15,10.7.46.153,TOTM,01,0,0,0,0,0,2017-10-12T00:40:00Z,2018-05-02T23:55:00Z,
+Mobotix AG,M24M-Secure,10.9.190.163,TCNW,01,0,315,0,0,0,2011-11-28T00:00:00Z,2019-02-24T21:00:00Z,
+Mobotix AG,M24M-Secure,10.9.190.76,TCNE,02,0,45,0,0,0,2011-11-28T00:00:03Z,2019-02-24T21:00:00Z,
```
staylorofford, line 38: How can this be view 02 if view 01 is not in the table?

install/cameras.csv
```diff
@@ -1,47 +1,47 @@
...
+Mobotix AG,M24M-Secure,10.9.246.200,TCEA,03,0,90,0,0,0,2011-11-28T00:00:03Z,2019-02-24T21:00:00Z,
```
staylorofford, line 39: As above.

network/views.csv
```diff
@@ -0,0 +1,3 @@
+Mount,View,Name,Description,Dip,Azimuth,Start Date,End Date
```
staylorofford, line 1: A better order might be: Mount,View,Name,Azimuth,Dip,Start Date, End Date, Description

Name is confusing: what does it mean? Is this the target?

rumachan (dismissed):
View is the equivalent of seismic 'location', which is what we discussed. Looks good.

staylorofford:
Had a well thought out response but the page refreshed...

In short, do we still need targetcodes in our camera site code convention if we have views like this? The targetcode would provide some simplification in that the data type would be available, but that is all: the targets.csv contains which cameras capture which target, and the views.csv has the targets each camera captures too.

install/cameras.csv
ozym: I just made these up so may not be real life, clearly needs a test to check this.

ozym: The other thing is I haven't changed the name of the camera in this instance ( tcne should be something else with 01 02 03 etc)

network/views.csv
ozym: Yeah agree, thinking has changed between headers. Maybe title? or. Caption?

ozym:
The problem is a camera with multiple targets.

ozym:
The idea was a view is about where it's looking, rather than what it's looking at (target) if that makes sense.

ozym:
I must also remember the doas installs which may not have been reflected as well as they need to.

network/views.csv
staylorofford: Or... target?

staylorofford (dismissed):
True, if we consider DOAS instruments then the targetcode makes sense to keep.

OK, all looks good to me then.

install/lenses.csv
```diff
@@ -0,0 +1,3 @@
+Make,Model,Serial,Mount,View,Attachment,Start Date,End Date
```
Camasher, line 1: Make: Mobotix AG
Model: M15-43D *Would need to list each lens type in other repo, could probably remove the D as this just states its as a day lens?*
Serial: 43D-15430
Mount: WHWI *connects to install/camera*
View: 01
Attachment: Day

ozym:
The view part has been merged into the #859 PR.

elidana:
Without entering in the technical details that seems already well ironed out, a couple of minor comments from me related to the documentation:

`install/README.md` file:
1. under "Files", add en entry to mention the newly introduced file `lenses.csv`, 
2. Under the _Camera_ paragraph, add a table explaining the different fields contained in `lenses.csv` 

`docs/sheets/CAMERA.md`
add an outline of the required Delta prerequisites, checks and references

ozym:
@Camasher that's still a question in my head. Do we go for a "generic" lens install or the actual lens (where it has a serial number and we know what it is)?

Currently this is coded as if the lens can be described by a label and left to the downstream user to know what to do (e.g. in this case it is Night or Day) does knowing the exact serial number (i.e. actual lens) help with processing/knowledge about an image?

ozym:
This rebase over #932 once that's merged

Camasher:
> @Camasher that's still a question in my head. Do we go for a "generic" lens install or the actual lens (where it has a serial number and we know what it is)?
> 
> Currently this is coded as if the lens can be described by a label and left to the downstream user to know what to do (e.g. in this case it is Night or Day) does knowing the exact serial number (i.e. actual lens) help with processing/knowledge about an image?

True, the serial number includes a section indicating the lens type (6D237-6066883 is a 6MP Day B237 lens), but for a delta data user, the lens type, rather than the exact serial number is more important. For the SIT side, the serial number is essential

ozym:
I'll close this as we're working through the actual implementation.

PR #838, opened 2020-06-14
fix missing 200hz continuous streams

ozym:
This should update the 200Hz strong motion streams.

It adds the missing 200Hz for GLWS, and switches the 200Hz streams not to be triggered. 

For RCBS it will choose the lower location code (alphabetically) , i.e. 20 vs 21


mnaguit:
Changes are all good for me, thanks @ozym .
Just a note though that for GLWS, the data for 200hz streams started coming in on 2020.040 at LC 21.
But the idea of opening the 200hz consistent with the 50hz is also good.
What do you think @salichon ?

salichon:
I would addume that the  GLWS HN  would have been configured as "triggered" and wasn't at the end of the day, until turned into continuous.

PR #848, opened 2020-06-25
WRAS strong motion is a permanent site, fix network attribution.

elidana:


salichon:
this should be 
`SM | NZ | National strong motion network`
I reckon :)

elidana:
thanks @salichon ! I have updated this pr to use the `SM` network code for WRAS.
can you please re-review?
Thanks!

PR #850, opened 2020-06-29
HBHS recorder swap

mnaguit:
from Basalt SN 559 to Basalt SN 1770


mnaguit:
Hi @salichon . Do the changes herein look fine with you? Thanks!

mnaguit:
Follow up: can you please review? @salichon

PR #856, opened 2020-07-03
Add M16 Cameras

Camasher:


assets/cameras.csv
```diff
@@ -27,6 +27,8 @@ Mobotix AG,M15,10.18.168.61,16016,
 Mobotix AG,M15,10.7.46.153,16028,
 Mobotix AG,M15,10.20.245.115,16033,
 Mobotix AG,M15,10.20.210.120,,
+Mobotix AG,M16,10.26.62.41,12902,
+Mobotix AG,M16,TBA,12903,
```
ozym, line 31: might be best to add this when the serial is known, it will need another PR in any case.

Camasher: @ozym, serial number now added

ozym:
Looks like the asset numbers have been used before?

Camasher:
@ozym ah shouldn't have been, but maybe. I asked @Notashtonkutcher to grab me two numbers, might have been from a used pile

ozym:
Sercel sensor number 642 etc.

Notashtonkutcher:
LGTM

Camasher:
Will need a Delta authorised person to merge @staylorofford

PR #859, opened 2020-07-05
add camera views

ozym:
This update adds a new table called 'view' which allows multiple cameras to be mounted at a single location while using the same code, much like a seismic station code and site code pair.

The current set of camera installs have been added with a blank view code, this should allow for transition as required.

There are plans for followup work to cleanup this current mount entries (de-duplicate information) as well add lens support.


network/views.csv
```diff
@@ -0,0 +1,18 @@
+Mount,View,Label,Dip,Azimuth,Description,Start Date,End Date
```
staylorofford, line 1: So label captures what the camera is seeing, description covers more where the camera is _and_ what it is seeing. From what I see and know this seems quite complete.

PR #863, opened 2020-07-13
Installation of Cameras at DISC Discovery Lodge

Camasher:


staylorofford:
Have to close this while delta undergoes an update - will reopen another PR to capture these changes

PR #864, opened 2020-07-13
MTHZ recorder swap

mnaguit:
from Basalt to TitanSMA


ozym:
Was this related to the previous install? Or just an operational thing?

samgeonet:
@ozym an operational thing. They couldn't easily install true north so they installed south

PR #874, opened 2020-07-31
unified cmg3tb paz values

ozym:
This PR alters the sensor response pole and zero values for the early CMG-3TB sensors installed exclusively at RPZ and URZ.

The current pole/zero values are correct from a calibration sense, however these values need to be normalized using complex value found at the normalisation frequency, rather than the amplitude.  This impacts some processing software and has become a defacto standard.

For reference see [guralp](https://www.guralp.com/howtos/poles-and-zeroes-with-positive-normalisation-factors.shtml)

The sensor models that were purchased later have this adjusted response, it is this value that the earlier models will be set to.

e.g. from the above reference

|Zeros|Poles|Normalisation Factor at 1 Hz|
|------|------|-------------------------------|
|0 | −80.0 | A = 2 304 000
|0 | −160.0 |  
|  | −180.0 |  
|  | −0.00589 + 0.00589j |  
| | −0.00589 – 0.00589j|


ozym:
Have added the CMG-40T updates as well.

In both cases this involved setting the number of zeros to two (0, 0) and replacing the real values poles with three: -80, -160, -180 as per the CMG-3TB above.

salichon:
FYI @mnaguit

salichon:
FYI @mnaguit

salichon:
@JonoHanson bump :)

JonoHanson:
If @salichon , then that satisfies my requirements to merge.

PR #875, opened 2020-08-03
update Fiordland GNSS sites monuments description

elidana:
Hi @CBurton90 , 
I realised that for "Wyatt/Agnew Drilled-Braced" that are shallow (foundation less than 5 m), we are using a different description in the monument.csv file, which is "Shallow Rod / Braced Antenna Mount".

Both monuments have depths of ~3.2 m, so this pull request is updating our database accordingly.

Thanks!

Elisabetta


elidana:
nice! Thanks @KOgeonet !!

PR #876, opened 2020-08-06
GBIT_Tide_Sensor_Change

salichon:
- Change the sensor as mentioned 
- Upon sensor relocation (location code): 
     -->> rounding  the site location 40/41 Lat/lon since  NO site code duplication allowed  ( can t Open  AND close One)


mnaguit:
The PR is all good for me. But before I approve, here's a quick question @salichon . Would it be better to put in the full coordinates rather than rounding it off? Not that it matters a lot, but just checking to be able to do the same for similar tickets.

> Compare sensor location: Former: -36.189045 , 175.48886 | New -36.189678 175.489206
> -->> I will round the site positions to -36.189 175.489

salichon:
> 
> 
> The PR is all good for me. But before I approve, here's a quick question @salichon . Would it be better to put in the full coordinates rather than rounding it off? Not that it matters a lot, but just checking to be able to do the same for similar tickets.
> 
> > Compare sensor location: Former: -36.189045 , 175.48886 | New -36.189678 175.489206
> > -->> I will round the site positions to -36.189 175.489

Hello @mnaguit to me this rounding reflects a most common coordinates between both tide gauges  installations using the same site location codes  (coupled 40/41 tide gauges ) since we can't open/close duplicate site locations (several 40/41's) 
( --> in the spirit of: one site location one set of fixed coordinates (for ever). )

PR #877, opened 2020-08-06
attach trillium 360 to dataloggers

ozym:
The Trillium 360 had been added to the response files as a "Trillium Vault 360" but had never been attached to a datalogger set.

* This PR changes the name of the sensor to "Trillium 360"
* Adds the Trillium 360 to the set of sensors attached to Quanterra dataloggers.


salichon:
Herllo @ozym 
A quick question: Is the `auto.go` change legit in the PR since tyhis file  is created bu the process ? 
Rest is all good :)

ozym:
Yes, the auto.go needs to be there.  It gets generated automatically when the program `go generate` is run in the resp directory.

PR #883, opened 2020-08-16
consolidate druck polynomial responses

ozym:
Before switching to stationxml version 1.1 the polynomial responses have to be consolidated down to a single sensor response. This impacts the druck pressure sensors.  There are now two responses, one for salt water (the default) and one for fresh (i.e. LAND).

There should be no change to the overall response, for reference the processing is configured to use:

depth =  -48543.689320388352 + 0.11573728469682271 * counts (for sea water, depth in 10th mm)
depth =  -50000 + 0.119209289551 * counts (for fresh water, depth in 10th mm)

asssuming Q330 datalogger.

(e.g for freshwater depth = -5 + .0000119209289551 * counts in m).


salichon:
FYI @JonoHanson @mnaguit  and @the_Future_Tidegauge_Crew

PR #890, opened 2020-08-28
LPLS Lake Paringa is Closed 2020

salichon:
Instrumentation back in 2020 / last data from 2015


CBurton90 (dismissed):
LGTC!

mnaguit (changes requested):
Closing date/time for station and site is good (2020-08-05) which corresponds to the removal of recorder on site.
How about for the closing of the streams? Shall we use the date/time that corresponds to the last available data (e.g. 2014-10-27T22:41)?

salichon:
Heya @mnaguit 
I took the last bit of digital file present in the archive that is from  early 2015
if you prefer having it closed over the last recorded data digital  it is all goos yeap. so i make the streams closed at 2014-10-28T00:00:00

mnaguit (approved):

PR #891, opened 2020-08-30
Staylorofford patch 1

staylorofford:
TEMO webcam swap.

I have used timing conventions of the last session ending 1 second after the last data, and the new session starting 1 second before the new data. This is because there are images between these two times that are not taken from the mount + view and are not in the correct format, otherwise both times would be 1 second before the new data.


rumachan (dismissed):
Hadn't realised Cam had to change the camera. Night images are beautiful!
The changes  look good, though I haven't checked the date/time of the transition. I'm assuming you've got that correct. What you've got makes sense.

EDIT
Have now checked date/time and they are good.

staylorofford:
Thanks Steve. If you could can you please check the timing? You may find or feel that what I have chosen is not appropriate.

rumachan:
Sorry, was being lazy, will do.

rumachan:
Sam, have revised review. I think what you chose to do to avoid non-normal images is an obvious way to deal with the situation.You still pass!

staylorofford:
The travis build failed so I have shuffled the lines around. I suspect it wants serial numbers to be ascending at a given site when the camera model is the same. This is interesting as it puts the newer installation ahead of the older one in the file which is unusual. We will see if it resolves now.

staylorofford:
FYI @ozym, was this intended?

ozym:
Yup, without looking, the goal is to make it easier for git to merge rather than for user to find/edit ... (smiley face here).

staylorofford:
OK, @rumachan could you please cast your eye once more over the changes just to check they're still making sense with the line swap and then I will merge.

rumachan:
TOTM example above TEMO has current entry above older ones such its consistent. Think its still clear. Given its a complex CSV file, anyone wanting to use it will be prepared to examine it closely anyway.

staylorofford:
Thanks Steve, can you please reapprove then I can merge.

rumachan (approved):

staylorofford:
This classes as a moderate change, so I do need @CBurton90 or am approved RIM team member to review.

CBurton90 (approved):
Serial looks correct to me

PR #895, opened 2020-09-01
GNSS NETR9 firwmare upgrades

elidana:
firmware 5.22 to 5.45 upgrades for 20 NetR9s


CBurton90 (approved):
Looks good, the review probably takes longer than making the PR now 🤣

PR #901, opened 2020-09-07
DART NZF BPR replace

salichon:
Deploy nr 02 DART Buoy
Site change Only


ozym (commented):
Do we actually that level of precision, i.e. -29.68346667 is down in the mm range?

JonoHanson:
I'd imagine 4 decimal places is more than enough, and as honestly accurate as these measurements can be (assuming GPS measurements, though I guess boats could be pretty accurate)

JonoHanson (changes requested):
Please only include 4 decimal points of precision

CBurton90:
I agree with the above 😉 both lat/long should have a similar level of precision (as Jono suggested 4 decimal places) and consistency. I.e. NZF had previously latitude precision to 6 decimals places and longitude to 4 decimal places and now it has both lat/long to 8 decimal places.

salichon:
Fair enough i went back to my xkcd precision  chart - will  do so sure

ozym (approved):

CBurton90 (approved):

JonoHanson (approved):

PR #906, opened 2020-09-21
GNSS firmware upgrades v5.45

elidana:
Those sites have been upgraded to fw 5.45
LKTA
TAUP
KAIK
DNVK
VGMT
NPLY
WHKT
WAIM
NLSN
LEXA
SCTB
PUKE
AUCK
CHTI
DUND
HAAS
MQZG
PYGR
WARK
WGTN


staylorofford (approved):
Many of the sites have the firmware change occurring within the hour with the change time and the data records are unambigous about when the upgrade happens. Some of the sites don't have this (upgrade starts in an hour or more prior to finish), but the data records the new firmware, and the change time is ~1 minute before the data begin to be recorded. It might be worth checking the last file prior to the outage to see that it has the old firmware, but given the behaviour of the sites capturing the firmware change in the data of that hour and the change time I feel that what I have done is sufficient.

elidana:
Thanks a lot @staylorofford !
In the review process, if the "current hour" is fully recorded with the new firmware, then we generally check the hour before to ensure the firmware was the previous release. so it should be fine

CBurton90 (approved):
Checked all and there is a clear transition between 5.43 and 5.45 (PUKE 5.22) from the previous hour and hour of change. CHTI has a missing file in the preceding hour (06:00:00Z) as fw upgrade was still completing.

PR #910, opened 2020-09-23
GNSS receiver firwmare upgrades

elidana:
GNSS firmware upgraded to 5.45 for the following sites
WHNG
TRNG
WEST
METH
WANG
GLDB
MAHO
MAVL


CBurton90 (approved):
All good!!

PR #912, opened 2020-09-25
GNSS receiver firmware upgrades

elidana:
GNSS receiver firmware to v. 5.45 for the following stations 
BLUF
CORM
GISB
HAMT
HAST
HIKB
HOKI
KTIA
MTJO
WRPA


CBurton90 (approved):
All look great!

PR #915, opened 2020-10-01
seismic sensor orientations: add new column to /install/sensors.csv a…

staylorofford:
…nd add documentation regarding sensor orientation methods to /docs


ozym (commented):
Just a few comments at the moment as code needs to change to get this to pass the tests.  

We need to get the length of the title and the labels down, github has a fixed width for displaying csv files and it's best if it can fit without an horizontal slider.

e.g.  *Orientation Method* could perhaps simply be *Method*, and the 
coherency methods could be "Remote Coherency" and "Local Coherency" ...

Likewise, compass should be "Compass" and "APS" could be expanded into what it represents (not entirely obvious from a quick google search, Azimuth Pointing System?.

Once these are in place we'll need to add these as key words to the sensor code, then write tests to accept the values only.

staylorofford:
@ozym thanks for the review, I have updated the PR appropriately.

salichon (approved):
This is looking good so far

staylorofford:
As per https://github.com/GeoNet/delta/pull/931, I have reduced this PR to just documentation.

staylorofford:
@ozym or @salichon, can I get a review just to allow merge - no changes from last time besides removing the install/sensors.csv from this PR

ozym (approved):
We can look at the actual labels to use in the columns next time round.

PR #919, opened 2020-10-05
GNSS KORO receiver swap and firmware upgrade

elidana:
I found a receiver swap that went untracked back in January. 
This pull request fix the delta entry (plus I just upgraded the firmware to latest version for NetR9).


staylorofford (approved):
Changes check out, but did you also update the antenna type and number in the reciever config.? `runpkr00` gives `! Alert ! 'KORO202010052200a.tgd': change of antenna type from 'TRM57971.00' to 'TRM55971.00' ~ 2020 Oct  5 22:33:00.000
! Alert ! 'KORO202010052200a.tgd': change of antenna number from '41031186' to '30203658' ~ 2020 Oct  5 22:33:00.000`

elidana:
Thanks for checking @staylorofford ! 
No I haven't. I think one that the receiver that was undeployed had the wrong antenna configuration set on it. I did check records in our gallery and cannot find any evidence of an antenna change. 

The GNSS vertical time series is not showing any obvious jump either

staylorofford:
The raw files capture a change in the antenna configuration when the new receiver was installed, but if that is just because the old one had the wrong configuration then all is well!

elidana:
bump @CBurton90 , can I please have your thumbsup for this one , from the "RIM team" perspective? thanks!

staylorofford:
@CBurton90 please can you review this ASAP as we need to merge or close all open PRs to allow for an update to delta

staylorofford:
@elidana Conrad will be away until next week, so if we want to clear this one we might have to find a third way.

elidana:
let's call in the big dogs: @peanutty , can we please have your thumbsup here?

staylorofford:
He is also away. @timjmcdougall is the only one. Tim, are you comfortable reviewing this from RIM team perspective?

timjmcdougall (approved):

elidana:
thanks @timjmcdougall !

PR #920, opened 2020-10-06
updated delta tests

ozym:
This is a reworking of the standard tests to be a little more consistent.

There's an update to the GNSS streams check, but apart from that it's all just moving code around.

The new layout is to have a table of functions that get called with sets of lists, these are then called in a loop by each test.

To see them in action, run `go test -v` in the `tests` directory (this will also show warnings).

### DIFF
--- a/tests/antennas_test.go
+++ b/tests/antennas_test.go
@@ -6,135 +6,194 @@ import (
 	"github.com/GeoNet/delta/meta"
 )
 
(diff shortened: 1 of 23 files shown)


elidana (approved):
I have not fully reviewed the code, but tried to mock some error to check the tests and all looks good to me, thanks @ozym !

PR #922, opened 2020-10-06
fix some of time mismatches captured with delta tests

elidana:
Hi,

this fix some of the mismatches found with tests in https://github.com/GeoNet/delta/pull/920

not all of them. I have quickly checked against presence (or not) of data in the archive before applying those changes.


salichon (commented):
some are related to known mismatches

salichon (approved):
Good as

PR #933, opened 2020-10-28
Add EQR120 response 200HZ strong motion

salichon:
Team work on EQR120 New response 
integration of the recorder

### DIFF
--- a/resp/auto.go
+++ b/resp/auto.go
@@ -75,6 +75,13 @@ var DataloggerModels map[string]DataloggerModel = map[string]DataloggerModel{
 		Manufacturer: "GNS Science",
 		Vendor:       "",
 	},
+	"EQR120": {
+		Name:         "EQR120",
+		Type:         "Strong Motion Recorder",
+		Description:  "Eqr120",
+		Manufacturer: "Canterbury Seismic Instruments",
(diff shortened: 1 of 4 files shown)


elidana:
can we please add EQR120 model also for sensor description? This way might be more clear when we add new equipment in /install/recorders.csv

elidana (changes requested):
please add EQR120 sensor model as well (not just datalogger)

CBurton90 (approved):
As I was there it all looks good. The only thing that may want to be thought of is adding the sensor type D to the EQR120 i.e. calling it in the cusp.yaml file `EQR120D` like the conventions of the CUSP sensors before it.

mnaguit (approved):
_.yaml_ files look consistent with the already registered equipment. Knowing little about these files, I'm assuming added details are not case sensitive (e.g. EQR120 vs Eqr120 in description). For the _.go_ file, from what I understood this is created automatically.

elidana (approved):

salichon:
@CBurton90 - i will merge this PR for the EQR120 strong motion CSI 
This should ease the deploys 
FYI @JonoHanson @peanutty

PR #936, opened 2020-11-04
 GLOS is a StrongMotion network SM ot a NM 

salichon:
Most likely typo error on the subnetwork naming.
Cf `network/netwroks.csv`


mnaguit:
Was it NM before as part of Nelson and Marlborough regional seismic network?

ozym:
> Was it NM before as part of Nelson and Marlborough regional seismic network?

If it was it would probably have a Z on the end ...

mnaguit:
Noted @ozym . Then most probably SM is the most suitable network code.

mnaguit (approved):

PR #938, opened 2020-11-05
DART Paroscientific sensor additions 

CBurton90:


ozym (approved):

salichon (commented):
Hi this sounds good 
Note Until we get the direct reach back data
from  the Engineering data messages on the buoys 
We ll rely on the Deployments paper  logs information
 provided as it comes from NIWA,  by NEMA
So this information may change with time. 
(perhaps when instrument is recovered)

Engineering data are providing:
 Paros serial And BPR firmware and serial.

In this PR I deem and consider we re informing there only the Paros serial

ozym:
I think this should be added to the sensors list rather than recorders.

The recorder part could be the BRP package as a whole.

CBurton90:
That was my original thinking @ozym but please see

https://github.com/GeoNet/tickets/issues/7167#issuecomment-706815608

Happy to put them in the sensors.csv but from what I could find there is not a lot of information on the datalogger (digitiser/computer) for the 4G DART systems.  But am happy to put something generic in the dataloggers.csv i.e. `DART 4G 16 bit computer` until we have more specifics?

ozym:
That was my thinking too, I'm not 100% about the engineering to go one way or the other.  The problem is calling something a sensor when its a recorder (or vice versa)

CBurton90:
Changes made @salichon @ozym as I cannot find much information on how the analogue measurements are digitised I have made some assumptions from this document. The entries in dataloggers.csv should serve as a placeholder regardless until we have more information. 

[DBCP-30-item-7.10-ITP.pdf](https://github.com/GeoNet/delta/files/5497578/DBCP-30-item-7.10-ITP.pdf)

salichon:
@conrad at Paros web they have some PDFs with information,  maight help :) 
http://paroscientific.com/pdf/Realtime%20Tsunami%20-%20Broadband%20Vibrating%20Quartz%20Pressure%20Sensors%20for%20Tsunameter%20and%20other%20Apps.pdf

salichon:
@conrad  
it s a digiquartz sensor  rigth - so the ouput should be directly digital output right ?  
(No digitizer to convert  something volt to Pressure/Temp  for instance,  just direct input to a computer to output Water depth ? )

from https://www.researchgate.net/publication/4129988_Broadband_vibrating_quartz_pressure_sensors_for_tsunameter_and_other_oceanographic_applications

salichon (changes requested):
hello @CBurton90  from the Deployments logs stor\ed on the wiki page 
I got  `Paroscientific,BPR Subsystem, 2-190706,,DART 4G` 
instead of  190706 
Just a minor correction 
cheers

CBurton90:
Changes made (good spot) @salichon !

How does this look to both of you @ozym @JonoHanson ?

salichon (approved):
The BPR serial Number would be finally firmlyu confirmed
upon specific remote query  OR when recovering the instrument. 
Serial  numbers will help to match the instrument coefficient to get  
the water height as a function of the temperature and pression  (i reckon)

salichon:
Hello @ozym @JonoHanson 
Any comments \ thumbs up about this PR ?  

thanks 
Jerome

PR #942, opened 2020-11-12
BOWS -  new EQR120 recorder CUSP 

salichon:
- EQR120  tags the datalogger and the sensor
- First real life deploy of new CSI
- will validate future ones

- Please check the azimuth of the components - ;) ta!


CBurton90 (changes requested):
I thought the orientation with respect to Y was 9 deg for the EQR120? 280 may indeed be correct for the old CUSP but as the sensor may have been housed with a 90deg offset (see the ticket) : )

mnaguit (changes requested):
Thanks @salichon .
While the orientation is getting sorted, in streams.csv --should it be that another entry is introduced to separate the triggered 200hz from the continuous 200hz?

salichon:
Orientation  Change made @mnaguit @CBurton90

salichon:
@mnaguit once travis is done would you review this a last time
thanks!

mnaguit (approved):
Many thanks @salichon . 👍

PR #945, opened 2020-11-15
update old DNVK GNSS session set to 5s sampling rate

elidana:
DNVK was set to track at 5s sampling rate for a few months in 2003.

This pull request modify the GNSS receiver session file to still keep this information as an additional session entry, but now consider the session that has sampling rate set at 30s as a single entry.


megankortink (approved):
Lgtm

PR #954, opened 2020-11-24
Update WIWR metadata showing camera bracket slip events

Camasher:
2018 slip: https://github.com/GeoNet/tickets/issues/4792

2020 slip: https://github.com/GeoNet/tickets/issues/7537

Estimated historic and new dip/azimuth from 3d models.
Changed height to a negative to indicate above ground


Thomas-Benson:
Delta 27/3/25: Similar to here: https://github.com/GeoNet/delta/pull/1126#issuecomment-2756067137
Probably don't need to add these events to delta at this level of detail

PR #960, opened 2020-12-03
DART Stations/sites parameters

salichon:
- Elevation to ~9999~ 0
- Depth from N1 average on first deployed stable day
- Site code to match Pressure/height site codes


ozym (commented):
I'd suggest not having an elevation, i.e. ",," rather than ",9999," ... that way it's known there's no elevation ...

ozym:
@salichon or perhaps, it should simply be 0, as this is the height of the ocean (in theory), we assume for these sites that elevation means the height of the water surface which is then above the "height" or "depth" of water. 

That is, distinguish between a water sensor in crater lake, vs on the ocean floor.

salichon:
okay i was not  sure this latest would pass the test - this was my first  guess ...

ozym (approved):
nice.

salichon:
@JonoHanson @CBurton90 bump :)

CBurton90 (approved):

JonoHanson (approved):

PR #965, opened 2020-12-08
GNSS correcting end and start time stamp for NLSN

megankortink:
Editing start and end time stamps for NLSN


elidana (approved):
looks good!

PR #967, opened 2020-12-08
GNSS changing start and end dates for TGHR

megankortink:
On the 17th and 18th there was no data after the first hour of the 17th while receiver was 4435237647, when switched to 4445240300 data started on the 19th. Have ended first receiver on end of the 18th and started the new one on the 19th. Please check.


elidana (approved):
looks good!

PR #973, opened 2020-12-15
AMBC new EQR recorder (swap cusp 3D)

salichon:
Orientation is kept at 266 degrees
unless mentioned otherwise


mnaguit (dismissed):

CBurton90 (dismissed):
This looks okay to me but please see previous discussions about orientation 

GeoNet/delta/pull/942

I believe that because of the way the EQR120 fits in the grey box with respect to the old CUSP there is a 90 degree offset between their Y axis. The field team can confirm when they are back. But you can either merge it for now and note it may need a correction or we can wait.

CBurton90:
Orientation does not need updating. It should be the same as the old CUSP.

salichon:
heya 

just a poke to set the review on again after resolving conflict
Merci @mnaguit  :)

mnaguit (approved):

PR #975, opened 2020-12-16
LSRC recorder swap

mnaguit:
from CUSP3D 42042 to EQR120 46502


salichon (approved):
thanks @mnaguit lgt

CBurton90 (approved):
Brendon has informed me the orientation of all EQR120's are the same as the old CUSP so no need to update.

PR #977, opened 2020-12-21
GNSS TURI Receiver Swap

megankortink:
TURI Receiver Swap. Note that previous receiver number in ticket should be 5207K82248 not 5122K76160.


elidana (approved):
looks pretty good

KOgeonet (approved):
LGTM

PR #982, opened 2021-01-16
updated pod export for nanometrics sensors

ozym:
This may help with building dataless miniseed with newer sensors.


salichon (approved):
Thanks!

PR #986, opened 2021-01-21
CMWZ strong motion Not axial since 2016 and Continuous 200Hz in 2020

salichon:
Metadata correction:
- ENZ component since 2016
- 200Hz continuous since 2020


CBurton90 (approved):
Cool!

PR #990, opened 2021-01-27
GNSS PNUI Receiver Swap

megankortink:
Swap from NetR9 to Alloy


CBurton90 (approved):

elidana (approved):
looks good! Thanks @megankortink , and thanks heaps for fixing the old firmware entry as well :)

PR #995, opened 2021-02-03
update consistency checks

ozym:
While checking the camera code I noticed a gap in the consistency checks. This fixes the gap related to sorting files and the associated files that fell in.


salichon:
hi Mark  thanks,
what would be  the "rule" around this sorting? 
-number of characters and numeric ordering of some sorts
i ll try to pay attention to it - to avoid an error on the PR tests after the change

salichon (approved):

ozym:
@salichon this is mostly for the few files that seemed to have slipped through the net. There shouldn't be any impact, the ordering will be as expected. The tricky part I think is serial numbers which are not numbers ...

PR #996, opened 2021-02-04
Swap DISC discoverylodge M12 camera to M16

Camasher:
Theres definitely a few things in this repo that need to be cleaned up....


Camasher:
Just need to add correct timing and orientation details

rumachan (commented):
Are the dates correct? You are adding the M16 on Saturday morning?

It's a bit hard to judge as your changes and Mark's are overlapping. I assume the reference to Mountain Air is because Mark will be adding a change to correct this?

Camasher:
Have updated times to match true last and first images. HAve updated tilt and orientation.

staylorofford:
Superceded by https://github.com/GeoNet/delta/pull/1064

PR #1000, opened 2021-02-14
Firmware upgrade for BOWS & 5 Chch stations

mnaguit:
- first delta PR to include firmware upgrade for EQR120 (and later for other seismic equipment)

Are you guys happy with the proposed changes? Kindly let me know if I have to modify the details.
Thanks!


CBurton90 (changes requested):
I'm not sure if they need the notes of `Fix GPS timing 1 sec`  as it was a seedlink timestamping issue primarily. Also the firmware is `m001.03.06` , I would have thought that line could have been solely that rather than `EQR120+m 001.03.06` . 

I'll let @ozym confirm though as that's just how I see it.

ozym:
Yup, no 1 second fix here. Just the first upgrade (fixes ping etc) but this doesn't need to be recorded.

ozym:
Also, no need for the site codes. This should be obvious from the other files and is one more thing that could cause confusion (i.e. we don't test it).

mnaguit:
Thanks @ozym & @CBurton90 .
I modified the notes (to indicate first upgrade) & version (to remove EQR120). For your review.

ozym:
It actually doesn't need any comment (we abuse the notes column too much).

mnaguit:
Hahaha. Will remove the comment now. =)

mnaguit:
Another round of your review, as per changes made.
Thanks @ozym * @CBurton90

ozym (approved):
Looks fine, I'll leave it up to @CBurton90 to confirm what the firmware name is.

CBurton90 (approved):
Looks good!

ozym:
Only other thought was attempting to add in the firmware version prior to this.

mnaguit:
> Only other thought was attempting to add in the firmware version prior to this.

Add in the notes, before I merge? @ozym

ozym:
No, merge away. There should be another PR to find original firmware if we can.

CBurton90:
> No, merge away. There should be another PR to find original firmware if we can.

@ozym @mnaguit for reference previous fw was `m001.03.03` but I am unsure if this is consistent across all units recently purchased. In the RFP we tested `m001.03.01` I believe?

PR #1002, opened 2021-02-18
add lrdcp pilot sites to stations.csv https://github.com/GeoNet/tickets/issues/8047

rumachan:
Next stage in LRDCP meta data.


ozym:
@rumachan line 953 (and friends) is missing a depth entry (basically needs an empty value after the elevation). See the line above it.

rumachan:
Bugger, thanks Mark, I'd missed that. That depth entry is for general information about the location of a sensor, not for specific use in 'data reduction'? Some sensors at some sites are a few metres below a water surface, but not all sensors. I want to use another file for the detailed 'data reduction' depths (where appropriate for a specific sensor). So I'll go with the empty value you suggested.

rumachan:
@ozym I've corrected and pushed again. IS that all I need to do before you can review again? Sorry, not too experienced here.

ozym (approved):
@rumachan yes, I look at the stations entry being equivalent to a dot on the map (good for plots and general details) whereas the sites entry will, along with the installed sensor, give the nitty gritty of where the actual measurements are taken.   In the stations file the depth is really there for things like ocean bottom sensors, or dart etc.

PR #994, opened 2021-02-03
Add new Mobotix M16 camera assets

Camasher:

PR #1008, opened 2021-03-03
RIMK.01 Raoul Mount Moumoukai M15-M16 Camera Swap

Camasher:


Camasher:
I have left site as RIMM and not RIMK untill this data is converted across all at the same time, but have added view code (01) to both pre and post swap. I have added the dip as measured before swap, which was not previously captured. Removal time and install time set as 1 second after and prior to (respectively) the first image in the FTP archive

Camasher:
This PR will continue to fail until this PR (https://github.com/GeoNet/delta/pull/994) is merged which contains the entry of these assets in to the system

staylorofford:
Superseded by https://github.com/GeoNet/delta/pull/1064

PR #1009, opened 2021-03-08
Recorder VRZ start/end to allow for Response with streams mods

salichon:
In order to allow for last response to be created with changes in 
axial and continuous streams, 
the recorder.csv is updated accordingly


mnaguit (approved):
Note: similar opening & closing times as the changes previously in `streams.csv`, to solve the issue in creating the response information

PR #1012, opened 2021-03-09
GNSS PAKI Receiver Swap

megankortink:
Could you please double check the firmware entry? Not sure if I needed to add a new line or not. As alloy was previously installed as KAIK for 3 days before being swapped. Also in receiver file added entry below KAIK (going off time order - is this correct?)


elidana (approved):
looks good, thanks Megan! And firmware entry does not need updating as you already did. Thanks :)

CBurton90 (approved):

PR #1013, opened 2021-03-15
add feature table

ozym:
@rumachan still a work in progress but the bones are there.

The `network/features.csv` file can list detailed descriptions about where a site is located, it shouldn't be a general notes column but something that has meaning in a position, geological, or sampling setting. If we need general notes we'll try and address that using another mechanism.

Once this looks like it will work for you, we'll make a request to add it to the code base.


rumachan (approved):
So you chose to include start and end dates. Probably easier to have them now than add later. Those dates would be same as used in network/sites.csv I assume.

Happy with that, especially like the focus on position.geology, sampling setting.

ozym:
Have updated to move into environment.

rumachan:
@ozym This is still waiting on review? I've reviewed earlier so can't again? I will if that is possible and it unblocks this.

rumachan (approved):

network/README.md
```diff
@@ -4,6 +4,7 @@ _Lists the geographical and physical characteristics of collection points._
 
 ### FILES ###
 
+* `features.csv` - Physical site location descriptions.
```
elidana, line 7: I think this one needs to be removed from here, as the `features.csv` is now under `/environment `?

meta/testdata/features.csv
```diff
@@ -0,0 +1,3 @@
+Station,Location,Description,Start Date,End Date
+CNZ,12,Over the rainbow,2014-05-16T00:00:15Z,9999-01-01T00:00:00Z
```
elidana, line 2: best place ever, love it!

environment/README.md
```diff
@@ -20,6 +22,16 @@ _Lists the geological and physical environment details of collection points._
 | _Amplitude_ | Analysis Amplitude, in cm
 | _Lag_ | Analysis Phase Lag, in degrees
 
+#### _FEATURES_ ####
+
+| Field | Description |
+| --- | --- |
+| _Station_ | Code used to uniquely identify Recording _Station_.
```
elidana, line 29: does this table also need to add a column for the site or will we rely on the `network/sites.csv` and `network/stations.csv` for that?

elidana (commented):
Hi,

I like it a lot, thanks Mark and Rich! 

### use of site/station
I still don't have enough understanding of the station-site-location-channel, so the following comment might be wrong. 
I wonder if we need to add a definition of "site" and how it is related to the definition of "station" in the `https://github.com/GeoNet/delta/blob/main/docs/SEISMIC_CHANNEL_NAMING_CONVENTIONS.md` document, just to help users to clarify the relation. 
or just create a specific "naming convention doc" that can capture the relations for Low Rate Data Collection Platforms.

### timestamp in environment/features.csv
relatively to the date, we can adopt the same logic as the gnss visibility table, where the start date of the "feature" is the same as the "first deployment of the site" as described in the `network/...csv` file, and the end date is open.
Then, the feature will change only if there is a major change in the "environment" around the site/station, (e.g. lake level sensibly dropped and now the sensor is exposed in the air) and in that case we can add a new entry with the date of the change. 
would that make sense?

network/README.md
staylorofford: agree

environment/README.md
staylorofford: technically the "site" doesn't exist, only the station and the locations at the station.

staylorofford (changes requested):
I agree with Elisabetta that we could add some further clarity to our descriptive docs around station-site-location relations. For the time being I think we should keep these docs separate as our data and users remain separate, but in the future we could surely blend them given all they share.

Regarding feature end dates, I think we should avoid putting in anything that we would need to change as a result of things beyond our control, such as in the example you give. We can describe what feature is *intended* to be sampled, and the data can speak for itself as to whether that has occurred (I imagined in your example the data would be clearly anomalous).

What do you think?

rumachan:
Can I suggest we go with what we have on this. Without any records in `features.csv` its a little difficult to see exactly how it's going to work; though it's intent is clear. I have a dozen LRDCP sites I can add and I hope that will clarify things a lot.

elidana:
Hi Steve, happy for this one to progress and be merged.
The main change requested by me and Sam is to move the "features" entry in the `environment` folder readme (and not in the `network` one). It's a minor thing, we can also merge this pr and apply the change with another pull request, as long as we don't forget.

In terms of clarifying what a site is, happy for that to be discussed separately, and potentially generalised to all sensors handled by the stations-streams-... tables.

In terms of timestamp, happy to do whatever the expert think is appropriate. I like Sam's point on including things that are within our control, but sometime things are not and at least we have a place to record "external events".

ozym:
Have updated the networks readme to remove the features line.

staylorofford (approved):

elidana (approved):
thanks for updating the Readme!
Looks good, we will pull offline the other comments

PR #1016, opened 2021-03-18
Update README.md

megankortink:
Added Gauges and Constituent information from delta/network readme.md. Please check the 'Crex Tag'. Also added something under 'Checks' please add if feel like it needs more details.


ozym (approved):
looks good

PR #1019, opened 2021-03-25
SACS recorder swap

mnaguit:
from CUSP3D to EQR120

Note: from the ticket, the expected orientation is 15 degrees. I'm still double checking if this is the case, or the orientation is the same as the original set up, which was 105 degrees. Will correct this PR accordingly, as needed.


mnaguit:
Azimuth has been changed to 15degrees, as confirmed by @MilesRyan .

CBurton90 (commented):
Hate to be a pain but is it possible to add an entry in delta/install/firmware.csv for this unit. It is running m001.03.06. Cheers @mnaguit

mnaguit:
Good idea @CBurton90 . Thanks. Not a pain at all. Much easier to put all changes in one PR...here it goes. Please review again.

salichon (approved):
sweet as thx @mnaguit

CBurton90 (approved):
LGTM

PR #1023, opened 2021-03-29
updated gain for trillium compact ph-2

ozym:
based on http://ds.iris.edu/NRL/sensors/nanometrics/nanometrics_trillium_sensors.html

the model number TC20-PH2 ordered with a sensitivity of 750 V/m/s should have a response via

http://ds.iris.edu/NRL/sensors/nanometrics/RESP.XX.NS124..BHZ.TrilliumCompact120.120.754

the last line gives the sensitivity of 754.3 V/m/s


ozym:
Pretty sure the confusion will be around the differences between a second generation and a second post hole model, that is "trillium compact 120 generation 2" vs "trillium compact 120 PH-2"

CBurton90 (approved):

ozym:
@mnaguit @salichon we've been doing some more investigation into the centaur compact post hole responses, it looks like we've used the wrong response for the "centaur compact posthole PH-2"  we've only got two of these but with only one currently installed.

From nanometrics:

The sensitivity of all Trillium Compact 120 sensors is 754.3 V/m/s.  
The sensitivity of the Trillium 120 Posthole is 1200.7. 

The PAZ responses are the same for the two sensors, see the first comment for links.

mnaguit (approved):

salichon (approved):
- [ ]

PR #1024, opened 2021-03-29
Garmin clock upgrades

elidana:
Hi,

here is a first attempt for a keeping track of Garmin GPS antennas firmware upgrades to correct clock errors of a number of seismic recorders.


salichon (approved):
so there is the version  400.74.1 from 2019
 and subversion(patch)  400.74.1 (GPS18X4.40) with another  GPS fix from  2021  ?
right ?

CBurton90 (commented):
Might be worth copying the Garmin release notes to our notes with:

`Fixed GPS week number rollover`

Apart from that it looks ok

mnaguit (approved):
Many thanks @elidana for creating the script that would speed up the updates on firmware metadata. 👍 
On a side note, same query raised by @salichon -- to understand the sub-version or 2 versions for the same patch?

elidana:
thanks @CBurton90 ! I will update the comment accordingly.

@salichon the `400.74.1` is the CUSP firmware, whereas the `400.74.1 (GPS18X4.40)` is the attached GPS antenna firwmare upgrade. I think Mark was suggesting to provide the information that way, but happy to change to anything that end users can "easily" understand

elidana:
Thanks @mnaguit ! my understanding is that the new firmware is not a patch of the firwmare installed on the recorder, but rather the one of the GPS antenna attached to it to provide timing info. Since we are not recording in delta metadata about the GPS antennas attached to seismic recorders at all, the proposal here is to include this information in the recorder metadata.

elidana:
Garmin release note now included.
might be worth checking the timestamps are correct too, I have used what I saw in the ticket, but if there are gaps in the data we can refine the timestamp (if needed)

CBurton90 (approved):
Double checked the times and it all looks good. As long as we are all happy with the naming (GPS18X4.40) then LGTM!

ozym:
@elidana There's a conflict that needs sorting, something must have changed while you were working on this.

elidana:
Hi @ozym I can't see the conflict from here or rebasing this branch to main.. Maybe the branch update fixed it?

elidana:
this one was blessed during the last delta meeting, merging

salichon (approved):
all sweet as

PR #1025, opened 2021-03-29
 Not for merging -  CentauR conf - bb posthole

salichon:
Dummy branch for Centaur + Posthole test 

---------------
for testing only 
and use  files as necessary
Relevant bits and pieces  are into the **resp/** directory 

**To be removed**


ozym:
Have been also working on #1026

Need to get the equipment deployed so that I can test the stationxml generation (independent of merging the response code).

ozym:
outdated

PR #1027, opened 2021-03-31
WEL Centaur Integration Test 

CBurton90:


ozym:
@CBurton90 you will likely need to add a "Place" value to the datalogger install (say "Integration" maybe) and then use this in a "connections.csv" entry which connects WEL.13 to the Dominion Observatory - "Integration").

ozym (approved):
I think the Integration label should be split, but I don't think it matters in this context.

salichon (commented):
in datalogger.csv I m not sure if  changin the name of the place to 
"Wellington Dominion Observatory Integration" is relevant (since it is the same)
anyhow can do atm 
perhaps using the _"role"_ field of the _datalogger.csv_ file is more relevant to  follow up with  or inform this specific installation
Same applies to the _"note"_ field into the _sensor.csv_   i reckon

Anyway so far so good, 
i don t have the "aprove" super power for some reason : 


- [ ] **but I approve anyway**  

(ha .. it s already merged ... )]

@ozym @CBurton90

PR #1029, opened 2021-04-01
PEEC recorder swap

mnaguit:
from CUSP3D to EQR120, with firmware detail for EQR120


CBurton90 (approved):
As long as azimuth is correct LGTM!

salichon (approved):
Hi 
the HN data looks actually to have some (not so good data)  UTC ~00:00 
over the 31/04/2021 Julian day 090
the instruiment itself was still present till that day-ish 
and not removed by the  _2021-01-04T02:58:00Z_

I would tend to recommend to fit the closing date to the actual  day when  the instrument was removed 
(so that it gives some sort of  practical ease 
"hypothetically" if data are retrieved afterwards close to the actual removal they can be matched to the metadata properly.

PR #1035, opened 2021-04-13
Firmware upgrade for EQR120s

mnaguit:
from v6 to v8, for 12 stations
Note: EBFS has no firmware details prior to issuance of this PR. Metadata for EBFS is upgraded herein as well.


elidana (approved):
looks good!

CBurton90:
I've double checked the timestamps as well and looks good. Thanks a lot @mnaguit

CBurton90 (approved):

PR #1045, opened 2021-04-27
KPOC recorder swap

mnaguit:
Re-deployment of EQR120 for upgrade in voltage:
KPOC from SN 46500 to SN 46523 (no asset number supplied)


CBurton90 (approved):

salichon (approved):
okidoki
thanks

PR #1046, opened 2021-04-27
SMTC recorder swap

mnaguit:
Re-deployment of EQR120 for upgrade in voltage:
From SN 46504 to SN 46531 (no asset number supplied)


salichon (approved):
oui oui  tout de bon 
thanks :)

CBurton90 (approved):

PR #1047, opened 2021-04-27
SACS recorder swap

mnaguit:
EQR120 swap from 12V to 24V
From SN 46505 to SN 46526 (no asset number supplied)
Firmware version .08


salichon (approved):
“Верно” ->  “Correct”

CBurton90 (approved):

PR #1048, opened 2021-04-27
LSRC recorder swap

mnaguit:
From EQR120 SN 46502 to EQR120 SN 46522 (no asset number provided)
Firmware v.08


salichon (approved):
Napakahusay!

CBurton90 (approved):

PR #1050, opened 2021-04-28
HVSC recorder swap

mnaguit:
EQR120 voltage upgrade
From 12V (SN 46508) to 24V (SN46529 --no asset number)
Firmware v.08


salichon (approved):
Ganz güet!

CBurton90 (approved):

PR #1054, opened 2021-04-29
PEEC recorder swap

mnaguit:
EQR120 voltage upgrade
From 12V (SN 46498) to 24V (SN46527)
Firmware v.08


salichon (approved):
Hǎo de

Xièxiè

CBurton90 (approved):

PR #1055, opened 2021-04-29
DORC recorder swap

mnaguit:
EQR120 voltage upgrade
From 12V (SN 46493) to 24V (SN 46532)
Firmware v.08


CBurton90 (approved):

salichon (approved):
Dobre (polish) = good

PR #1059, opened 2021-05-06
ORPS recorder swap

mnaguit:
From Basalt 1245 to Basalt 615
From triggered to continuous data streaming


CBurton90 (approved):

salichon (approved):
Swap like4like no response issue over time
All good 
Thanks!

PR #1062, opened 2021-05-07
Revert "moving travis api token from web to yaml"

yanjunding:
Reverts GeoNet/delta#1061


ardrigh (approved):

ardrigh:
The two scripts currently using the API token are using the value from `# geonetci_api_token` so this extra line in the file and the ENV in the web console settings were redundant.

PR #1064, opened 2021-05-10
Updated camera metadata with changes over period of camera convention…

staylorofford:
… review

FYI @Camasher


rumachan (approved):

staylorofford:
@ozym hints on why the Travis build fails for this?

ozym:
looks like it just needs to be in alphabetical order,  guessing you need to treat "10.26.194.151" like a string rather than a bunch of numbers.

ozym (approved):

PR #1078, opened 2021-05-26
updated NBRT constituents to match SUMT

ozym:
Adding the NBRT constituents from SUMT will allow an initial offset display. When the actual values kick in we'll upload those.


megankortink (approved):
Lgtm, checked values to SUMT, the 'x' after the LINZ number is a good idea. I imagine we'll remove this one SUMT is decommissioned?

PR #1081, opened 2021-05-31
OTAT Tsunami gauge sensor and datalogger swap

megankortink:
Hi all, 
One question about the name of the site, in the datalogger and connections file, the site was called 'Dunedin Tsunami', I changed the network/stations.csv file to match (as was previously just 'Dunedin'). Do we want to change it completely to 'Observation Point'?


mnaguit (changes requested):
Changes look good.
Only we have this new metadata we need to maintain for the firmware.
Kindly place the details for Centaur under `install/firmware.csv` version 4.5.20

As for the station name, `Observation Point` is better than `Tsunami`.

megankortink:
After a few changes, have changed the stations file back to just 'Dunedin' after discussing with Elisabetta as this what is being used on the geonet webpage and I shouldn't have changed it in the first place. Also using Observation Point will be confusing, so haven't touched that. 
Have added the firmware for this station, it is also possible to see that this hasn't happened for other centuar installs, shall we leave this??? @CBurton90 @mnaguit @salichon

mnaguit:
I'm okey incorporating the changes in firmware (`install/firmware.csv`) in this PR, if that makes it a lot easier (as opposed to issuing separate PRs per station).

- Centaur 7976 (WLGT)
- Centaur 7977 (NBRT)
- Centaur 7985 (GIST)

megankortink:
Have added firmware details for three other centaurs (WLGT, NBRT and GIST)

mnaguit (approved):

CBurton90 (approved):

PR #1090, opened 2021-06-02
add new lrdcp sites, NA002, TO003, TO063, TOUEM

rumachan:
Adding all the recently installed site's meta data


elidana:
Hi @rumachan , I might be wrong, but is TOUEM-002 missing from the sensors file?

rumachan:
TOUEM is missing as it breaks the check rules for the files, as its the same sensor. Need to check with Mark about that.

PR #1096, opened 2021-06-03
lrdcp pilot streams connections

rumachan:
Complete streams and connections files for LRDCP pilot sites. Dataloggers needed one change so site name was consistent throughout.


mnaguit (changes requested):
All good with consistency in station names and opening times.
OT001 at LC 05 is open in connections but not streams. Is there a need to match the opening in the connections with that of streams?

rumachan:
Thanks Muriel, added the missing OT001 line

mnaguit (approved):

PR #1098, opened 2021-06-07
update dart sample rate labels

ozym:
Said one thing, did another. This is to update the dart sampling rate codes from VTZ, UTZ -> UTZ, WTZ.


salichon (approved):
sweet thx

megankortink (approved):
Awesome, thanks!

PR #1100, opened 2021-06-08
MLZ sensor swap

mnaguit:
from Guralp CMG-3ESP SN T3B46 to Nanometrics TH120-1 Horizon SN 804
For LC 10 HH(ENZ)


CBurton90 (approved):

salichon (approved):
legit :)

PR #1101, opened 2021-06-10
remove spaces from dart filter names

ozym:
To avoid one possible problem when loading the derived stationxml.


megankortink (approved):
LGTM

salichon (approved):
thanks ! 

If all good as a "fix"   for the DART
will need to implement the same apporach for the "tsunamiTide" gauges yaml files.

PR #1109, opened 2021-06-15
FOXAG connections, streams

rumachan:
@elidana 
This is something I forgot from the pull request we sorted earlier, apologies

Forgot to update install/connections.csv and install/sensors.csv.

Had to change 7 files and only did 5!

Shouldn't be any errors as ran test.sh before pull request.


elidana (approved):
looks good!

PR #1110, opened 2021-06-15
update README file in network folder

elidana:
Hi,

this pr updates the `network/README.md` file now that the tidal constituents have been moved to the `environment` folder. The deleted details where already present in the correct README file.

I have also noticed that the recently introduced `network/views.csv` file was not described in the README, so am adding this here (@rumachan , can you please check that the content is correct?).

Thanks!

### DIFF
--- a/network/README.md
+++ b/network/README.md
@@ -10,8 +10,9 @@ _Lists the geographical and physical characteristics of collection points._
 * `networks.csv` - Grouping of GNSS Marks & Recording Stations.
 * `sites.csv` - Specific Observation Points at a Specific Recording Station.
 * `stations.csv` - Location for a Group of Recording Sites.
+* `views.csv` - Network Camera View codes.
 
-#### _MARK_ ####
+#### _MARKS_ ####
 
 * _Mark_ - Code used to uniquely identify GNSS _Mark_.
 * _Network_ - Code used to group marks together by project or operator.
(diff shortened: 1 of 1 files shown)


rumachan (changes requested):
There is an issue with the `Label` column in views.csv. This is not an issue with the changes proposed in this PR, but in the content of that column in the views.csv. To be honest, I am not sure what that column is suppose to contain.

The `Mount` is the location of the camera. The `Description` contains where the camera is, plus what it is looking at. Does `Label` function like `Name` in `stations.csv`, in which case it should be the name associated with the `Mount` code? At the moment, `Description` is a mix of where it is, and what it is looking at. I think we need a small group to chat about this to decide (@ozym 
 @staylorofford @rumachan @elidana). Once that is done, this PR can either be approved or a small change made. Then, we can change the content of `Label` so it is consistent for all `Mount`.

elidana:
Hi @rumachan ,

when I was looking at the `view.csv` file, my understanding is that the `Label` was used to provide a general and non free format description of the geographical feature the camera is pointing to, whereas the `description` field is more free format and detailed. 
The name used in `stations.csv`, is instead the name of the location where the camera is installed.

Again, that's my current understanding, let's have a chat on that and, as you said, finalize the README and then eventually review the content of the views.csv if needed.

Thanks!!

rumachan:
If `Label` is the view of a camera, is there anywhere where I can find that the `DISC` code corresponds to a site called `Discovery Lodge`? While it isn't necessary to know what `DISC` stands for, we have this for other platforms. @elidana

elidana:
yes, you can find it in the `mounts.csv` file
```
[elidana@akatea network]$ grep DISC *
mounts.csv:DISC,,Discovery Lodge,-39.156143224,175.491281443,9999,WGS84,Images of Ruapehu and Ngauruhoe.,2020-07-28T02:20:00Z,9999-01-01T00:00:00Z
views.csv:DISC,01,Ruapehu North,155,,0,Images of Mount Ruapehu from the volcano camera situated at Discovery Lodge.,2020-07-28T02:20:00Z,9999-01-01T00:00:00Z
views.csv:DISC,02,Ngauruhoe,90,,0,Images of Mount Ngauruhoe from the volcano camera situated at Discovery Lodge.,2020-07-28T23:40:00Z,9999-01-01T00:00:00Z
```

staylorofford:
We really need @ozym to feed in on this. Is the label column necessary or just a bit of legacy from the old camera conventions? If label and description are more or less the same, I think we should just have one to keep things simple.

elidana:
Hi @staylorofford and @rumachan , @ozym shaded light on the `Label` question: 

the `vies.csv/label` column is used to add an identification label to the camera images published on the GeoNet website volcano camera images page (https://www.geonet.org.nz/volcano/cameras).

rumachan:
So many files, I missed `mounts.csv`, sorry.

That answers some of my questions in https://github.com/GeoNet/tickets/issues/8817

elidana:
- [ ] include details on improved camera views as in comment https://github.com/GeoNet/delta/pull/1153#issuecomment-905149414

rumachan (approved):

staylorofford (commented):
What happened to all the gauge information?

PR #1114, opened 2021-06-28
add a doas equipment type

ozym:
The name `doas` doesn't easily roll off the tongue, and neither does the plural `doases`.  I recall we also have a "minidoas" and a "flyspec" in our history of equipment so I've dropped the scanning bit from the name to allow for this type of equipment.  This may also be gas type specific so maybe in the feature it may need expanding out for new equipment types.

Currently it looks the same as a Camera except for a missing Notes field (the use of free-form notes was strongly discouraged, either notes become regularised in the form of a column, or they can live elsewhere such as the environment "visibility.csv" file).


ozym:
I haven't updated and docs etc,, as that will be the next step.

rumachan (approved):
Happy about doases.csv. Understand about the Notes field. Getting away from freef orm notes is a good idea. Just need to make sure I follow my own good intentions.

I can't comment on the go code.

PR #1117, opened 2021-06-29
update simple docs and add empty doas files

ozym:
This PR gets things ready for installing a scan doas equipment. It has empty files for assets and installs, plus tests based on the camera tests to make sure the data is internally consistent.

### DIFF
--- a/assets/README.md
+++ b/assets/README.md
@@ -6,7 +6,7 @@ Each file represents an equipment type and is listed with the following attribut
 
 * _Make_ - Equipment manufacturer or commercial brand name.
 * _Model_ - Equipment model name or general identification label.
-* _Serial_ - Equipment serial number, needs to be unique within its set of equipment make &amp; model.
+* _Serial_ - Equipment serial number, needs to be unique within its set of equipment make and model.
 * _Number_ - Equipment asset number, this is optional and is used primarily for cross-checking
 * _Notes_ - Any extra details which may be relevant to the piece of equipment.
 
@@ -26,4 +26,4 @@ Each file represents an equipment type and is listed with the following attribut
 
 Pre-commit checks will be made on these files to ensure:
 * No duplicated asset numbers - if given these need to be globally unique
-* No duplicated serial numbers - these need to be unique within each equipment make &amp; model set.
+* No duplicated serial numbers - these need to be unique within each equipment make and model set.
--- a/assets/README.md
+++ b/assets/README.md
@@ -6,7 +6,7 @@ Each file represents an equipment type and is listed with the following attribut
 
 * _Make_ - Equipment manufacturer or commercial brand name.
 * _Model_ - Equipment model name or general identification label.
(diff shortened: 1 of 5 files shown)


assets/README.md
```diff
@@ -6,7 +6,7 @@ Each file represents an equipment type and is listed with the following attribut
 
 * _Make_ - Equipment manufacturer or commercial brand name.
 * _Model_ - Equipment model name or general identification label.
-* _Serial_ - Equipment serial number, needs to be unique within its set of equipment make & model.
+* _Serial_ - Equipment serial number, needs to be unique within its set of equipment make &amp; model.
```
elidana, line 9: minor comment: can we just use `and` instead of `&`, just to make the readme "text-friendly"?

assets/README.md
```diff
@@ -25,4 +26,4 @@ Each file represents an equipment type and is listed with the following attribut
 
 Pre-commit checks will be made on these files to ensure:
 * No duplicated asset numbers - if given these need to be globally unique
-* No duplicated serial numbers - these need to be unique within each equipment make & model set.
+* No duplicated serial numbers - these need to be unique within each equipment make &amp; model set.
```
elidana, line 29: same suggestion as above (`and` instead of `&`)

rumachan (dismissed):
Unable to review go code, otherwise all good.

elidana (approved):
Hi Mark, that looks good to me.  I did some quit "mocked testing" and the tests are nicely capturing errors in metadata entries. Haven't done anything extensive, but looks very good to me.

Just a minor suggestion inline on the use of plain words.

assets/README.md
rumachan: Agreed `and` is tidier.

assets/README.md
ozym: > Hi Mark, that looks good to me. I did some quick "mocked testing" and the tests are nicely capturing errors in metadata entries. Haven't done anything extensive, but looks very good to me.
> 
> Just a minor suggestion inline on the use of plain words.

The problem is that markdown isn't really the same as plain text. Github hides some of the complexities, try typing in "&lt;hello&gt;" into a markdown file.  Using `&amp;` is the safest way, likewise `&lt;` and `&gt;` are needed.  I would consider swapping `&amp;` for "and"  however.

rumachan (approved):
& to and, all good

rumachan:
Took the liberty of merging this, hope that was okay.

assets/README.md
elidana: oh yes totally agree that we shall use markdown proper syntax when we need. That one was mainly because there was an easy alternative, but, as I said, only a very minor and not necessary change! That happens when everything else is perfect! ;-)

elidana: ha ha, nice one with the "<hello>" , just tried it!! and now I want to know if there are easter eggs... ;-P

PR #1120, opened 2021-07-06
WCZ weak-sensor swap

mnaguit:
from CMG-ESP (SN: T3G69)
to Nanometrics TH120-1 Horizon (SN: 814)

For LC 10, HH & LH (ENZ).


CBurton90 (approved):

salichon (approved):
Bonjour kia Ora!
Thanks 
all good

PR #1121, opened 2021-07-08
update features tests

ozym:
The reality of actual data, fixed a couple of bugs (!= vs ==) and variable overwrite.


rumachan (approved):
As far as I'm able to say, look fine.

PR #1124, opened 2021-07-14
WKZ LC change from 20 to 21 

mnaguit:
One of those strong motion sites that require backdating to the right location code (from 20 to 21), prior to archive correction.


ozym (approved):

salichon (approved):
Heloo All this cosnistent to the sensors closed  in 2017
`Kinemetrics Inc.,FBA-ES-T,1439,WKZ,20,... 2017-06-28T20:20:00Z `
👍

PR #1126, opened 2021-07-20
Add WIWR camera movements to Delta

Camasher:
In progress - not completed yet

Capturing camera movements

2018/2020 https://github.com/GeoNet/delta/pull/954

2021 - https://github.com/GeoNet/tickets/issues/9056


ozym:
@Camasher is this likely to be progressed?  As it's quite old (the PR that is) it may be easier to transcribe any changes into a new PR which has been reset to main.

Thomas-Benson:
Delta 27/3/25:
We don't have a way to represent the kind of slipping that is being discussed here
It's likely that anyone systematically using this data will be able to account for movement of the camera.
We would need our own systematic review of camera images to completely add this data, and is likely not worth the effort.
View codes anticipate changes in position, and instead focus on the key feature(s) in the image.
https://github.com/GeoNet/delta/blob/main/docs/CAMERA_METADATA_CONVENTIONS_AND_IMAGE_INFORMATION.md

PR #1131, opened 2021-07-26
Configuration update - Nanometrics Centaur and Lennartz LE3DliteMkIII 

CBurton90:


CBurton90:
Please merge once approved (if correct) as I have no permissions in delta @ozym @salichon

salichon (dismissed):
Hello @CBurton90 
thanks for this 
I reckon that is the beginnning of this implementation  ;) ( Perhaps Would you rise a ticket to have room for dicussion beyond that single PR?)
- needs to be tuned toward short period instrument  (not broaband)

ozym (changes requested):
You will need to process this to get the auto.go file which is the bit which will make the change.

Also, this should be for short period streams rather than broadband, e.g.

EH rather than HH.

We also don't run 1 Hz on the short periods so this needs to be removed.

ozym (approved):
I'll do the auto.go PR once you merge this.

salichon:
May need to confirm the frequency  1Hz in place of the 15 Hz in the datalogger response built

PR #1134, opened 2021-07-29
Centaur + Lennartz LE3DliteMkIII - change 1Hz > 15Hz

CBurton90:


salichon (approved):
So now **Centaur** datalogger can be combined with any **Lennartz LE-3D** short period sensors (and only these SP atm ;) ) on our shelves.
thanks!

PR #1136, opened 2021-07-30
initial scandoas data

rumachan:
Add initial configuration for two new scanDOAS on Tongariro

assets/doases.csv
install/doases.csv
network/mounts.csv
network/views.csv


ozym (approved):
looks okay, proof will be in the pudding

PR #1132, opened 2021-07-26
Configuration update - Nanometrics Centaur and Kinemetrics EpiSensor FBA-ES-T configuration

CBurton90:

PR #1137, opened 2021-08-01
EpiSensor+Centau CTR Auto.go script output from PR#1132

salichon:
Go Script output for new combo 
Bound to https://github.com/GeoNet/delta/pull/1132


CBurton90 (commented):
Hmm it this entry behaving as though it is a strong motion recorder?

Tried to hunt down a Quanterra datalogger Connected to Strong Motion Sensor (so I can compare Q330 to FBA-ES-T) but I cant see an entry as such for that either.

salichon:
@CBurton90 i ll go through the stationxml creation to check that the full response is created.
thanks for pointing this :)

ozym:
That doesn't look right, I'd say the original yaml files are bad.

ozym (changes requested):
need to fix underlying yaml problems first.

CBurton90:
yaml file fixed in the other PR, may need to run this again @salichon

salichon:
@ozym @CBurton90  auto.go (re) generated as per #1132 PR corrections

- stationxml  generated  OK --> BSWZ  has FBA Sensor coupled with CTR4-6S (checked on HNZ
 20) 
 this is looking good (to me ; - you ll have the last word )

ozym (approved):

CBurton90 (approved):
yup looks legit!

PR #1140, opened 2021-08-03
DART Adding Metadata for NZD, NZJ, NZK and NZL

megankortink:
Initial DART entry into delta, made additions for NZD, NZJ, NZK and NZL. Have used the BPR coordinates from the beta-data (pre-approved from this ticket: https://github.com/GeoNet/tickets/issues/7321), depths from values decided on in the tickets for the BPRs.


ozym (approved):
looks good to me although I didn't check the dates. we can adjust this and the depths as needed.

salichon (approved):
- consitent with the wiki  entries

PR #1141, opened 2021-08-03
L4C-3D and Centaur combination 

CBurton90:


salichon (commented):
Kia Ora! thanks for this
I d say we don't need the L4C to be included in that PR 
They are no longer used since the late 2000 ... with  of course ... :) the
exception of ...
`Sercel Inc. | L4C | WTAZ_AKN | WTAZ | 10 | 0 |   | 0 | 5 | 0 | 0 | 0 | 0 | 1995-01-10T16:30:00Z | 9999-01-01T00:00:00Z`
mounted with a Q330
So we can inlude this almost retired instrument in our PR or not. (depending of its planned future for the next decade :))
Up to the `network dev` maybe ? 
Cheers! 
(and L4C  are one component afaik)

resp/responses/configurations/shortperiod.yaml
```diff
@@ -144,6 +144,18 @@ response:
 
   Nanometrics Dataloggers Connected to Short Period Sensors:
     sensors:
+    - sensors:
+      - L4C
```
ozym, line 148: Unless we plan on plugging a 1D sensor into a centaur, I'd not bother with this section.

ozym (changes requested):

resp/responses/configurations/shortperiod.yaml
ozym: Ah, just spotted @salichon bet me too the punch!

CBurton90: @salichon @ozym consider it gone!

ozym:
still needs the `auto.go` to be updated.

ozym (approved):

salichon:
will approve and merge afetra stationxml  generation locally  :)

salichon (approved):
BSWZ CTR4 + LC4-3D Paired into the xml  
great!

PR #1142, opened 2021-08-04
update view test to allow for multiple installs

ozym:
PR to allow for re-installation of cameras with a previous view.


rumachan (approved):
This looks like it should do what is required, I'm sure you've tested it.

staylorofford (approved):
yes

PR #1143, opened 2021-08-04
Adding DART BPR details for NZD, NZJ, NZK, and NZL

megankortink:
Adding BPR details, used the BPR SN from the deployment logs as we used this last time instead of the Paros SN from the N$9 files. Also note NZJ, NZK, and NZL have FW versions 2.08 (other buoys have 2.07). Used dates from wiki and beta-data readme.


salichon (commented):
Amongst  the 3 DART  Deployments, 
  The serial Numbers are looking like a mix between the  BPR serial numbers from the logs and the  Paros Serial numbers from the N$9
It may have implication to identify them properly and get their specific coefficients right upon retrieval f the high rate continuous data set.
It can always be amended.
Which standard are we using
 now that we have more information  compared to before .. SN logs or SN N$9 ?  (or both..)

ozym (approved):
cool, we can update the serial numbers to match what we want to note.

megankortink:
Will merge this to put something into delta and then we can modify the SN's once we have consensus. I feel like it will most likely be what ever value we have for all of the buoys, whether that is the Paros SN from the N$9's or the deployment logs. I've tried to use the deployment log SN's that were recorded in the wiki but from a discussion from Jerome it sounds like we may have not had deployment logs for each buoy.

PR #1147, opened 2021-08-10
DCZ weak motion sensor swap

mnaguit:
from Guralp SN T3G80 to Trillium Horizon SN 812
For LC 10 with HH & LH (ENZ) streams


salichon (approved):
sensor swap  all good :)

CBurton90 (approved):

PR #1153, opened 2021-08-25
Revised view labels

rumachan:
This is the revised view labels: Pretty much all camera view labels have been changed and made more consistent

### DIFF
--- a/network/views.csv
+++ b/network/views.csv
@@ -1,30 +1,28 @@
 Mount,View,Label,Azimuth,Method,Dip,Description,Start Date,End Date
-CHTB,01,Christchurch East,90,,0,Images of Christchurch towards the east,2011-11-28T00:00:03Z,2019-02-24T21:00:00Z
-CHTB,02,Christchurch North East,45,,0,Images of Christchurch towards the north-east,2011-11-28T00:00:03Z,2019-02-24T21:00:00Z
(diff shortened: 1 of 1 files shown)


elidana (approved):
Hi @rumachan ,

I like (a lot!) those new labels, and I think this is a very good improvement.

One minor comment, but no need to change or use it if you don't agree.

I notice that there is still some (much less than before) "inconsistencies" across labels.
- majority are indicating direction of view, but does not specify a point
- when camera is located "far away", then a geographical location is provided (New Plymouth or Whakatane)
- for a couple of those (i.e. RIMK, WINR and WIWR) a specific mention to the crater is made

do you think there is a way to "harmonize" those descriptions? Alternatively, we can add some words around this choice in the readme file

rumachan:
@elidana I realise there are still some inconsistencies, but I think what we have at the moment is the clearest we can, while still allowing for different situations.

Where we say `from New Plymouth` or `from Whakatane` it is because the location of the camera is a well known locality. In other cases, the camera location is on an unnamed hill, or the end of a road, or a motel no one knows. In those cases I think using `from Some Big Hill` isn't going to tell end users much as all, so I went for `from south` or similar.

For the Whakaari/White Island cameras and Raoul Island we are looking at specific features (as the cameras are close by) so I felt the direction of view wasn't that important, and naming the main features in the view was. 

It's always a bit of a line call. I'd prefer to revise the readme file if that's okay.

elidana:
totally fine by me, and thanks for the explanation!! I'll resume the old pr of the README file and we can add those details there.

PR #1155, opened 2021-08-31
VRZ Strong Legacy 2014 loccode and Comp name correction

salichon:
***legacy change** 
Relates to archive data correction back in the past 
comes along with an AWS goenet archive correction

Correct for the component orientation  EN to 12 (60 deg azimuth)
and 
for loc code 20 to 21 
from 2014 to  2019


mnaguit (changes requested):
- is there a need to modify `install/recorders.csv` to match the changes in LC from 20 to 21?
- Just a comment:  I thought it's a bit of uncommon set up for this station, to have the triggered 200 hz from 2003 to 2014, without the 50 hz continuous.

mnaguit (changes requested):
Also, in `install/streams.csv `, are we changing the components from non-axial (ENZ) to axial (12Z) for LC21?

salichon:
I updated this PITA PR inline with your  recommendations :) thanks!

mnaguit (approved):
LGTM
Thanks for the changes @salichon

PR #1157, opened 2021-09-03
LTZ Strong loccode legacy change 20 to 21

salichon:
from 2013.296 to 2019.304 loccode is 21


mnaguit (changes requested):
- in `network/sites.csv`: start time should be 2013-10-23/05:05 and end time should be 9999-01-01/00:00
- `install/recorders.csv` needs updating, to align the start-end times with changes in LC from 20 to 21
- Just a comment: so if the changes in delta is related to archive correction, Travis does not return any error?

salichon:
@mnaguit oopsy daisy! 
corrected ! 
it passed the test.sh changes too 
maybe something to track further ;) . 
human review is the best  hey !

mnaguit (approved):
Looks good @salichon 👍

PR #1163, opened 2021-09-09
MQZ strong Legacy loccode change 20 to 21

salichon:
In addtion to site code legacy change, 
LN 1Hz channel connections  misinformed
Site 20 connections misinformed


mnaguit (approved):
Looks good. 
I just find it weird that there is 1 hz from 1998 to 2017, but 50hz just started streaming on 2009.
Well it looks like its a 1hz-200hz combo from 1998 to ~ 2009, until 50hz was turned on.

mnaguit (changes requested):
Going back to that, if LC20 is already closed on `2016-02-22/20:00`, then that means LC20 for 1hz should also be closed on that date/time?

salichon:
I ve been through the raw seimic data for checking:

1- The actual change seems to have occurred on `End: 2016.54 08:00` `Start 2016.55 04:00`  
resp 23/02  and 24/02  instead of the 22/02 (delta information  is  not accurate) 
image
2- The LN 1Hz configuration as apposed to other sites did not stop on that date it goes along till 2017
Hence indeed not only a change in dates but also a change in lcation code for some 2016/2017 LN data 

3- With regards to the significant  inversion/swap of the coordinates for the period of time (technical reasons i suppose)
 2019 - Now  as `ETNA 2,100564,MQZ,21,270,,0,0,2019-06-17T23:20:00Z,9999-01-01T00:00:00Z`   as opposed to previous i d recommend a another (hence How `1,2 ` is more representative than having the `N  component` being  `West` and `E component` being `South`) - sigh -  Perhaps in another tickets for this change  what do yout think  .. )? 
    Although is this relevant to  perform  @mnaguit @staylorofford @elidana ? 


----------------------------------------------------------------

Anyhow i gonna deal with points 1 and 2 in this PR will see for the PITA rest of this later upon discussion.

salichon:
Split from above 

3- With regards to the significant  inversion/swap of the coordinates for the period of time (technical reasons i suppose)
 2019 - Now  as `ETNA 2,100564,MQZ,21,270,,0,0,2019-06-17T23:20:00Z,9999-01-01T00:00:00Z`   as opposed to previous i d recommend a another (hence How `1,2 ` is more representative than having the `N  component` being  `West` and `E component` being `South`) - sigh -  Perhaps in another tickets for this change  what do yout think  .. )? 
    Although is this relevant to  perform  @mnaguit @staylorofford @elidana ?

salichon:
@mnaguit updated the PR according to more relevant dates and added LN 1Hz site code 21.
It willl be reflected into the data change PR

mnaguit (approved):
Looks good @salichon 
Thanks for considering my comments :)

mnaguit:
Sure Sir @salichon

PR #1186, opened 2021-09-27
add new fox stations

rumachan:
Added new Fox sites. For FOXVF, used a dummy start date, which we can adjust later.


jacksonshanks (approved):
looks good to me

PR #1191, opened 2021-10-18
Tsunami Gauge AUCT new datalogger

megankortink:
New Centaur for Devonport Navel Base. @mnaguit I can assign Jerome to review if you're busy :)


mnaguit (dismissed):
Looks good @megankortink 
Just a comment/query: 0.1 hz is still open for both LC 40 and LC 41 (in `install/streams.csv`). Do we still need this for tsunami gauges?

CBurton90 (approved):

PR #1203, opened 2021-11-01
Trillium 360 and Centaur combination

CBurton90:


ozym (approved):
Looks good from what I can see.

CBurton90:
@ozym @salichon please merge when your happy (I cant)

salichon (commented):
This is lokking goog, though question 
Are we doing `ZNE` component only with the Trillium 360 ? 
Is it taken care of later  on ?

ozym:
@salichon the ZNE part is a little misleading, what's really saying is the channels that would notionally be Z N E are in that order (the downstream code will map Z 1 2 to Z N E and then do the response lookup).

This is indicate things like the strong motion which might be  XYZ  or ENZ etc.  (possibly could have used 012 but that may be even more confusing).

salichon (approved):

PR #1204, opened 2021-11-01
MWZ sensor swap

mnaguit:
From Guralp CMG-3ESP (T3G70) to Trillium Horizon (801)


CBurton90 (approved):

salichon (approved):
(edit) More! than17 yeards old sensor  -good stuff haye

PR #1211, opened 2021-11-09
GBIT Tsunami Gauge Centaur swap

megankortink:
Please double check the serial number for the Q330/6. This was the only one logged for Great Barrier Island. I couldn't find this datalogger that was in the SIT PR and at the top of the ticket. So it may have not been entered into the install/dataloggers file or could have been entered in the ticket incorrectly. 
```
Manufacturer: Kinnemetrics
Model: Q330/6
Serial Number: 6068
```


CBurton90:
You have been stitched up slightly here @megankortink,  the locality for GBIT is Korotiti Bay and there is another site very near with a Q330 at the Navy station so easy enough to mix up

`Kinemetrics | Q330S/6 | 6068 | Korotiti Bay |   | 2017-12-12T02:45:00Z | 9999-01-01T00:00:00Z`

megankortink:
Thanks @CBurton90, I did check Korotiti but I got confused with the ordering of the file and couldn't find the Q330S/6 6068 entry as was looking under Q330/6 (without the 'S').

CBurton90 (approved):

mnaguit:
> Thanks @CBurton90, I did check Korotiti but I got confused with the ordering of the file and couldn't find the Q330S/6 6068 entry as was looking under Q330/6 (without the 'S').

I usually search using serial number (same as you, I get confused too with the models). =)
I also get confused with station name (e.g. stations.csv vs dataloggers.csv) . They're not the same sometimes.

mnaguit (approved):

megankortink:
I know @mnaguit, got a tad confused with this one! Thank you both for quite approval!

PR #1212, opened 2021-11-10
DSZ sensor swap

mnaguit:
from Guralp SN: T3G71 to Trillium Horizon SN: 802
for LC 10, HH(ENZ)


CBurton90 (approved):

salichon (approved):
Great  rretirement of a 17 years old BB sensor Good s*--t !

PR #1215, opened 2021-11-23
add feature aspect

ozym:
This PR adds an "aspect" column to the features table. It's not in use yet so will have no impact on operational systems.


rumachan (approved):
All good

PR #1219, opened 2021-11-28
GVZ recorder swap

mnaguit:
- from Basalt 1765 to TitanSMA 001708
- for LC 20 HN-ENZ
- TitanSMA firmware 4.5.22
- Note (BKZ firmware was also updated --with TitanSMA previously noted as SN 001261 in the ticket)


salichon (approved):
the continous infor amtion in 2019 is weird - but not crititcal 

GVZ | 20 | 200 | false | false | false | 2019-06-18T00:25:00Z | 2019-11-13T01:54:00Z
-- | -- | -- | -- | -- | -- | -- | --

CBurton90 (approved):

PR #1221, opened 2021-11-29
WestMelton_geomag

merijnthornton:
Adding West Melton devices into geomag


elidana:
ehi @merijnthornton , I have created a new pull request () as this was missing some of the entries (and using SBAM records, instead of EYWM) https://github.com/GeoNet/delta/pull/1223

PR #1119, opened 2021-07-05
add geomag proof of concept

ozym:
This PR is a concept implementation for getting geomag metadata into delta.

Most of the metadata can be handled by existing tables (csv files) as long as a GSM magnetometer is considered to be a `recorder` (that is, the sensor and the digitiser are integral to each other) and a FGE fluxgate can be considered an installed `sensor` and a deployed `datalogger`.  This will need the appropriate entries in `assets`, `connections`, and `streams` files.

To handle the per-component gain settings I have introduced a `calibrations` table. This essentially maps a deployed sensor's scale and bias over a period range per component (often called a `pin` number).  The idea is that the output value will be equal to `a * V + b` where `a` is the scale, `b` is the bias, and the input will be `V` volts.

This can easily be extended to other equipment (e.g. pressure sensors, thermometers) as required, or it could be used where a calibration of a seismic sensor has been made and individual component scale values are wanted rather than the generic model scale.

I'm not 100% convinced on the name, it implies a lab measurement (which is fine for pressure sensors and seismometers etc.) but for the fluxgate it also represents features related to the installation (the bias values are used to zero the `y` component) and need to be measured when installed.

To build a StationXML response the datalogger and sensor parts need to be joined with the `calibration` details per component, some standard assumptions would be needed (i.e. map x, y, z to components 0,1,2) and the connections/streams files.

PR #1223, opened 2021-11-30
add metadata for EYWM geomag station

elidana:
A first attempt to add metadata for West Melton Geomagnetic observatory.

The file `install/adjustment.csv` is a mocked attempt to add site specific fluxgate "calibration" values (as suggested in https://github.com/GeoNet/delta/pull/1119/)

Real offset and scale values need to be derived (would appreciate some help there on how to do that), based on fluxgate sensor calibration (added here for reference, not directly use in code), and biases estimated during the installation.

@merijnthornton this one, if ok, will supersede #1221


ozym:
couple of comments first if we go with adjustments then the FGE model names need to be changed to match their "nominal" gains, i.e. each sensor model with a different sensitivity needs to be called something different so that the "nominal" value can be looked up, and then any adjustments applied to it can be found. As opposed to the values in the adjustments being the absolute values (and they can have individual gains). This could be used if  you actually wanted to put in the calibration values for a seismic sensor (say) and not rely on the nominal value (which is where my name for calibration came from, but that's not quite right either).

- The values used in the "adjustments" are dummy, i.e.  123.4  and 1234.567  (I'll need to  track down the real numbers).
- There's no need to add a recorder (location 51) to the connections file as by definition the recorder has been given the location already (51) so that it doesn't need to look it up via the connections file.

It will need code changes to do the validation as needed but I can add this before this gets merged, but we're getting closer.

ozym:
For the fluxgate
```c
obs[i] = p->coil * p->polarity * (f->volts[i] / p->resolution + s->step * (double) p->bias) + p->polarity * p->offset;
```
For the temperatures

```c
sensor_temp= f->volts[i] * p->gain - 273
```

This is an example for Scott Base (only one to hand) from old system.
```
  <fluxgate code="scotbase" sensor="257" driver="299" step="0.003922" scale="1280" model="Q330/3">
    <pin no="0" stream_id="SBZ/50-LF3/NZ" tag="x" resolution="29.5" coil="38296.0" bias="0x4f" e="-0.2" polarity="1.0" offset="0"/>
    <pin no="1" stream_id="SBZ/50-LF4/NZ" tag="y" resolution="29.5" coil="38367.0" bias="0x00" e="-0.4" polarity="1.0" offset="0"/>
    <pin no="2" stream_id="SBZ/50-LF5/NZ" tag="z" resolution="29.5" coil="38023.0" bias="0xc1" e="0.1" polarity="-1.0" offset="38000"/>
    <pin no="3" stream_id="SBZ/50-LF0/NZ" sensor="y" gain="200" e="0.7"/>
    <pin no="4" stream_id="SBZ/50-LF1/NZ" driver="y" gain="200" e="-0.7"/>
  </fluxgate>
```

ozym:
In this case the adjustments are related to the difference between using a coil of 38000, the bias, and a potential offset with polarity changes thrown in for good measure.

elidana:
thanks @ozym !!!
I will
- revert name to sensitivity instead of adjustment 
- are you happy to have the sensor lab calibration value in the `calibration` folder?
- remove the 51 channel entry

elidana:
Hi @ozym , the suggested changes are now applied.

Tomorrow (hopefully) @merijnthornton and I will try to use those formulas to update the dummy numbers I've used here for EYWM , and have real ones.

I'm keen to know (and change) anything that still does not makes sense

ozym:
@elidana just looking now, the setting for 51 in connections is not needed.  But there is a missing installed recorders for the GSM

elidana:
thanks @ozym just to confirm, the channels we are going to use are:
- 50 is the proton (GSM)
- 51 is the fluxgate (FGE)
?

elidana:
@ozym , hopefully I got it right this time.

to the very little I understand we should have
50 for the fluxgate and 51 for the proton.

Can you please let me know how this look now and if ok I'll squash all commits, rebase and then we can hopefully merge?

ozym:
@elidana that looks good, few things we can tidy later (docs etc).

Yes, 50 is FGE and 51 is GSM.

I'll approve after you squash or rebase etc.

elidana:
thanks @ozym !!
How does it look now? if you're happy, please approve and I'll merge! (it's passing the local test)

ozym (approved):
all good for a first go.

PR #1224, opened 2021-12-01
extend features to add sublocation and property

ozym:
This helps map dataloggers with sensors that record multiple parameters.


rumachan (approved):
Thanks Mark

PR #1225, opened 2021-12-02
CTZ Seismic Connections fix

salichon:
- CTZ datalogger strong and BB fix fro response creation
- CTZ strong LN 1Hz stopped (useless)


mnaguit (approved):
looks good. hope it clears the bug!

PR #1226, opened 2021-12-03
Kinemetrics Obsidian 4X response correction

CBurton90:
Ideally we should also check the decimation/fir filters from:

http://ds.iris.edu/NRL/dataloggers/kinemetrics/rock/RESP.XX.NK205..HHZ.ROCK.1.40.200.C (Casual 200Hz 40Vpp)

http://ds.iris.edu/NRL/dataloggers/kinemetrics/rock/RESP.XX.NK175..BHZ.ROCK.1.40.50.C (Casual 50Hz 40Vpp)

and add them if need be in `delta/resp/responses/filters/kinemetrics.yaml`

We should also check standalone Obsidian Strong Motion recorders as their response is continuing to use the Basalt fir filters/gain in this PR (but maybe they should be using the Obsidian 4X gain at 40Vpp). TBC, I will have a look at existing config Monday!


salichon:
- ideally requires a test and checks (over an event for instance)

CBurton90:
@ozym @salichon just noting the final filter for 50Hz should be `BASALT_B2C-80` to match the IRIS resp rather than `BASALT_B2-80`

This would need a new entry with a gain of `1` though (currently `0.5` for that filter)

All the other earlier stage filters match the IRIS resp, however, they are all listed as false for casual (may need changing to true?)

The frequency we use seems to be `0` in the datalogger response and `1` in the configuration response yet IRIS uses  `2.000000e-02 HZ` for each stage 

Our:
```
Estimated delay (seconds):           
Correction applied (seconds): 
```
are also different to the IRIS resp, although I'm not sure how we derive these (or if it's important enough to worry about)

ozym:
@CBurton90 this is looking pretty good,  I'd suggest you rebase this and updated the gains on the responses for the Obisidian 4x 200 Hz and 50 Hz.
current:

`gain: 4.201680e+05`

needs to be

`gain: 6.7108864e+06`

CBurton90:
> @CBurton90 this is looking pretty good, I'd suggest you rebase this and updated the gains on the responses for the Obisidian 4x 200 Hz and 50 Hz. current:
> 
> `gain: 4.201680e+05`
> 
> needs to be 4 times bigger.
> 
> `gain: 1680672`

@ozym but that's the whole problem, the Obsidian 4X is using the Basalt-200 filter which is
```
BASALT-200:
  - type: a2d
    lookup: A2D
    frequency: 0
    samplerate: 30000
    decimate: 1
    gain: 1.6777216e+06
    scale: 0
    correction: 0
    delay: 0
    inputunits: V
    outputunits: count
```

So that ^^ is assuming 5Vpp right? But the Obsidian 4X is actually 20Vpp therefore should have a gain 4 times less i.e. `4.201680e+05`

Similar to this comment

https://github.com/GeoNet/tickets/issues/9937#issuecomment-978385124

CBurton90:
@ozym @salichon As a refresher I have painfully gone through all the non-casual filters in the 50Hz and 200Hz responses here:

http://ds.iris.edu/NRL/dataloggers/kinemetrics/rock/RESP.XX.NK325..BHZ.ROCK.1.40.50.NonC
http://ds.iris.edu/NRL/dataloggers/kinemetrics/rock/RESP.XX.NK355..HHZ.ROCK.1.40.200.NonC

I can confirm that all filters in delta already existing for the Obsidian4X match exactly the above, therefore only the gain change is needed which this PR has applied.

I think this is ready for you to pull locally and test @salichon

CBurton90:
> @ozym @salichon just noting the final filter for 50Hz should be `BASALT_B2C-80` to match the IRIS resp rather than `BASALT_B2-80`
> 
> This would need a new entry with a gain of `1` though (currently `0.5` for that filter)
> 
> All the other earlier stage filters match the IRIS resp, however, they are all listed as false for casual (may need changing to true?)
> 
> The frequency we use seems to be `0` in the datalogger response and `1` in the configuration response yet IRIS uses `2.000000e-02 HZ` for each stage
> 
> Our:
> 
> ```
> Estimated delay (seconds):           
> Correction applied (seconds): 
> ```
> 
> are also different to the IRIS resp, although I'm not sure how we derive these (or if it's important enough to worry about)

I was wrong here as I should have been looking at non-casual response rather than casual (which changes the filter used)

salichon:
thanks @CBurton90 

- [ ] Check locally created instrument resposne for Obsidian 4X

salichon:
@CBurton90  first step :)  
i locally created a station XML response for RIz and checked for the "sensitivity"
Left is original | Right is the newly created response xml  extract

```
   <Model>Obsidian 4X Datalogger</Mode|              <Model>Obsidian 4X Datalogger</Mod
   <SerialNumber>656</SerialNumber>   |              <SerialNumber>656</SerialNumber>  
    <InstallationDate>2016-09-17T13:00:|              <InstallationDate>2016-09-17T13:00
    </DataLogger>                        |            </DataLogger>                       
    <Response>                           |            <Response>                          
    <InstrumentSensitivity>            |              <InstrumentSensitivity>           
    <Value>1.70934447112192e+06</Valu|                <Value>428087.6205816</Value>  
```

The ratio between the 2 overall sensitivities is `3.99` Close to that expected. Nice :+1: 

I shall use https://github.com/GeoNet/help/issues/85 to process one example event for the next step.

CBurton90:
@salichon Glad it went well, let me know how the next stage goes!

salichon:
@CBurton90 we shall a lat chat about the response.

Before it is merged (still need some final review perhaps)  into Delta (hence all the metadata services), we most likely need
some documentation and comms in order to deal with the tickets and questions floating around these SM instrumentation ..
cheers!
j

salichon (commented):
@CBurton90 
i might need highlights about some of the parameter that are set to 0 as  `gain` and `scale`  for some of the filters. (0 for not used?)  
Also the `frequency` of reference for the gain.
it is more parameter highlight than asking for change   to me :)

CBurton90:
@salichon from my limited understanding if it is set to `0` it will be ignored. @ozym has told me numerous times and I still don't quite grasp it 🤣

CBurton90:
@ozym can you run your eye over this again when you have a chance? @salichon and I have discussed it and believe it is ready to merge!

salichon (approved):
I m okay 
with it :)

ozym (approved):

PR #1230, opened 2021-12-08
add content to features.csv, and related changes

rumachan:
Add most LRDCP site details to features.csv. Make some small related changes for multi-measure sensors


elidana (commented):
Hi @rumachan 

for what I know about those sensors (which is lacking a bunch of details around installation time, and why some of the changes in location codes you are proposing) this looks very good to me.
I like to see that the `features.csv` is finally populated.
A couple of more general questions
- RU001 does not have any detail on the feature description / property etc. I believe this is intentional, so I'm just curious : why is that? What is that sensor measuring?
- some dates are now switched between sensor+location codes (for example: OT001-02 OT001-03 in the connection files): is that intentional?
- some of the locations codes are removed or number is changed: is that intentional, and this is correcting some wrong info we previously had?

I will spend some more time on it tomorrow, just wanted to ensure that I'm looking at the right (level of) details here

rumachan:
Thanks @elidana 

- RU001 is intentionally blank at this stage. The site has no comms and is only visited twice a year (it's at the Crater Lake). I'll update the file when the guys visit. The contents have no impact at this stage, and I'd rather leave it and get it right the first time.
- The OT001 sensor code and date switch is intentional. It's not ideal I know, but it's reflecting the way Richard coded the logger and the order in which sensors were installed.
- Yes, I had made some mistakes before. The key mistakes I made involved starting a new sensor code when a sensor was changed, which isn't required. I hope I've got rid of all of those now.

elidana:
thanks a lot for the explanation @rumachan !!

PR #1231, opened 2021-12-08
LBZ sensor swap

mnaguit:
- from Guralp CMG-3ESP T3G61 to Trillium Horizon TH120-1 820
- at LC 10 for HH(ENZ)


CBurton90 (approved):

salichon (approved):
:+1 yup

PR #1233, opened 2021-12-12
WKZ sensor swap

mnaguit:
- from Guralp CMG-3ESP (SN T3G77) to Nanometrics Trillium Horizon TH120-1 (SN 816)
- at LC 10 (HH-ENZ)


salichon (approved):
sweet as looks legit :)

CBurton90 (approved):

PR #1235, opened 2021-12-14
Adding new values for NZC and NZE add NZA later from December Redeploy

megankortink:
Added NZC and NZE. Added end dates where appropriate (network/sites, install/recorders and install/streams).

Only changed network/stations lat/long for NZE because NZC didn't change enough.

ADD NZA data when get voyage logs - don't merge until we get NZA


ozym:
@megankortink  Probably not a big deal, but I'd suggest we used the time windows to keep track of where this station is/was,  As we lose previous information on an update.

```diff
- NZE,TD,Offshore East Cape Kermadec,-36.049,-177.703,0,5779,WGS84,2019-12-19T00:00:00Z,9999-01-01T00:00:00Z
+ NZE,TD,Offshore East Cape Kermadec,-36.049,-177.697,0,5779,WGS84,2019-12-19T00:00:00Z,9999-01-01T00:00:00Z
```

Not 100% the tests have been updated to handle this, but if not they should be.

ozym:
Second thoughts, this may not be worth it (adds more complexity) and given the "real" locations are tracked in the sites.csv file, and stations in this instance is more related to a pin-on-a-map then ignore me.

megankortink:
Yeah that's what I was wondering in the network/stations as we don't have '40'/'41' I thought it would get confusing if I was changing the end time and then having another entry. I suppose this usually wouldn't be a problem because if a sites location was changing usually the name would as well

megankortink:
@salichon all three of the DART changes are now in this PR

FYI @ozym

salichon (changes requested):
Hello @megankortink 
as for the seismic stations, though to a slightly larger extend 
_My view is not to change the station position  once it has been settled for the first time._

Unless we have reasons.
The changes are tracked along with the sites changes.

To me a "station" is a general concept of  geolocalizing observation devices, recorders and transmission  about right. 
The location accuracy on the measuring point is provided using the instrument site location 

As opposed to seismic stations it ll be hard for us to change the names of the DART systems if it is (re)located away  from an intitial drop.  Sites will handle this  to some extent.

(NB After if a system is  ever dragged  more than `xx` km from its original place we may have to handle this in combo with the "ocean techs" guys  :)  at best way before the deployment -  this could be a good "data" question to one of the DART  technical meeting  )

salichon:
for the DART i d suggest a warning distance of say  10 km change before we start to  question the tide constituents and all  that sort of things in the stations  and delta. Although... even in the deep ocean tides might have some rather different values 10 kn appart (but i m not tsunamo-oceano-tido-logist honestly ;) )

megankortink:
@salichon So are you asking that we don't change the network/stations.csv? I think that will get confusing though ... I think network/stations should have the CURRENT position and then network/sites shows the record of the past locations by using the location codes `40`/`41` and the start and end dates. I'm also keen to merge this and then we can have a further discussion about this in a DART technical meeting in the new year

salichon:
I m indeed not in favour of changing a station position to be also consistent with delta in general :)

megankortink:
@salichon so just to get my head around this, we aren't changing network/stations.csv and we just keep it with the first position?

network/sites.csv
```diff
@@ -1476,12 +1476,15 @@ NTVZ,31,-39.0984817,175.6759733,1260,,WGS84,Unknown,2016-01-14T02:10:00Z,9999-01
...
 NWEZ,10,-39.27488,173.86743,230,,NZGD2000,Topographic Map,1996-04-03T17:09:00Z,2012-10-16T00:00:00Z
 NWFS,20,-41.589065731,175.233958687,14,,WGS84,Internal GPS Clock,2006-05-05T02:01:00Z,9999-01-01T00:00:00Z
-NZA,40,-42.3707,176.9109,0,2690,WGS84,Unknown,2019-12-22T00:00:00Z,9999-01-01T00:00:00Z
+NZA,40,-42.3707,176.9109,0,2690,WGS84,Unknown,2019-12-22T00:00:00Z,2021-12-17T23:59:59Z
+NZA,41,-42.3717,176.9093,0,2635,WGS84,Unknown,2021-12-18T00:00:00Z,9999-01-01T00:00:00Z
 NZAS,20,-46.586807,168.378357,13,,WGS84,Internal GPS Clock,2008-10-08T22:00:00Z,2015-06-30T23:59:59Z
 NZB,40,-40.6003,179.0996,0,3184,WGS84,Unknown,2020-09-18T00:00:00Z,9999-01-01T00:00:00Z
-NZC,40,-38.2001,-179.7978,0,3545,WGS84,Unknown,2019-12-13T00:00:00Z,9999-01-01T00:00:00Z
+NZC,40,-38.2001,-179.7978,0,3545,WGS84,Unknown,2019-12-13T00:00:00Z,2021-12-10T23:59:59Z
+NZC,41,-38.2004,-179.7978,0,3545,WGS84,Unknown,2021-12-11T00:00:00Z,9999-01-01T00:00:00Z
```
salichon, line 1484: In `sites.csv`  NZC 41  lat should be `-38.204` 
(picky me:))

salichon (changes requested):

network/sites.csv
megankortink: Hey @salichon pretty sure in voyage logs it is `38 12.023 S` (degrees minutes) so then that is `-38.20038` decimal degrees, so rounded is `-38.2004`, how did you get `-38.204`?

megankortink:
@salichon I just noticed some weird things in the OLD network/stations.csv values 

So for NZA it had `-42.371, 176.914, 0, 2690`
For NZE `-36.049, -177.703`

But in network/sites.csv we have for location code '40' the previous and also first location:

`NZA,40,-42.3707,176.9109,0,2690,`
`NZE,40,-36.0493,-177.708,`

So for network/stations.csv would you like me to use the previous values which could be wrong or match them to the network/sites.csv values which would be:
NZA=`-42.371, 176.911, 0, 2690`
NZE=`-36.049, -177.708`

salichon:
@megankortink 
Thanks ! 
hello yeap i reckon the early DARt information was coming from mmultiple channels and let  ground for not totaly consistent info :)

Looking back at NZA N1 datat  i think that the depth is more `2638` than the one ionfromed at `2690`

megankortink:
@salichon so assuming we are keeping network/stations at the original values, I can leave the detpth for NZA at `2690`, but shall I change the network/sites value from `2635` to `2638`? That's quite a minor change...

network/sites.csv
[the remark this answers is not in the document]
megankortink: @salichon so NZE is currently `5779`, do you mean `5747`? Happy to change :)

salichon: All good i miiplace  the  decimal separator ....

megankortink: No worries! Just wanted to check :)

network/sites.csv
```diff
@@ -1476,12 +1476,15 @@ NTVZ,31,-39.0984817,175.6759733,1260,,WGS84,Unknown,2016-01-14T02:10:00Z,9999-01
...
 NZD,40,-36.0998,178.6037,0,2445,WGS84,Unknown,2021-07-23T00:00:00Z,9999-01-01T00:00:00Z
-NZE,40,-36.0493,-177.708,0,5779,WGS84,Unknown,2019-12-19T00:00:00Z,9999-01-01T00:00:00Z
+NZE,40,-36.0493,-177.708,0,5779,WGS84,Unknown,2019-12-19T00:00:00Z,2021-12-12T23:59:59Z
+NZE,41,-36.0499,-177.697,0,5779,WGS84,Unknown,2021-12-13T00:00:00Z,9999-01-01T00:00:00Z
```
salichon, line 1487: NZE 41 seems to be shallower to 5777 meters from the dat inspection

network/sites.csv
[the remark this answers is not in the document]
megankortink: Will change depth of NZE to `5777`

network/sites.csv
megankortink: Will change depth of NZE to `5777`

salichon (approved):

PR #1238, opened 2021-12-16
TON?  Tongariro  portable/temp  seismic stations family  are closed since the 2010's

salichon:
Closing  the TON?  stations - not used and confusing


rumachan (approved):
I can't vouch for the closure dates of these stations, but I know they are closed. I assume @salichon has checked the archived data to find the closure dates/

mnaguit (approved):
- Closing dates for TON1 to TON8 corresponds to closing of stream, connection, sensor and site. Hence all good.
- For TON9, just a side note that unlike other TON* stations, this one does not have any other metadata aside from the items in `stations.csv` .

PR #1239, opened 2021-12-16
 BYTT  is closed since late 2000's

salichon:
BYTT is temporary and has partial metadata
station is closed  until further interest


mnaguit (commented):
Closing date is aligned with sensor and site end date.
Shall we also update the closing date for the `install/connections.csv`?

salichon:
@mnaguit yeap indeed this is looking like a mistake :)  (even though we don t  have  datalogger with it:)))

mnaguit (approved):
Thanks @salichon 
All good! 👍

PR #1241, opened 2021-12-16
ARCxx temp SM Default closure date

salichon:
Strong motion temporary
Partial metadata
Still opened


mnaguit (approved):
Similar to #10058 SR** stations, I'll approve given the closing date corresponds to end-of-data from these temporary stations.

PR #1242, opened 2021-12-16
IFxx  strong motion temp stations closed since late 2000's

salichon:
Close the IFxx temporary SM stations


mnaguit (approved):
Closing date matches the closing of connections, sensors and sites. 
Thanks @salichon

PR #1243, opened 2021-12-17
MSxx Temp station - Default closing date - metadata incomplete

salichon:
MS seismic experiment (most likely)
Default closing date


mnaguit (approved):
Ok with suggested closing date, as there is other metadata available to align it with.
Thanks @salichon

PR #1244, opened 2021-12-19
update geodetic contact to comply with IGS latest recommendations

elidana:
Hi, 

IGS has recommended to avoid personal contact in IGS stations sitelogs.

The IGS public sitelogs of our stations has now been updated on the Site Log Manager.

This pr is to reflect a similar change on sitelogs of all our GNSS continuous sites. 

I have tested locally the change and it works fine. This pr is also updating the repository to https://data.geonet.org.nz


megankortink (approved):
LGTM. Quite like the use of reception for the contact phone number, makes it easy.

PR #1248, opened 2021-12-23
SS0xx strong motion stations are closed - metadata incomplete

salichon:
stations only no other information 
Name not standard


mnaguit (approved):
Approving, noting that there is no other metadata available for these temporary stations aside from the ones in `stations.csv`

PR #1250, opened 2021-12-23
AP2W and APSZ Stations only metadata and closed

salichon:


mnaguit (approved):
Approving, noting that `stations.csv` is the only file providing metadata for AP2W and APSZ

PR #1251, opened 2021-12-23
CCH1 CCH2 stations XX are closed - default date

salichon:
CCH? strong motion serie entirely closed


mnaguit (approved):
No other metadata for CCH2 and CCH2 other than `stations.csv`

PR #1254, opened 2022-01-11
add dart tidal constituents

ozym:
A straight copy, other than conversion from meters to centimeters.


megankortink (dismissed):
LGTM

salichon:
Thanks @ozym 
as a reminder these are the NIWA work output  right? 
and we re not assuming large dart site  distance difference in the near future.

salichon (approved):
Add a really brief look at niwa email content for one BPR - i had 
looks legit :) 

Edit ... ** besides the `autogo-test` to be remade i suppose :)

ozym:
@salichon @megankortink needed to add the sites to the gauges file (to pass the tests). But this also adds some details about the latitude that was used in the processing so it's good (just it doesn't quite fit with the LINZ/tide gauge bias).

salichon:
Thanks @ozym 
The `Analysis Longitude`  is the same for the "BPR" in gauge.csv  i guess it s a default value latitude being the important one.
Are we required to fill up with the DART CREX tag in complement  ?

ozym:
@salichon yes the value is what was used and is a default I assume.  We don't need to document the CREX as this is done outside of our systems and is done as part of a bulletin rather than individually as per the tide gauges.

salichon (approved):

PR #1257, opened 2022-01-12
JCZ sensor swap

mnaguit:
- from Guralp T3G78 to Trillium Horizon 819
- for LC10 with HH/LH (ENZ) streams


CBurton90 (approved):

salichon (approved):
sweet as right!

PR #1258, opened 2022-01-12
update dart response for downloaded data

ozym:
I have updated the DART response, to be:

- only for 15s
- uses a gain based on conversion from counts as downloaded to meters.
- uses Int32 encoding (the maximum difference is too great for Steim2, > 30 bits).


megankortink:
@ozym why are we removing the '-900' lines from the `install/streams.csv`?

ozym:
@megankortink this represents the 15 minute sampling which isn't recovered from the downloaded data, so there will not be any 15m data available (via this mechanism) as otherwise the response information would be contradictory.

megankortink (approved):
As far as I can tell this LGTM!

salichon (commented):
@ozym it looks good 

i m wondering why the gain change ?

ozym:
The gain is related to the new data just pulled off (it's in PSI) whereas the old gain was related to the RAW messages which had already been converted to height (so change was to get the height units right).  As we are not loading RAW data into the archive as per miniseed this isn't needed, but we will be adding the "uploaded" data.

PR #1263, opened 2022-01-18
VUWB metadata update

mnaguit:
- transition from axial (12Z) to non-axial (XYZ) components


salichon:
@mnaguit Hello ! 
as confirmed druing our delta catch up 

I think you d have to keep the "axial" component whether it is XY or 12 
Reckoning also that the XYZ is not accounted for totally into Delta/response creation 
afaik ..
cheers

mnaguit:
Ah, I thought from delta discussion, we can update the metadata and leave the data as is? And the arrays are are now mostly streaming XYZ (or ENZ)?

mnaguit:
> @mnaguit Hello ! as confirmed druing our delta catch up
> 
> I think you d have to keep the "axial" component whether it is XY or 12 Reckoning also that the XYZ is not accounted for totally into Delta/response creation afaik .. cheers

Got it @salichon . Thanks.
I got confused with XYZ being non-axial, hence the change in this PR from axial (12Z) to non-axial (XYZ).
Will close this PR now and delete the branch. =)

PR #1264, opened 2022-01-24
Add code.csv

ricardojohansson:
Each individual LRDCP site runs a unique version of code that is tailor made to suit the requirements of the site. As the sites evolve (eg when the number of sensors is increased/decreased, a different type of sensor is introduced to the site or the timing of an instruction is modified,) we need to have a current & historical record of what version of code is/was running at a site at any particular time.

I have drafted 'code.csv' based on the records I have for the volcano LRDCP sites. 

@elidana @rumachan please check you are happy with the format etc - (and I'm guessing this is the best place for it? )

cc @jacksonshanks


ricardojohansson:
@rumachan with this csv in place we should also change the cr6programs folder (https://github.com/GeoNet/data-logger/tree/main/cr6programs/complete_logger_programs) so that there is a single folder for each site that contains ALL code versions for the site (past, current & future). Previously we have attempted to keep current versions in the top layer and use sub folders for 'previously deployed' and 'to be deployed' versions - this will be unnecessary with the csv in use.

ricardojohansson:
@elidana did you still want to have a chat about this?

elidana:
@ricardojohansson yes, I'll set a time for the 3 of us

ozym:
This needs to have more of an explanation as to what it is doing, plus the background checking software (i.e. code) needs to be developed before merging.

The use of a free-form field is also strongly discouraged (we've been burnt with this before).

ozym:
First question, does this impact the data recorded?

ozym:
Buy, change the data recorded, I mean. Given we have other information about what sensor was installed with what datalogger was used where, does this table add anything extra from a data point of view?

elidana:
@ozym , for the third question, there is also a need to track the firmware of each datalogger, this is now done in https://github.com/GeoNet/delta/pull/1195

Very good questions, and I agree with your points. Have now created an internal issue to discuss this topic

ricardojohansson:
Closing this PR following a discussion Steve & I just had - we will now put the code.csv in the data-logger repo but can reassess in the future if a need for any of the information captured in code.csv would be more appropriately stored elsewhere.

PR #1266, opened 2022-02-03
RLNS recorder swap

mnaguit:
- from TitanSMA 001257 to TitanSMA 001706
- for LC 20 (HN-ENZ)
- firmware version 4.5.22


salichon (approved):
if  we get some data back we will amend.
@CBurton90 for that TitanSMA: 
 Does Offline meant not transmitting data but still recording good data or "Kaputt" datalogger? 
 (if good data could be used to train to modern offline backfilling practice ... if ever ... )

CBurton90:
@salichon good question, I can check when I get it back! Yes there should be backfilling progress for pulled field equipment that may have "good data" on it. It's on my to do list with Mark

CBurton90 (approved):

PR #1268, opened 2022-02-10
fix equipment sort order for capitals

ozym:
Fixing the sorting of equipment by using lowercase compare.


elidana (approved):
thanks a lot @ozym , that's awesome!! (and thanks for fixing all the metadata bits around the various files as well!)

PR #1271, opened 2022-02-13
add expr package

ozym:
This is a helper package that makes adding numerical configuration where it makes more sense to use an equation rather than the final answer.


internal/expr/expr.go
```diff
@@ -0,0 +1,106 @@
...
+	return res, nil
+}
+
+// ToFloat64 evaluates the given string as a float64, or it returns an error if the result is invalid or the expression is Unknown.
+func ToFloat64(s string) (float64, error) {
+	val, err := eval(s)
+	if err != nil {
+		return 0, err
+	}
+	res, _ := constant.Float64Val(val)
```
bpeng, line 42: Does it make sense to validate the type before getting the value, to avoid possible panic?

bpeng (commented):
Looks good, little suggestion which may not be necessary

internal/expr/expr.go
ozym: I think, to some extent this is already handled by the underlying code.

e.g.
https://cs.opensource.google/go/go/+/refs/tags/go1.17.7:src/go/constant/value.go;l=548

ozym: but yes, there's a panic in the.

ozym:
@bpeng I have added some guardrails around the values after evaluating, this is a good enhancement as I suspect this may have caused issues when dealing with floating point values that are given as integers.

I've also added a complex number value, as this is something that crops up a lot in response configuration.

bpeng (approved):
OK

PR #1272, opened 2022-02-14
VRZ sensor swap

mnaguit:
- from Guralp CMG-3ESP (SN T3B65) to Trillium Horizon TH120-1 (SN 805)
- for LC 10 HH(ENZ)


salichon (approved):
sweeet

CBurton90 (approved):

PR #1179, opened 2021-09-20
WVZ strong site code change legacy update + Delta correction on relevant start

salichon:
- sire code changes legacy 
- 2011 starting of data  wrt to the metadata is more relevant now.

PR #1274, opened 2022-02-14
Fix WVZ strong motion datalogger stream 20 End (2011)

salichon:
Correct metadata wrt to past correction https://github.com/GeoNet/delta/pull/1179

WVZ strong motion datalogger end


mnaguit (approved):
Looks good, as per comment here https://github.com/GeoNet/tickets/issues/9485#issuecomment-948173964 & delta PR issued previously

PR #1275, opened 2022-02-14
BFZ Metadata correction to match orphan data 2015 and 2021

salichon:
IRIS orphan data check
BFZ 2015 and 2021 metadata adjustment


mnaguit (approved):
Looks good, as per the solution proposed in https://github.com/GeoNet/tickets/issues/9800#issuecomment-1035839695

PR #1278, opened 2022-02-22
SYZ sensor change

mnaguit:
- for LC 10 with HH-LH(ENZ) streams
- from Guralp CMG-3ESP to Trillium Horizon TH120-1


salichon (approved):
yeap

CBurton90 (approved):

PR #1283, opened 2022-02-23
WHVZ datalogger swap

mnaguit:
- from Q330S/6 SN6053 to Centaur SN8743
- from LC 10 (HH/LH-ENZ streams)
- Centaur firmware: 4.5.20


salichon (approved):
goo
d as

CBurton90 (approved):
Just fyi there is also an acoustic stream on LC30 but this looks fine!

mnaguit:
> Just fyi there is also an acoustic stream on LC30 but this looks fine!

I'll check the acoustic stream at LC 30 at data-metadata verification. Thanks @CBurton90

PR #1291, opened 2022-03-07
add start and end times to the gauge csv file to handle multiple dart deploys

ozym:
This PR is aimed at getting the gauges file in step with the constituents file.

I'm expecting an entry into the gauges file like:

 ```
NZA,TD,,0,-42.3707,176.911,,2019-12-21T17:00:00Z,2021-12-17T01:00:00Z
NZA,TD,,0,-42.3717,176.9093,,2021-12-18T00:00:00Z,9999-01-01T00:00:00Z
```


meganmadley (approved):
Shall we change the actual start and end times in another PR? Or will this find them from somewhere else?

ozym:
Yes, that would be a good plan, this PR is aimed at getting the mechanism in there with fairly wide time windows. There is a test about overlaps that I haven't done (only starting times) which I will need to look at.

salichon (approved):
yup

PR #1292, opened 2022-03-08
Adding start and end dates for DART A, C, and E

meganmadley:
Will need to update other buoys, will do that when also updating constituents file, will also check if the coordinates changed


environment/gauges.csv
```diff
@@ -11,11 +11,14 @@ MNKT,TG,90,180,37.03,174.31,-3704657 17451175 MNKT,2000-01-01T00:00:00Z,9999-01-
 NAPT,TG,97,180,39.29,176.55,-3947566 17692007 NAPT,2000-01-01T00:00:00Z,9999-01-01T00:00:00Z
 NBRT,TG,370x,180,43.34,172.46,-4357010 17277383 NBRT,2000-01-01T00:00:00Z,9999-01-01T00:00:00Z
 NCPT,TG,162,180,34.25,173.02,-3441483 17304870 NCPT,2000-01-01T00:00:00Z,9999-01-01T00:00:00Z
-NZA,TD,,0,-42.3707,176.911,,2000-01-01T00:00:00Z,9999-01-01T00:00:00Z
+NZA,TD,,0,-42.3707,176.911,,2000-01-01T00:00:00Z,2021-12-17T23:59:59Z
+NZA,TD,,0,-42.3707,176.911,,2012-12-18T00:00:00Z,9999-01-01T00:00:00Z
```
ozym, line 15: should that be 2021

ozym (changes requested):

environment/gauges.csv
meganmadley: Haha yip

ozym (approved):

PR #1294, opened 2022-03-10
Adding tsunami gauges and DART start and end times

meganmadley:
Used install/connections to get tsunami gauges values into gauges.csv
Used network/stations to get DART values into gauges.csv
For constituents.csv, used the values in gauges


salichon (approved):
looks good.
i  don t recall any Tide gauge that is closed.
thanks!

ozym (approved):

PR #1295, opened 2022-03-16
update network code of CBLB 

elidana:
CBLB is part of a building array, so this pr is fixing the network code to reflect that.

@ozym , is there any downstream effect we shall consider before merging it?

Thanks!


mnaguit (approved):
Thanks Eli

PR #1296, opened 2022-03-17
GNSS LEVN Fixing constellation start and end dates

meganmadley:
Adding the start and end dates to a 48 day gap where we only had GPS+GLO instead of the full constellation for LEVN. Used the rinex files to find the start of the GPS+GLO period, @aleks-spes reconfigured the receiver right on the change of day, so went for that for the starting time of full constellations again.


elidana (commented):
nice! Thanks!
approved in principle, though I did not check against rinex data

aleks-spes (approved):
Looks good!

PR #1302, opened 2022-03-29
KHZ/SCTS Titans got their serial nr right - Good Titans

salichon:
making the titan nr right from equip tickets wrong (updated now)


salichon:
FYI @mnaguit  i amended the older  equip tickets with that

CBurton90 (approved):

mnaguit:
I'm late for the party :smiley:

PR #1311, opened 2022-04-06
MTSX fix metadata

elidana:


salichon:
Connections and streams updated  to match  MTSX information and creation of the _full_ metadata volume strong  BN and  weak EH 
All sweet  as

salichon (approved):

PR #1313, opened 2022-04-14
Update sessions.csv

meganmadley:
Fixing an error we spotted with WAKA. Noticed it was missing a session entry for the ALLOY that was installed there at the end of March


aleks-spes (approved):
Looks good!

PR #1316, opened 2022-04-20
Guralp Fortis response integration

CBurton90:
A few things:

- This is a first try
- I am a little unsure on sensitivity/gain value but for a 40Vpp sensor (+/-20V) these make sense to me
- We don't have to include 2g. Our standard is now 4g and all deployments should have this clip level but I am trying to foresee circumstances where decisions change or a mistake in the deployment is made and a 2g sensor is deployed (great, we already have the response info in that instance)
- I will run go generate to produce the auto.go file once we have finalized this initial PR


ozym:
Problem is there will not be two Fortis models. We will need to work this through via perhaps a gains file mechanism or similar.

CBurton90:
@ozym all 2g response has now been removed

salichon:
Hello @CBurton90 
How are we about the Fortis response.
is it wiorth for me to get inot the details - or this PR has some dependencies ? thanks
cheers
j

salichon:
Kia ora @CBurton90 
I would like to have the confirmation that the FORTIS are 100Hz strong motion as opposed to to the global 200Hz network of strong motion.  
Are we introducing differentiated intrumentation for strong motion now ?  
What would be their use ?
or i did not understand the specs  too ha  ... :)

ozym:
@salichon the name of the response is going towards the NRL nomenclature: https://service.iris.edu/irisws/nrl/docs/1/help/#instrumentconfigs

  Fortis-DC-100Hz-4g

will end up as:

sensor_Guralp_Fortis_HF100_SG0.51_STgroundAccel

HF - High-Frequency_Corner (100 Hz)
SG - Sensitivity
ST - Sensor_Type

CBurton90:
@ozym @salichon @mnaguit Hi all, could I please get a thumbs up/thumbs down for this PR. Testing seems to be wrapping up for the Fortis (I am happy) so would be looking to deploy to a canary site soon.

salichon:
Hello! 
@CBurton90 
My thumb is blocked in the middle :) 


Would you check the Code labels ?
I sense that _Code A_ is    _Transfer function type:  A [Laplace Transform (Rad/sec)]_ 
and you re suing a transfert function that is in Hertz.
https://github.com/GeoNet/delta/blob/8dca4f90a2846490c2a21873a1ad577eff0ee1cd/resp/responses/sensors/guralp.yaml#L322-L323

I also  have a wonder aboutt eh Gain at _0.51_ 
woudl you elaborate about ?
i d speculate .._1.940e+9_  >?

CBurton90:
> Hello! @CBurton90 My thumb is blocked in the middle :)
> 
> Would you check the Code labels ? I sense that _Code A_ is _Transfer function type: A [Laplace Transform (Rad/sec)]_ and you re using a transfert function that is in _Hertz_.
> 
> https://github.com/GeoNet/delta/blob/8dca4f90a2846490c2a21873a1ad577eff0ee1cd/resp/responses/sensors/guralp.yaml#L322-L323
> 
> I also have a wonder about the Gain at _0.51_ would you elaborate about ? i d speculate .._1.940e+9_
> 
> See on https://www.guralp.com/apps/paz?x=yes%3Fr%3Ddc100&r=dc100+Acceleration

@salichon I am using the PAZ in regards to hertz (rather than radian/s). Please see:

https://github.com/GeoNet/tickets/issues/9598#issuecomment-939551017

Gain of 0.51 translates to an output of 0.51 V/m/s^2 (this is only the sensor output and has nothing to do with the digitiser). For a +/-20V sensor output (40Vpp) and matching digitiser scale (40Vpp) this will produce a clip level of 4g.

salichon:
Ola @CBurton90 
so i have still issues  ;)  haha.

from https://github.com/GeoNet/delta/blob/cd733cc324d1d3faa82e0c7a347d32b7b85ce40f/resp/generate/paz.go#L15 

As per delta  " secrets "
_func (p PAZ) PzTransferFunction() string {
	switch p.Code {
	case "A":
		return "PZFunctionLaplaceRadiansPerSecond"
	case "B":
		return "PZFunctionLaplaceHertz"
	case "D":
		return "PZFunctionLaplaceZTransform"
	default:
		return "PZFunctionUnknown"
	}_
	
	
I reckon then  that this shoudl be amended
 by  `code: B`
 
 @ozym  What would be the culprits when "Paz Code" is not the right ones ? (if it is) . It is just  a "Label"
Woudl we need to check this paz parameter?

CBurton90:
@salichon good spot (this is why we have multiple eyes on things), updated!

salichon (approved):
this is looking good

CBurton90:
@ozym @salichon We are planning to install the Fortis on Weds/Thurs this week. It looks like this PR is mostly finalised so please take another look over it and review when you are ready! Cheers.

ozym (commented):
we'll see how it goes.

CBurton90:
@ozym Just been updated again and it seems the weather ain't looking too flash so its likely the deployments may be postponed to the following week (22nd-26th). At least it gives me an extra bit of time to test the newly made cables...

cc @salichon

salichon:
a fix there:  https://github.com/GeoNet/delta/pull/1421

PR #1324, opened 2022-05-01
update connections to allow for a datalogger pin offset

ozym:
One of the trickier parts with building equipment response meta data is handling the exceptions, in this example the Quanterra Q330HR series has a different channel response depending on whether the first bank (0-2) or second bank (3-5) channels are used. Normally the broadband (high resolution) sensor will be plugged into the first bank, and the strong motion will be the second.

This PR adds the ability to indicate which bank (or channel) the sensor has been plugged into on the datalogger. In most instances this will be zero as the number of datalogger channels matches the number of sensor components, or at least they both start at zero.

Where there are two sensors attached to a single datalogger, the offset will not be needed unless the datalogger has a different response for any channel (as per the Q330HR) or if the full calibration details of the datalogger are to be included. The last part will be required to match the sensor with the datalogger channel.

The follow on from this PR will be:

- to update any downstream applications that may use a `connections.csv` file (mostly for testing)
- look at adding a Channel (or Pin) column in the `calibrations.csv` file to allow for datalogger calibration (if needed).


ozym:
This is likely to cause a merge conflict with any outstanding PRs that adjust the `connections.csv` file.

salichon:
thanks !
work in progress i suppose. 

So  Offset in Delta is the start of the  first bank (0-2) or second bank (3-5) channels (either 0 and 3 (ideally) 
(as far as i understand then once the Offset is given. this will point to the build of the datalogger yaml
for which a response by offset will be  created)

ozym:
Yes, @salichon this is the first step in a chain of updates needed to make the response generation a lot simpler.

I wasn't sure whether to use the work "Pin" as historically a datalogger had its channels defined as "pins" (like in a cable plug) this is to distinguish between a "Channel" code as used in say the FDSN "channel" naming convention.

We can change this later as it's only used in the header and code, not the data.

It also means we can add different sensors to different parts of the dataloggers, if we are worried about datalogger calibrated values rather than the nominal sensitivities.

ozym:
I've switched to using "Pin", this is to avoid using Offset which is used elsewhere related to sensor offsets from a position.

The "Pin" is the lowest channel number which matches the lowest sensor component number (i.e. a 6 channel Q330 has pins 0 to 5, and a broadband sensor as 3 components, 0 to 2.  So the Pin will be 0 if plugged into bank A (0-2), or 3 if in bank B (3-5).

salichon:
cool  
Some of the `connections.csv` in the PR  are still using "offset" label in place of "pin" is it correct  for these case (tools or tests) ?

ozym:
@salichon good spotting, these are just for testing other things so don't flag as a problem.  I have updated them.

CBurton90:
@ozym I understand this links sensors to digitiser bank (0 or 3) but for those circumstances where the bank is set with a different number of bits (e.g., 24bit vs 26bit Q330HR) where is the actual bit weight value (counts > V) stored for the response creation?

ozym:
@CBurton90 that will be left to the response label as found in  #1325

CBurton90 (dismissed):

install/connections.csv
```diff
@@ -1,996 +1,996 @@
...
-WVRV,10,Whirinaki West,Short Period,1995-01-13T16:07:00Z,1995-05-14T15:44:00Z
-WVZ,10,Waitaha Valley,,2003-09-06T00:00:00Z,9999-01-01T00:00:00Z
-WVZ,10,Waitaha Valley,Short Period,1990-05-16T14:12:00Z,2003-09-05T23:51:00Z
-WVZ,11,Waitaha Valley,Short Period,2003-09-06T10:56:00Z,2003-12-11T16:33:00Z
-WVZ,20,Waitaha Valley,,2003-09-06T00:00:00Z,2011-10-12T23:59:59Z
-WWRF,10,White Water River,Short Period,1993-03-20T06:03:00Z,1993-06-25T07:15:00Z
-YARM,10,Yarra River,Short Period,1993-11-23T05:18:00Z,1994-03-11T18:48:00Z
-YUPC,10,Yup Farm,,2001-01-15T11:33:00Z,2001-06-27T20:55:00Z
-ZIHA,10,Zillian Hill,Short Period,1999-08-01T08:42:00Z,2005-05-20T12:47:00Z
+Station,Location,Place,Role,,Start Date,End Date
```
elidana, line 1: change to 
`Station,Location,Place,Role,Pin,Start Date,End Date`

elidana (changes requested):
Hi @ozym ,

sorry for being slow here. At a first pass looks good to me. With a small change in the connections.csv header.

A couple of other comments that can be addressed at a later stage:
1. I would like to find a different name than pin, but agree that channel is overloaded and not ideal. This is not crucial now, so happy to merge this and having some name party later on (but soon enough not to cause too much disruption).

2. is it worth adding a test that checks that "pin" is only used when necessary? not sure how this can be done in a programmatic way. I'm thinking along the lines of ensuring that:
- only inital pin in a bank is inserted
- only pin with a "meaning" are insterted
- format is what the code that create the response file would expect (integer value, 0 or 3)

3. we need to add some documentation around that, but I think this can come to the end, once we have captured all use cases and scenarios (geomag, quanterra dataloggers, possibly envirosensors, etc).

salichon:
_- FAIL: TestConsistency/check_file_consistency:_connections (0.00s)
        consistency_test.go:39: unexpected connections.csv content -got/+exp_
        
Can this be the header name ?

ozym:
@elidana the pin is only added when it's needed.  If no pin is given then it defaults to 0

ozym:
Happy to use another word that means the wire coming out of the sensor cable plug is attached to the wire in the datalogger cable plug.  (where there's not much standard in how they are numbered or sometimes letters are used).

elidana:
I heard @CBurton90 using "bank" terminology to describe Centaur dataloggers and cabling.

have tried to have a look at the IRIS response library to see if they use a specific terminology but couldn't find anything obvious.

In the case of a fluxgate (and possibly LRDCP?) we cannot use too many standards probably, considering that a "pin" can directly relate to the measure that is done using a given sensor. In our first attempt for EYRW, we were using pin 0-2 for the XYZ component, and pin 3 for the temperature sensor. 

Let's try pin for now. I guess we will need to iterate a few times

elidana (approved):

ozym:
To try and move this along a bit, I have replaced the word "Pin" with the word "Number".

elidana:
Hi @ozym ,
happy for this to be merged.

Perhaps when we will continue this body of work, is it worth looking at nomenclature used within StationXML and (in particular) the concept of source and subsource as in http://docs.fdsn.org/projects/source-identifiers/en/v1.0/channel-codes.html? 

But please merge away (from the little of my knowledge here)

elidana (approved):

ozym:
> Hi @ozym , happy for this to be merged.
> 
> Perhaps when we will continue this body of work, is it worth looking at nomenclature used within StationXML and (in particular) the concept of source and subsource as in http://docs.fdsn.org/projects/source-identifiers/en/v1.0/channel-codes.html?
> 
> But please merge away (from the little of my knowledge here)

Yes, we need to think about that, but it's still more aimed at the miniseed side of things (as recorded) which is still a work in progress. The stationxml part: http://docs.fdsn.org/projects/stationxml/en/latest/reference.html#channel  will still have a "Code" which may end up the same, but will probably not use those names just yet.  But this is the direction things are going, see the "sourceID" at the bottom as an example of that.

PR #1325, opened 2022-05-03
add channel and component

ozym:
This attempts to add a reference mechanism to link simple responses (given by a label) to a sensor component (and also include base characteristics like dip and azimuth), or to a datalogger channel using a pin offset and a sampling rate. The edge case with multiple streams with the same sampling rates but different responses has been skipped for now.

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -20,6 +20,8 @@ Meta information for the GeoNet equipment network.
 * `streams.csv` - Datalogger and recorder sampling configurations
 * `gains.csv` - site specific settings applied to individual datalogger and sensor that may impact overall sensitivities
 * `calibrations.csv` - Individual sensor sensitivity values that can be used rather than default values.
+* `components.csv` - Individual sensor elements including measurement position and responses.
+* `channels.csv` - Individual datalogger recording elements including digitiser position, sampling rate, and responses.
 
 * `cameras.csv` - Installed field cameras.
 * `doases.csv` - Installed field DOAS (Differential Optical Absorption Spectrometer) equipment.
@@ -262,6 +264,44 @@ For the component, sensitivity, and frequency either a value can be given direct
 | _Stop_ | Calibration stop time|
 
 For a second order polynomial response, the output is expected to be `Y = a * X + b` where `X` is normally the input voltage, and Y the corrected signal. The terms `a` and `b` are the factor and bias respectively. The gain adjustments (`a'`, `b'`) update this via `Y =  a' * X + b'`
+
+#### _COMPONENTS_ ####
+
+Sensor model component descriptions. The type is generally of the form "Accelerometer, Short Period Seismometer" etc.
+The number represents the order of sensor components, this generally maps to the sensor cable and how it is connected
+into the datalogger.
+Subsource is the general term used for labelling the sensor component and is usually the last character in the SEED channel convention.
+Dip and Azimuth are used to indicate the relative position of the sensor component within the sensor package and will be used with the
+overall sensor installation values to provide component dips and azimuths.
+
+| Field       | Description | 
+| ----------- | ----------- |
+| _Make_      | Sensor make
+| _Model_     | Sensor model name
+| _Type_      | Sensor type
+| _Number_    | Sensor component offset
+| _Subsource_ | Sensor component label
+| _Dip_       | Internal dip of the compnent relative to whole sensor
+| _Azimuth_   | Internal azimuth of the compnent relative to whole sensor
+| _Types_     | A shorthand reference to the SEED type labels
+| _Response_  | A reference to the nominal StationXML response 
+
+
+#### _CHANNELS_ ####
+
+The individual channels configured for a given datalogger model, these include the channel numbers and sampling rates.
+The channel number is an offset into the digitiser or digitisers and are used to match the connected sensor component
+and the expected response. Some digitisers have different nominal responses for different groups of digitiser channels.
+
+| Field           | Description | 
+| --------------- | ----------- |
+| _Make_          | Datalogger make
+| _Model_         | Datalogger model name
+| _Type_          | Datalogger type
+| _Number_        | Datalogger channel offset, an empty value will map to zero
+| _Sampling Rate_ | Configured Channel sampling rate
+| _Response_      | A reference to the nominal StationXML response 
+
 
 ### CAMERA ###
 
--- a/meta/channel.go
+++ b/meta/channel.go
@@ -1,0 +1,145 @@
+package meta
+
+import (
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+const (
+	channelMake = iota
+	channelModel
+	channelType
+	channelNumber
+	channelSamplingRate
+	channelResponse
+	channelLast
+)
+
+// Channel is used to describe a generic recording from a Datalogger.
+type Channel struct {
+	Make         string
+	Model        string
+	Type         string
+	SamplingRate float64
+	Response     string
+	Number       int
+
+	number       string
+	samplingRate string
+}
+
+// Less compares Channel structs suitable for sorting.
+func (c Channel) Less(comp Channel) bool {
+
+	switch {
+	case strings.ToLower(c.Make) < strings.ToLower(comp.Make):
+		return true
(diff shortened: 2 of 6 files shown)


salichon:
Cool  this  goes with that on right :) https://github.com/GeoNet/delta/pull/1324 .
join the pin  channels and components 

i might need a sketch at some point : ) 
This revisits to  the "connections" side in the delta sense  of the things,

CBurton90 (approved):

CBurton90:
an example on test data with a 26/24 bit digitiser would be useful to see how it works when pin offsets are not blank or zero

ozym:
@CBurton90 I have rebased this, and added a couple of extra test lines for the channels.csv example.

elidana (approved):
this looks good to me, though I will want to review the use of "channel/component/pin" terminology.
can be skipped for now, and we can keep building and review documentation at the end.

salichon (approved):

CBurton90 (dismissed):

ozym:
Further work has led me to think that the channel file should have a "gain" column. This being one if not used, the response for the given gain would be coded as needed (like the other responses). This would likely need an extra column in the installed dataloggers to identify when a non unity gain has been set.  This would leave the "gains.csv" table to reflect the site specific sensor settings rather than dataloggers configuration.

ozym:
I am now thinking that the name "Order" may be better suited than "Pin", it is meaning the same thing but in a less jargony way. The "Order" is the numerical numbering of output signals intended to pair between sensors and dataloggers (with optional offsets for either end).

salichon:
Order can  be or a composite  "PinLabel"
we could figure a Pin mapp table for our instruementaion soemthing thatmay look like the one below ( IF i ubderstand the process ...:)) 
image

salichon (commented):
This is looking good to me ! 
Makes sense 
This might need some testing and practice from my side over practical case. 
i got it for seismic sensor-  and i guess we have to check for the other types,

CBurton90 (dismissed):

ozym:
@elidana any chance of having a look at this one?

elidana:
Hi @ozym ,

@salichon and I had a good look at this, mostly to gain understanding. 
This looks very good from what we can understand, though would be nice to have an idea of implications. 

So far we have one suggestion and two questions.

### 1. suggestion
in `components.csv` and `channels.csv`, the column "Number" is used in two different ways, and is a bit confusing.
in components is the individual "pin number" (just to use some previously used terminology)
in channels.csv is the "pin starting number"

So maybe we can find a different "name" to disambiguate the concept, but don't have a suggestion on what.


### 2. questions

#### question 1 - how this will work for "recorders"?

To test our understanding, we mocked up entries taking as an example the station HIZ.

HIZ has a broadband (described in sensor.csv and datalogger.csv) and an accelerometer (described in recorders.csv).

##### example for a broadband
If we only focus on the broadband, we would add to channels and components the following entries
 ```
[elidana@hutl24201 install]$ more channels.csv 
Make,Model,Type,Number,SamplingRate,Response
Nanometrics,Centaur CTR4-6S,Datalogger,,100,datalogger_nanometrics_centaur_100_response

[elidana@hutl24201 install]$ more components.csv 
Make,Model,Type,Number,Subsource,Dip,Azimuth,Types,Response
Nanometrics,Trillium Horizon TH120-1,Broadband Seismometer,0,Z,-90,0,G,sensor_nanometrics_trillium_response
Nanometrics,Trillium Horizon TH120-1,Broadband Seismometer,1,N,0,0,G,sensor_nanometrics_trillium_response
Nanometrics,Trillium Horizon TH120-1,Broadband Seismometer,2,E,0,90,G,sensor_nanometrics_trillium_response
```

##### example for an accelerometer
if we were also going to describe the accelerometer, were would it go? Would that be an entry in channel like
`Nanometrics,,TitanSMA,Recorder,,200,recorder_nanometrics_titan_200_response` ?

or should it be something like
```
[elidana@hutl24201 install]$ more channels.csv 
Make,Model,Type,Number,SamplingRate,Response
Nanometrics,TitanSMA,,200,datalogger_nanometrics_titan_200_response

[elidana@hutl24201 install]$ more components.csv 
Make,Model,Type,Number,Subsource,Dip,Azimuth,Types,Response
Nanometrics,Titan,Accelerometer,0,Z,-90,0,G,sensor_nanometrics_titan_response
Nanometrics,Titan,Accelerometer,1,N,0,0,G,sensor_nanometrics_titan_response
Nanometrics,Titan,Accelerometer,2,E,0,90,G,sensor_nanometrics_titan_response
```

### question 2: what will happen to entries in existing tables?
this can be discussed "face by face", probably easier
are we going to remove entries from sensors.csv and dataloggers.csv? will they stay the same or will we remove things like "role"? (there might be others)

**Comment by ozym** [comment]
I'll try and answer bit by bit:

For 1, there is no reason not to simply put all the pins in.
e.g.
```
Nanometrics,Centaur CTR4-6S,Datalogger,0,100,datalogger_nanometrics_centaur_100_response
Nanometrics,Centaur CTR4-6S,Datalogger,1,100,datalogger_nanometrics_centaur_100_response
Nanometrics,Centaur CTR4-6S,Datalogger,2,100,datalogger_nanometrics_centaur_100_response
```

As nothing changes except the Pin (Number) in this case it's easier simply to just only add the lowest number needed, and assume any changes will be reflected by adding a fixed Number. For the Centaur it doesn't make much sense as there is no difference to the response on any of the signal channels (pins) -they are all the same.

It makes sense only on a datalogger which has a different response depending on the pin (Number) i.e. a Q330/HR where pins 0,1,2 are higher gain than 3,4,5

I think it's a convention thing really, if you want to be explicit put in all the pin (numbers) and the responses, you will need one for each variation of sample rate (200 Hz, 100 Hz, 1 Hz, etc x number of pins).

Jumping to the last question,

**what will happen to entries in existing tables?**

Nothing should change for any existing table except perhaps the `connections.csv` table.

```
Station,Location,Place,Role,Number,Start Date,End Date
A11,10,Alfredton,Short Period,,1993-02-01T01:23:00Z,1993-02-21T08:48:00Z
```

The Number entry has to point to the first channel (pin) that the sensor is plugged into (i.e. sensor component 0 -> datalogger channel -> pin).   This is related to above, and will be used to select the response from the channels.csv file.

The dataloggers/sensors/recorders etc should all be fine as is.

ozym:
For question 2,

For the recorder, this is modelled as simply a datalogger and a sensor that are joined together in such a way as they don't need a `connections.csv` file entry and will be mapped 1 to 1.  That is, sensor pin number 0 component will be mapped to pin number 0 datalogger channel.

ozym:
For the second part of question 2.

The second accelerometer option will fail as it doesn't have a type, but other than that it looks okay to me.

The "Type" will be used to fill in the datalogger/sensor entries in stationxml which makes them a little easier to find.

http://docs.fdsn.org/projects/stationxml/en/latest/reference.html#datalogger
http://docs.fdsn.org/projects/stationxml/en/latest/reference.html#sensor

e.g.
https://service.geonet.org.nz/fdsnws/station/1/query?network=NZ&level=channel&sta=HIZ&format=text

where the sensor description may take this value.

ozym:
I don't expect these files to get too big, basically one will be related to number of sampling rates, and the other the number of sensor components.  We may need to adjust model names if things get non-standard (which we will probably need to do for geomag fluxgate sensors built for different parts of the world, i.e. the sensor at Apia is different to Scott Base in its base settings).

ozym:
The use of "Number" is the end point of much cycling around between "Sensor Component Number",  "Sensor Pin Number", "Channel Number", "Channel Pin Number" ....  pin is used mainly when dealing with cables which join the two things together (i.e. Pin X will join Sensor Component A with Datalogger Digitiser Channel B).

CBurton90 (approved):

elidana:
hei @ozym ,
thank you SO MUCH for clarifying all the points , it is really helpful.
- happy to not explicitly add each "pin/number" entry when not necessary
- let's leave with "number" for now. @salichon and I had spent a good chunk of time last week to familiarise with the stationxml standard and see if there was some mention of nomenclature there, but of course we couldn't find it 
- thanks for checking the mock up. was mainly to test our understanding

happy for this one to be merged as is and go from there

elidana (approved):

PR #1329, opened 2022-05-09
add ruapehu doases

rumachan:
RUD02 has no specific install time recorded, so used 00:00:00

Re-ordered entries in assets/doases.csv, though I now realise that isn't required.

Think these four file are all that is required.


assets/doases.csv
```diff
@@ -1,3 +1,5 @@
 Make,Model,Serial,Number,Notes
-Avaspec CompactLine,Avaspec-Mini2048CL,2002127M1,,
 Avaspec CompactLine,Avaspec-Mini2048CL,2002125M1,,
+Avaspec CompactLine,Avaspec-Mini2048CL,2002126M1,,
+Avaspec CompactLine,Avaspec-Mini2048CL,2002127M1,,
```
staylorofford, line 4: Added in assets for DOAS installed at TOD01 in 2021

CBurton90 (approved):

PR #1337, opened 2022-05-16
CPIB metadata update

mnaguit:
- 50hz/20hz axial since 2012
- 200hz triggered up to 2019.12.16/ continuous thereafter


salichon (approved):
sweet

PR #1338, opened 2022-05-16
TRAB metadata update

mnaguit:
- all streams for 50/200hz axial (XYZ)
- 200hz triggered ends same time as 50hz continuous ends
-200hz continuous starts thereafter


salichon (approved):
sweet as

PR #1342, opened 2022-05-18
TROB metadata update

mnaguit:
- update to axial (XYZ) components for both 50hz/200hz 
- update triggered end-date continuous start-date for 200hz


salichon (approved):
sweets

PR #1343, opened 2022-05-18
add EYWM fluxgate temperature sensor calibrations

elidana:
Hi, this is adding gains and calibrations for EYWM. 

### Fluxgate calibrations 
Numbers are based on the Danish.Met. results provided in the `/calibration/fluxgates/` folder and added to the `install/calibrations.csv` file as a scaling factor.
Sensor specific scaling factor is obtained as:
`factor = [x,y,z]-coil / resistor`

### Fluxgate gain
Numbers are based on the adjustment done at the time of installation, and added in the `install/gains.csv` as a scale bias.
Site specific bias is obtained considering the sensor installed at the site and obtained as:
`bias =  [x,y,z]-coil * hex-switch-bias / 255`

### components
components and channels are represented as:
- 0 = X = N
- 1 = Y = E
- 2 = Z = vertical

is that correct and compatible with other equipment?

### questions
- we are treating "components" and "channels" as the same thing. Is that correct?
- how should we add temperature sensitivity (for driver and sensor?)
- do we need to add "sensitivity" values for the proton?
- we would like to generate a response file to be tested, how?


ozym:
_- we are treating "components" and "channels" as the same thing. Is that correct?_

The traditional way of wording these (see StationXML) for a channel is what comes out of the datalogger, i.e.
http://docs.fdsn.org/projects/stationxml/en/latest/reference.html#channel  which can mean multiple orientations and sampling rates (e.g. HHZ, HHN, EHZ, LHZ) etc.

But, in equipment brochures the datalogger is often referred to as an X channel recorder, which means the number of inputs it can take. So it sort of sticks with the datalogger.

Component tends to go with a sensor,  a 3 component or a vertical component  (it usually refers to the number of sub-sensors inside the outer sensor housing) and equates pretty much to the orientations where the sub-sensors have been aligned, assuming they are the same type.

_- how should we add temperature sensitivity (for driver and sensor?)_

I expect these will be handled like the other components, but with a temperature specific response (this is part of why all the response changes we've been working on are needed).

_- do we need to add "sensitivity" values for the proton?_

This will also simply be a standard response for this type of sensor (either a gain, or a polynomial).

_- we would like to generate a response file to be tested, how?_

A chicken and egg question,  the code to make the responses hasn't been written yet because it requires the input which is what you are working on. 

If we assume there is no frequency dependencies on the signal then it will just be a polynomial like response.

ozym:
in some ways we haven't gone round a loop yet so we dont know all the problems yet, which means some of the thinking may need to be revisited.

elidana:
thanks @ozym !
that helps.

I will mull over the components/channel question a bit longer (thanks for the stationXML reference), and hopefully we can keep the values as they are for now.

To complete the EYWM sensor metadata added in this pull request we will add details about:
- temperature response
- proton response

Once we are happy with the numbers, we can merge this, and iterate in a separate pull request the question around creating a response file. Also in relation to evolving bits in other pull requests.

elidana:
temperature added, and details on pin numbers versus fluxgate component added in readme

ozym:
I am working towards a solution, the initial response will have the sensor with a gain of 38000 nT -> mA. This will be converted to V via measuring over a resistor (the driver hardware), the site installed biases applied, and then an overall bias if needed.

i.e. Scott Base Z component.

Original gain: 38000

The calibration will change this to 38023.0

The driver has a resistor of 29.5 kOhms, the  polarity is reversed (it's a northern hemisphere sensor) an offset of 38000 is applied (as it is very far south) it has a bias off set of 0xc1 steps (each step is 1/255).

so the gain field to add is:

scale factor: -1.0/29.5  
scale bias: -193 / 255
absolute bias: -38000

elidana:
have now updated this pr to resolve some conflicts. 
Once https://github.com/GeoNet/delta/pull/1521/ is merged, I'll rework the gains and calibrations file to add absolute bias to gains and check things still make sense.

Thanks @ozym !!

/cc @merijnthornton

ozym:
see https://github.com/GeoNet/delta/pull/1662 for an example

elidana:
thanks @ozym !
I will keep this one open and update once #1662 is merged.
This one will be used to:
- add the further manufacturer calibration info
- add the metadata for the temperature channel (component n. 3)

elidana:
hello @ozym ,

I have now repurposed this old pull request to go alongside #1662, and only adding the temperature sensor information.

I have done some edits to the README in the `calibration` folder to (hopefully) reflect most recent updates.

I have talked to @merijnthornton , and the temperature conversion factors are always the same for a given manufacturer/model. So there won't be a sensor/driver specific temperature calibration factor to consider.

For the  one installed at EYWM (FGE from Danish Met Inst.) the conversion factor is always 10. 

I'm not sure if we want to add these info anyway in `calibrations.csv` , or if this should be just be handled by the
`git/delta/resp/files/sensor_Danish_Meteorological_Institute_FGE_Temperature.xml`

I'm happy either way, and whatever you recommend I can apply here!

elidana:
hi @ozym , just a ping on this one, that is now adding the info about temperature sensor of EYWM

ozym:
I'd also add the 5th channel too (another temperature). We don't mix/match the drivers (and even if we did the calibration would be the same).

elidana:
hope I got it right? have added the 5th (channel number 4) to the install file. still `10` as calibration factor

ozym (approved):

elidana:
thanks @ozym ! I had to resolve 1 conflict, can I please have another tick?

ozym (approved):

PR #1352, opened 2022-05-22
add placenames environment table

ozym:
This is part of the stationxml work to bring as much ad hoc meta data into manageable tables.

This PR provides a list of place names from which sites, or otherwise, can be compared against. It uses a level system which means that small places at large distances a skipped in favour of any nearby larger localities.

This is likely to need expanding beyond the current scope.

### DIFF
--- a/environment/README.md
+++ b/environment/README.md
@@ -10,6 +10,8 @@ _Lists the geological and physical environment details of collection points._
 
 * `gauges.csv` - Gauges Descriptions.
 
+* `placenames.csv` - Placename Descriptions.
+
 * `visibility.csv` - Sky View Descriptions.
 
 #### _CONSTITUENTS_ ####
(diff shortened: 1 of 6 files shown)


salichon:
All looks good to me.
The level is based on the size (or other criteria?)  of the "place" i guess. 
( anyway .. :)  I ll with not argue around the levels when I d subjectively give a 1 to  .. BalClutha )

salichon (dismissed):

elidana:
Hi @ozym , sorry to be a pain, and please disregard this comment if it is off topic.

From what I can understand, those "placenames" are going to be used to populate

elidana (commented):
this looks good, I like the automatic attribution, and the readme/info on how to use the "level" .
I have a comment on terminology.
Would also add an info in the readme around the authoritative source used to define "placenames" coordinates (suppose it is the LINZ database, I can find link and source if it helps)

ozym:
There is no authoritative source, other than us. The locations will have been derived from the old CUSP system.

elidana:
ha ha , ok, fair enough!  sounds good then

elidana:
sorry @ozym , for my other question:
what about using _town_ instead of _placenames_? (to align with terminology used in StationXML). Is it too much of a pain?

ozym:
Town is a little tricky, but "population centre" maybe. or "locality".

ozym:
Maybe just "Name"?

ozym:
At the moment it's only used for: "<Description>NW corner of Yellowstone National Park</Description>" type labels. The name is given in the stations table.

elidana:
"locality" is used and overloaded already. and "name" even more vague.

let's keep placename then.
I see that in our stationXML we also have the following
```
      <Site>
        <Name>Ashburton District Council</Name>
        <Description>within 5 km of Ashburton</Description>
      </Site>
```

and the StationXML standard specify all these geopolitcal boundaries elements , as optional (http://docs.fdsn.org/projects/stationxml/en/latest/reference.html#station-site-town)
<Name><Description><Town><County><Region><Country>

let's keep "placename"

elidana (approved):

PR #1353, opened 2022-05-22
add response labels

ozym:
This is part of the stationxml work to bring as much ad hoc meta data into manageable tables.

This PR provides a list of stream channel codes and types used in _StationXML_ when describing a stream.

I've created a new directory where "response" specific tables may live, it's possible that the channels and calibrations files could live here too eventually.

I've added an Axial column, this is optional and will defer to the Code value if not given. This is used when an installation has been deemed to be "axial", that is not-aligned North/South but at some other axis.

I'm not exactly happy with the name, but I couldn't think of anything else (other than channels, which is taken).

### DIFF
--- a/meta/label.go
+++ b/meta/label.go
@@ -1,0 +1,141 @@
+package meta
+
+import (
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+const (
+	labelType = iota
+	labelSamplingRate
+	labelAzimuth
+	labelDip
+	labelCode
+	labelFlags
+	labelLast
+)
+
+// Label is used to describe a channel stream.
+type Label struct {
+	Type  string
+	Code  string
+	Flags string
+
+	SamplingRate float64
+	Azimuth      float64
+	Dip          float64
+
+	azimuth      string
+	dip          string
+	samplingRate string
+}
+
+// Less compares Label structs suitable for sorting.
+func (c Label) Less(label Label) bool {
+
+	switch {
+	case strings.ToLower(c.Type) < strings.ToLower(label.Type):
+		return true
+	case strings.ToLower(c.Type) > strings.ToLower(label.Type):
+		return false
+	case c.SamplingRate < label.SamplingRate:
+		return true
+	case c.SamplingRate > label.SamplingRate:
+		return false
+	case c.Dip < label.Dip:
+		return true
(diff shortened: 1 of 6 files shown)


elidana:
what about "responsechannel"?
Is a bit long .
Have a meeting now, will look into this as soon as it's finished

salichon:
yup something like @elidana  propose or "responsesSpecs" for specification 
we split the Vertical from the horizontal that s ok.
G is for Geophysics as a Flag ? 
it gives some explicit information of what shoudl be a stream made of.

Would it handle for instance a HH streams that would bechanged over time  as in 100 to 80 sps ?

ozym:
> yup something like @elidana propose or "responsesSpecs" for specification we split the Vertical from the horizontal that s ok. G is for Geophysics as a Flag ? it gives some explicit information of what shoudl be a stream made of.
> 
> Would it handle for instance a HH streams that would bechanged over time as in 100 to 80 sps ?

The sampling rate is one of the fields, simply add a new entry for 80sps and give it the code that's needed.

ozym:
I'm leaning towards "generic.csv"  (I will veto any double barrel names as then it gets into what case is use, camel case, snake case, kebab case ...

The generic bit means, for any sensor of a type X, sampled at Y sps, the default cannel code will be Z given its dip and azimuth inside the sensor casing.  This is not specific to a particular sensor.

It feels a bit like it should be in the installs directory, but there's nothing about it that's time dependent or related to installation, it might not be specifically response related but it is needed for naming stationxml channels.

ozym:
How about `codenames.csv` the allocating of channel codes.

ozym:
This isn't going to work as is. While looking at the fluxgate sensor I've found an edge case where there are two sensors installed at slightly different parts of the sensor (without splitting it out) there needs to be a mechanism to indicate which "pin" is attached to which output.  It looks like a pin number (or equivalent name) is needed.

ozym:
I now remember the work with subsensors, perhaps that's a better approach.

Most sensors don't have subsensors but there are a few that do (dual measurements of different or the same parameters). So the fluxgate sensor would have a fluxgate and two temperature sensors combined into one package.

ozym:
I've found an edge case for an edge case, this is going to get unwieldy I'm going to try and incorporate it into the channels.csv work..

PR #1354, opened 2022-05-22
updated visibility checks

ozym:
Turns out the visibility file wasn't being verified in testing.

Have added checks and updated the file to pass.


meganmadley:
@ozym, slightly confused about the use of the double quotes (" "), for RGCR it was removed but then added for TGOH?

elidana:
Thanks @ozym !!

No need of double quotes. This is some inherited free text file from the old Delta Oracle DB. We will need to work on this at some point and remove the free text.
But double quotes surely unnecessary

elidana (approved):

meganmadley:
Do we want to remove the double quotes for TGOH before this is merged? @elidana @ozym

meganmadley:
Had a look @elidana and there are actually heaps with double quotes, probably could just get this merged now and look at removing when we decide to do a bigger update on this file

meganmadley (approved):

elidana:
yes, sounds good. We can definitively remove double quotes in a separate one. The rework is probably a bit more work than just removing double quotes.
@ozym , I can merge this if you're ok (or feel free to merge it)

ozym:
The double quotes are needed when there's a comma in the message.

ozym:
@elidana merge away.

elidana:
> The double quotes are needed when there's a comma in the message.

:facepalm: oh dear, of course! Thanks for clarifying that

PR #1362, opened 2022-05-26
CPLB loc 22 irrelevant SM streams - Only BN 50 and HN 200Hz exist

salichon:
building array house keeping


mnaguit (approved):
Good catch Sir @salichon 
👍

PR #1364, opened 2022-05-27
Add info about AUCK site vegetation clearing

aleks-spes:
Added records about vegetation clearing at the AUCK site on 2021/03/24.
Is it looks fine and comply with our style of records?


meganmadley (approved):
Looks good to me!

PR #1365, opened 2022-05-31
switch response model make

ozym:
This is a cosmetic change to bring the response details into line with what's contained in the dataloggers.csv file.

This sets the Q330 dataloggers to have a manufacturer or Kinemetrics.

Not sure why the enables are there, perhaps an update to the response yaml files was made but no auto.go was generated.


elidana (approved):
love the cosmetics changes, easy to review ! :)

PR #1366, opened 2022-06-06
RVAZ New Centaur datalogger

salichon:
RVAZ datalogger swap


meganmadley (approved):
LGTM. Just checked SN's and dates, am assuming a 12-day gap between end and start of new datalogger is fine for these sites.

CBurton90 (approved):

PR #1370, opened 2022-06-15
add delta set helper

ozym:
This is an initial mechanism to gather all the various csv based files together into a single structure.

This is aimed at mainly two primary use cases:

- testing 
- CI/CD

The first testing phase allows building a standard call fingerprint for running a given test, e.g.

```golang
var testInstalledAntennas = map[string]func(*meta.Set) func(t *testing.T){
       "check antenna installation overlap": func(set *meta.Set) func(t *testing.T) {
                return func(t *testing.T) {
```

which makes table driven testing much easier (this will be a follow up bit of work).

and for the second part with CI/CD.

The program simply needs to be compiled to have access to the current set of files, basically due to the embedding the files live within the go binary which makes writing programs to build configuration simpler to operate.  As delta is a public repo it means this can all be handled within the go tools (rather than needing git checkout etcs.)

### DIFF
--- a/delta.go
+++ b/delta.go
@@ -1,0 +1,42 @@
+package delta
+
+import (
+	"embed"
+	"io/fs"
+	"sync"
+
+	"github.com/GeoNet/delta/meta"
+)
+
+//go:embed assets/*.csv
+//go:embed install/*.csv
+//go:embed network/*.csv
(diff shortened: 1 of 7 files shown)


CallumNZ (approved):
Looks good

PR #1371, opened 2022-06-15
update span calculations

ozym:
To help with connecting sensors to dataloggers and response information (which all use time spans) these helper functions are available to work out whether two time spans overlap, and also given a bunch of time spans, what is the overlapping window between them, if any.


meta/meta.go
```diff
@@ -161,6 +161,62 @@ type Span struct {
...
+		return false
+	default:
+		return true
+	}
+}
+
+// Extent returns the Span that is the sum of the given overlapping Span values,
+// the extra boolean return value will be false if no window could be found. It
+// is assumed that the end must be greater than the start of the resultant Span.
+func (s Span) Extent(spans ...Span) (Span, bool) {
```
CallumNZ, line 181: The `s` value receiver is a bit confusing - is it needed?

CallumNZ (changes requested):

meta/meta.go
ozym: It's used a bit further down.

```golang
if !clip.Overlaps(s) {
		return Span{}, false
}
```

ozym: there are lots of clips and spans, so it does get confusing.

CallumNZ (approved):

CallumNZ (approved):

PR #1372, opened 2022-06-17
fix response yaml

ozym:
Looks like a missing bit in the yaml file.


salichon:
got it ! 
a miss in the entry. (i could not actually figured out when i tried  to "fix") 
interstignthat the final sets did not complain about a missing datalogger entry with the combo/ 
All good thanks!

salichon (approved):

PR #1374, opened 2022-06-22
add auto xml snippets

ozym:
This looks big but it isn't as bad as that.

The aim is to add a little bit more functionality in the response generation code (the part that builds the `resp/auto.go` file) to build another set of "snippet" XML formatted files (in `resp/auto`) for including in future response work.

The layout is similar to the NRL mechanism, the various "snippets" can be joined together on the fly to build the required sensor/datalogger StationXML response stages.

There is also a directory `resp/files` which can contain hand rolled, or externally derived response files.  I am anticipating adding a `resp/nrl` to store the NRL derived equivalents.

The future direction of development is in using these snippets rather than relying on the auto.go mechanism (although the yaml based configuration will still have some use as per this PR).

The major requirement has been to include the stationxml code for v1.1 as internal. There is some debate as to whether this should actually live in the main directory so that it can be used for other applications.


internal/stationxml/v1.1/decimation.go
```diff
@@ -0,0 +1,42 @@
...
+func (d *Decimation) SetInputSampleRate(rate float64) {
+	d.InputSampleRate = Frequency{
+		Float: Float{
+			Value: rate,
+		},
+	}
+}
+
+func (d Decimation) Copy() Decimation {
+	return d
```
bpeng, line 21: This may not make a copy

internal/stationxml/v1.1/email.go
```diff
@@ -0,0 +1,23 @@
...
+	Address string `xml:",chardata"`
+}
+
+func (e Email) IsValid() error {
+
+	if !(len(e.Address) > 0) {
+		return fmt.Errorf("empty email")
+	}
+
+	if !(regexp.MustCompile(`^[\w\.\-_]+@[\w\.\-_]+$`).MatchString(e.Address)) {
```
bpeng, line 18: the `regexp` could be a `const`

internal/stationxml/v1.1/gain.go
```diff
@@ -0,0 +1,21 @@
...
+// FrequencyEnd) in which the SensitivityValue is valid within the number of decibels specified in FrequencyDBVariation.
+type Gain struct {
+	// A scalar that, when applied to the data values, converts the data to different units (e.g. Earth units)
+	Value float64
+	// The frequency (in Hertz) at which the Value is valid.
+	Frequency float64
+}
+
+func (g Gain) Copy() Gain {
+	return g
```
bpeng, line 15: may not be a copy

internal/stationxml/v1.1/phone_number.go
```diff
@@ -0,0 +1,33 @@
...
+	if !(len(p.PhoneNumber) > 0) {
+		return fmt.Errorf("empty phone number")
+	}
+
+	if p.AreaCode == 0 {
+		return fmt.Errorf("no area code")
+	}
+
+	if !(regexp.MustCompile(`^[0-9]+-[0-9]+$`).MatchString(p.PhoneNumber)) {
+		return fmt.Errorf("bad phone number: %s", p.PhoneNumber)
```
bpeng, line 29: make `regexp` a `const`

internal/stationxml/v1.1/root_marshal_test.go
```diff
@@ -0,0 +1,69 @@
...
+		if err != nil {
+			t.Error(err)
+		}
+
+		var s FDSNStationXML
+		err = xml.Unmarshal(x, &s)
+		if err != nil {
+			t.Error(err)
+		}
+
```
bpeng, line 59: Do you want to check if `s == test.r`?

internal/stationxml/v1.1/data_availability_span.go
```diff
@@ -0,0 +1,42 @@
...
+
+	// The maximum time tear (gap or overlap) in seconds between time
+	// series segments in the specified range. </xs:documentation>
+	MaximumTimeTear *float64 `xml:"maximumTimeTear,attr,omitempty"`
+}
+
+func (d DataAvailabilitySpan) IsValid() error {
+
+	if v := d.NumberSegments; v < 1 {
+		return fmt.Errorf("invalid number of segments: %d", v)
```
bpeng, line 31: can be `d.NumberSegments < 1`?

bpeng (changes requested):
Looks goos and tidy, small changes

internal/stationxml/v1.1/decimation.go
ozym: It's fine as a copy given that the struct doesn't hold any slices or pointers. (this code will be replaced in a new PR soon).

internal/stationxml/v1.1/email.go
ozym: yes, although that just moves it elsewhere in the code. Again, this code will be removed as I will avoid doing validation prior to encoding (will rely on validators after building).

internal/stationxml/v1.1/gain.go
ozym: it's fine as a copy it's not a pointer reference and the struct doesn't contain slice or pointer values.

ozym: Also will be replaced.

internal/stationxml/v1.1/data_availability_span.go
ozym: but then I don't have the v as a nice short variable for the error message (just an idiosyncrasy of mine, much like my avoidance of >= and <=).

ozym:
See #1387 for context

ozym:
@bpeng can you please have another look.

bpeng (approved):
Code looks tidy and tested working

PR #1375, opened 2022-06-22
Adding new receiver (alloy) for GNSS TGHO

meganmadley:
The NetR9 died around July last year and I have used this time as the closing time. Please let me know if you'd prefer for me to use 1 second before the alloy started.


aleks-spes:
I have no strict opinion about this problem, but here is my thoughts:
1) about session.csv
Session.csv contains general info about "data files" which we collect from the sites. 
Here is a question: Are we sure there is no data files on NetR9 after 2021/07/16 03:03:59? If so we can use it like  'Session stop time' because we don't expect any new data, otherwise we should extend the period of time to cover "lost' files.
2) about receivers.csv
I would say it is better to use "1 second before style" record for "Deployment stop time" because the NetR9 was there during all this time untill Alloy installation.

What do you think?

meganmadley:
Hmm from that, I reckon we extend all end dates to 1s before the start of the alloy and then once we get the files back we can modify the session file if needed, what do you think @aleks-spes?

CBurton90 (approved):

aleks-spes:
Sounds good!

aleks-spes (approved):
Looks good for me!

PR #1377, opened 2022-06-23
new sensors at TO003

rumachan:
Changes to account for SP and associated temperature sensors


ricardojohansson:
I'm not sure if the reference electrode has been left out intentionally(?)
Just to clarify; there are four electrodes installed allowing three SP measurements, ie between the reference electrode and each of the three remote electrodes.

ricardojohansson (approved):

PR #1378, opened 2022-06-28
Add Scott Base specific download rate limits.

junghao:
This implements the discussion of https://github.com/GeoNet/tickets/issues/11277#issuecomment-1168002194

Should be fine if merged before the rate limit unit change https://github.com/GeoNet/gloria/pull/377, because this only affects sites at Scott Base which don't have rate limit ATM. 

If the unit hasn't changed for download side then it'll read as 1024KB which is almost the same as no limit.


ozym (approved):
Looks fine, other than SCTB the mark codes listed are for geomag sites.

I assume the removal of spaces is a tidy up to the code and will not impact what is used downstream as per formatting.

PR #1381, opened 2022-06-29
Gloria Scott Base metadata fix

junghao:
Previous update caused sites in Scott Base became non-IGS sites.


junghao:
This PR is wrong.

PR #1385, opened 2022-07-01
DREZ datalogger swap

mnaguit:
- from Q330/3 SN2334 to Centaur SN9295
- Centaur firmware v4.5.20
- For LC 10 with EH-ENZ streams


salichon (approved):
sweet

PR #1386, opened 2022-07-01
KHEZ datalogger swap

mnaguit:
- from Q330S/6 SN5439 to Centaur SN9290
- for LC 10 with HH/LH-ENZ streams (LC31 to LC33 also impacted)
- Centaur firmware v4.5.20


CBurton90 (approved):

salichon (approved):
ok :)

PR #1387, opened 2022-07-06
add stationxml schema decoding into golang modules

ozym:
With the recent updates to the FDSN StationXML schemas keeping up to date with new versions has required a reworking of the underly StationXML coding mechanisms.  This PR switches to using an automatically generated version module based on the XSD file that is used for validating the XML files.

This isn't a general schema decoder, I looked at a number of off-the-shelf decoders and all had good and bad points when it came to these particular schemas.  In all cases the amount of work to fix the encoders was about the same as writing one from scratch. However, the "aqwari.net/xml/xsd" package provided enough of the complex decoding that adding the tailored remaining code generation was more manageable.

This PR follows the development steps through each commit,  it starts with decoding the schema.  Then switches to encoding golang files based on type names and a couple of special types (the base one and one for date time).

The next steps follow the different schema types and their conversion into code.

The final part of the jigsaw is to build the `go:generation` which allows the programs to be run over different schema versions.

One question was whether these should be under `internal` or not. I have left them there as they are automatically generated,


junghao (approved):
Tested the code, works pretty well.

PR #1388, opened 2022-07-07
add fox sites lrdcp assets

rumachan:
Jackson,

Have added LRDCP assets for four sites (other than FOXAG). Haven't yet installed as want to check once comms up and logger code finalised.

Assets are as specified in CSV files in sit/data/assets.

Can you please review. Thanks


jacksonshanks (approved):
looks good. appologies for not doing this myself!

PR #1393, opened 2022-07-11
RLTN antenna and receiver swap

meganmadley:
Will wait for a photo from @timjmcdougall to confirm the new antenna serial number before merging


meganmadley:
Tim didn't take a photo, so will do some further checks after it has been merged

aleks-spes (approved):
Looks good!

PR #1394, opened 2022-07-12
Stationxml streams

ozym:
A part of the response generation work to be a little more explicit about where codes come from.


ozym:
This is follow on work from the connections file updates.

This PR uses the FDSN Channel Codes document as a starting point and adds the first two parts, the "Base", and the "Source" to the streams table.  These are essentially H and N respectively for the traditional channel code of HNZ, when joined with the sensor information the full channel code can be constructed.

Using this convention will have the roll-on affect of changing the label in the Gains table from "Channel" to "Subsource" as per the convention, which will itself suggest that SubLocation be renamed Sublocation for consistency.

This PR will alter all rows in the streams.csv file so will easily be in conflict following regular delta updates.

salichon (commented):
Thanks @ozym 
this is great for the sake of a clear statement of the tyope of streams in addtion to the site codes.

What implies the abscence (blanks) of  the source and channel declaration  in streams.csv ? 
The `blank` information points to a default configuration type ?      

For instance not having it for (some) building arrays or DART systems  or other type of instrument such as the WNVZ tiltmeters.

I guess this would need another set of eyed to bless the new entries in streams.csv. (with no too much worry :) )

ozym:
The blanks are caused because we don't have the response information for all channels (building) so the automatic filler-in process has nothing to use.

CBurton90 (approved):
Where is the code that decides what channel codes the stream entries should have? I assume this code is looking up the sensor/recorder model name for a site in the `/resp` folder and grabbing the appropriate channel codes from there?

salichon (approved):
We ll need to 
publicize this "This PR will alter all rows in the streams.csv file so will easily be in conflict following regular delta updates."
in ordre to merge the current ones 
before we merge that one ;)  (basiacally only the `streams.csv` users right? 
cheers
FYI @meganmadley @mnaguit at least :)

ozym:
I was intending to wait until @elidana had a look over it given her interest and previous comment about this. So not a hurry at the moment.

salichon:
Kia Ora @ozym 
i am  browsing the changes  this one beloow lloks odd. 
 **CMWZ,20,H,H,200** 
 Besides, 
  the overall main  change for channel and component.
 We may have to eye ball each stream to confirm (over another PR? - or whenever you fancy the most) it might take some attention and time  :) .
 cheers
jerome

ozym:
@salichon that's just test data. I can make it authentic if that helps.

ozym:
@elidana this is one that I have been waiting for your input. It essentially adds the "Band" and the "Source" to the streams channel.  This allows building the miniseed style label easily by adding the sensor code (i.e. Subsource).

ozym:
The input data is not used at the moment, and where there are gaps reflects streams which are not producing any output in FDSN currently.

salichon (approved):
all good that s cool

salichon (approved):
ok  for the updated streams :)

elidana (approved):
thanks @ozym , this one looks pretty good to me and I'm happy for it to be merged (provided we properly socialize it and update our internal documentation related to metadata updates).

A couple of comments :

- I like that names used here correspond to FDSN channel specs (band - source - subsource). I wonder if we need to change to a different "name" the term subsource used in the install/gains.csv, as those will now have a different meaning 
- the nominal sampling rate column will now assume a slightly different/more constrained "meaning" (given that the "band" indicates the general sampling rate and response band). Just for clarity, I would add some more details to the README, will propose something on this pr on that
- as @CBurton90 mentioned above, how can we ensure that all the info in install/streams.csv are correct? happy to run a few checks against the archive if that helps at all

ozym:
@elidana thanks for the review, here are some follow up comments:

- The gains.csv has already been updated in a previous PR
- The sampling rate is required as the band is only an indication of the sampling rate, not the actual rate (e.g. HNZ means 200 Hz as well as 100 Hz) etc. This will be more important when the response is needing to match exactly (a future PR)
- The current streams is built from a program that takes a streams csv and compares that with an FDSN query (this is where the sampling rate is important) (impossible to do this otherwise really).
- This will not have an impact until we implement the stationxml code, and then we can check if anything's missing.

Merging this has some downstream issues so needs to be coordinated, it will also add conflicts to any PRs that have changes to streams in them.

PR #1395, opened 2022-07-13
rename gains and features columns to be consistent with streams

ozym:
This follows on from the updates to the streams table column names.

Using the FDSN Channel naming convention ("HNZ" -> "Base" + "Source" + "Subsource") what is referred to in the current `gains.csv` file as "Channel" should now be called "Subsource" to refer to the same thing (or at least a plural version of the same thing).

But, having "SubLocation and Subsource" together is no longer consistent, as the Subsource is a refered standard, the proposal here is to rename "SubLocation" to "Sublocation".

This has implications for the `features.csv` table as well.


salichon (approved):
It looks okay to me thanks !
from my point of view The _Sublocation_ field relates to single component.

In the instrument response building part 
 this sublocation  will be related to the pins and pin offset right?

ozym:
The sublocation is related to a sensor that measures multiple outputs (e.g. pressure + temperature).  The sensor is installed at the location but the sublocations allow descriptions in the environment/features table.

CBurton90 (approved):

salichon (dismissed):
good to me

ozym:
@rumachan can you have a look as it touches the features file, but it's only renaming the column header.

rumachan:
Thanks Mark, I'm all good with this. Howard uses both features and gains for ingestion into Tilde. Am mentioning it to him in case it affects the LRDCP ingestion code.

junghao:
@ozym Could you please put this PR on hold for a few days so I can make changes to LRDCP accordingly.

ozym:
@junghao there's no rush, just getting the ducks in a row prior to other upcoming changes.

ozym:
@rumachan when this get's approved again, and merged. I will work with @junghao to vendor in the changes and updates as needed.  Because this isn't changing any files (other than a header line, which is ignored) there shouldn't be any interruptions to the lrdp config.  [The changes will need to happen once the delta code is vendored, at that point it takes over]

rumachan (approved):
Thanks for the reassurance about impact on LRDCP.

PR #1400, opened 2022-07-24
add volcano camera networks

ozym:
This is just getting things lined up for non-volcano cameras. I've just made some strawman network codes and labels.

Please adjust/review as needed.


rumachan (approved):
I like that idea a lot. What code are you suggesting for Landslides? Maybe LS?

ozym:
Actually LC (Landslide camera network, but it could be anything).

PR #1401, opened 2022-07-28
TO003 SP reference electrode

rumachan:
Adding necessary changes for the reference SP electrode at TO003. This is to document that electrode is installed, but isn't need for anything downstream.


ozym (commented):
looks good, except I'm not sure you need the stream as there isn't actually a direct data stream related to this sensor

rumachan:
Thanks, removed the stream entry

ozym (approved):

PR #1402, opened 2022-07-29
add volcano camera config tools

ozym:
This PR adds two config tools for camera images.

The first is used simply to caption an image if needed, it uses the mounts and views tables.

The second is related to volcams, so adds in some extra detail regarding actual volcano targets.

### DIFF
--- a/cmd/volcam-config/main.go
+++ b/cmd/volcam-config/main.go
@@ -65,76 +65,77 @@ func main() {
 	}
 
 	var volcs []Volcam
-	for _, l := range set.Views() {
-		if time.Since(l.Span.End) > 0 {
+	for _, view := range set.Views() {
+		if time.Since(view.Span.End) > 0 {
 			continue
 		}
 
-		for _, m := range set.Mounts() {
-			if time.Since(m.Span.End) > 0 {
+		for _, mount := range set.Mounts() {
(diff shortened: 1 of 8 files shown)


cmd/volcam-config/main.go
```diff
@@ -0,0 +1,155 @@
...
+
+	flag.Parse()
+
+	set, err := delta.NewBase(base)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	var volcs []Volcam
+	for _, l := range set.Views() {
```
CallumNZ, line 68: An alternative naming suggestion:
`for _, v := range set.Views()`
`for _, m := range set.Mounts()`
`for _, c := range set.InstalledCameras()`
`for k, vals := range keywords {`
`for _, val := range vals {`
`for k, val := range targets {`

CallumNZ (commented):
Looks good, just a naming suggestion.

cmd/volcam-config/main.go
ozym: Good suggestions, I've aimed to make the variables a little more readable. Have gone for longer form where it's a struct being returned and shorter for looping like variables.

ozym:
@CallumNZ have updated variable names etc.

CallumNZ (approved):

PR #1412, opened 2022-08-10
GNSS QUAR receiver swap

meganmadley:
QUAR has power! So we can now do our metadata changes for this receiver swap.


CBurton90 (approved):

aleks-spes (approved):
All good!

PR #1416, opened 2022-08-15
MRZ datalogger-sensor swap

mnaguit:
- from Q330/3 (1188) to Centaur (9291)
- from Guralp (T3E73) to Trillium Horizon (824)
- Centaur firmware v4.5.20
- for LC 10 HH(ENZ)


meganmadley (commented):
Hey @mnaguit, got some questions(/possible changes) about the install/firmware file:
- why is the start date for `9291` not 2022-08-12T00:51:00Z?
- also for `9292` why did you change it from 2022-08-03T22:47 - that was the right date in the last PR (I think) 

So I think for `9292` that should be changed back to `2022-08-03T22:47:00` and for `9291` it should have `2022-08-12T00:51:00` right?

mnaguit:
> Hey @mnaguit, got some questions(/possible changes) about the install/firmware file:
> 
> * why is the start date for `9291` not 2022-08-12T00:51:00Z?
> * also for `9292` why did you change it from 2022-08-03T22:47 - that was the right date in the last PR (I think)
> 
> So I think for `9292` that should be changed back to `2022-08-03T22:47:00` and for `9291` it should have `2022-08-12T00:51:00` right?

Definitely. Looks like my delta expertise is becoming rusty! 😯 
Thanks loads @meganmadley , will update that csv

meganmadley (approved):
Perfect! Thanks :)

CBurton90 (approved):

PR #1420, opened 2022-08-17
DART NZB, NZF, NZG, NZH, NZI Redeploy Equipment Change

meganmadley:
Biggest question on this: how accurate do we want our start and end dates? We could probably get the first and last values from beta-data/tilde if we want to be within 15 minutes. Otherwise have got them to a day e.g. starting on 2022-08-15T00:00:00Z, previous deploy would then close on 2022-08-14T23:59:59Z. 

Source of truth is the deployment logs stored here: J:\_Nat_Hzds_Risks\DART\Deployments and Maintenance\Voyage 5_Maintenance_Jul 2022\Deployment Logs (info also in main ticket and wiki page)


meganmadley:
The start and end dates could be updated to be more accurate, I've gone through beta-data N1 files for both primary and secondary and reckon we could have the following:
| DART| end | start | reason |
| ---| ----|----|----|
|NZB | 14-08-2022T18:00| 15-08-2022T20:15 | lost S in May 2022 - used P for end date |
|NZF| 09-08-2022T18:00 | 11-08-2022T00:15 | lost S in March 2022 - used P for end date |
|NZG| 19-03-2022T05:00 | 30-07-2022T20:15 | lost S in Jan 2022, P in March 2022 (used P for end date)|
|NZH| 15-12-2020T14:00 | 01-08-2022T20:15 | lost both in Dec 2020 (S came back an hour before P, used S here)
|NZI| 07-05-2021T04:00 | 03-08-2022T21:15 | lost S in March 2021 - used P for end date|

@ozym and @salichon shall I update the PR with these more accurate times before you do a review?

ozym:
There were problems with this the last time as the first sample recorded wasn't really usable as it was related to either going up or down.

I like the QC side of things more, where we want to have times when we know the data was good to use (i.e. the sensor was on the ocean floor) in theory data outside of this time if collected could be used but would need an extra "quality" mechanism.

meganmadley:
Have just updated the start and end dates for install/firmware, install/recorders, install/streams, network/sites, we may decide to change some of these ones we get the BPR data back but for now they are pretty good.

meganmadley:
@salichon we now have a gauges file :)

salichon (commented):
a  few WIP  comments  nothing major:
- DART systems (`gauges.csv`)  do we need  _Crex Tag | Tide gauge Crex format location_ 
As opposed to Tide gauges I think it is for the bulletin creation which is accounted for by the CDDS  .. but not quite certain
- the file `firmware.csv` is keeping the previous BPR firmware versions opened for the previous deploy.
not a big deal - Do we need to we close them accordingly ? 

The other changes are good to me

meganmadley:
Sweet so just putting this down:
- `gauge.csv` I think we don't need to worry about those for now as we have other ways of identifying the DARTs
- `firmware.csv` I like to not close these for now as we have no way of knowing if the FW is going to change

@ozym just pinging you for a review so we can get this merged :)

Also noting we may end up changing some of these details once we have the BPR data but that could still be weeks away so best to get this done now and do another PR when we have those details

salichon (approved):
okay ;)

ozym (approved):

PR #1422, opened 2022-08-17
GNSS OPTK monument height correction

aleks-spes:
Add correct monument height for GNSS site OPTK to network/monuments.csv


meganmadley (approved):
LGTM

PR #1424, opened 2022-08-21
add citations poc

ozym:
The work on polarities has indicated that there will be a need for some mechanism for citing references or other associated external information.

To this end, I've made an example "citations.csv" formatted table. I've placed it under the `references` folder for want of anything else.  I haven't used the word "reference" in the names as this clashes with the use of this term for defining where something is, i.e. with reference to this point.

I've also started looking at adding a DOI mechanism. This may be expanded as needed or appropriate to other tables. especially the network table.

This mechanism should also be able to present "web" type links so, other than the key, all fields should be optional.

I haven't defined a "key" style, but the constraints are that it be unique, and the table will be sorted based on this value. I was thinking of a short style, or "<author><year><series>" where author can be a lowercase or abbreviated name, the year is optional and the series (e.g. "a", "b", "c") may be able to distinguish multiple publications in a given year.

This is aimed as a starting point of discussion. the code has been written as a proof of concept but not included.

### DIFF
--- a/references/README.md
+++ b/references/README.md
@@ -16,9 +16,9 @@ All fields, other than Key, are optional and can be left blank.
 | Field | Description |
 | --- | --- |
 | _Key_ | The citation key used where appropriate
-| _Author_ | Reference author or authors in a natural format, with quotes needed as required
+| _Author_ | Reference author or authors in a natural format
 | _Year_ | Reference year, if appropriate
-| _Title_ | The title of the reference in a natural format, with quotes as required
+| _Title_ | The title of the reference in a natural format
 | _Published_ | Where reference was published if appropriate and known
 | _Volume_ | Series information for published reference if relavent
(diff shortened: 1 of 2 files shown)


mnaguit (dismissed):
Thanks @ozym . This is great! 👍

rumachan (changes requested):
Only suggestion is a little bit more explanation of when quotes might be required, i.e. when the text, usually the title or author list, contains commas.

ozym:
@rumachan I've updated the README to reflect the comma situation.

rumachan (approved):
All good

salichon (approved):
all good to me for this template

mnaguit (approved):

staylorofford (commented):
Looks good. Just wondering where the citations for this table will occur in the metadata or delta CSVs?

PR #1409, opened 2022-08-10
Update SEISMIC_SENSOR_ORIENTATION_METHODS.md

staylorofford:
Update orientation conventions and methods to reflect current understanding

PR #1426, opened 2022-08-22
add polarity poc

ozym:
This is following on from instrument response work, and trying to avoid hidden logic in the instrument meta data generation.

This table is basically attempting to identify times when a recorded signal may be reversed from what is expected. How this information is gleaned is left as a reference where not straight forward.

I haven't added the full range of methods yet. I haven't decided whether to (a) build this into the code directly, or (b) create a table (perhaps in the references section) which can define methods of doing things. This can be extended to other tables or processes.

Just wanting feedback, on whether this covers the bases, whether there's something missing or otherwise. The code is there to implement this (as a proof of concept) but it doesn't need to be part of this initial documentation PR.

Views on the "method" table would be useful.

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -18,6 +18,7 @@ Meta information for the GeoNet equipment network.
 * `dataloggers.csv` - Recording dataloggers
 * `connections.csv` - Datalogger and sensor connection details
 * `streams.csv` - Datalogger and recorder sampling configurations
+* `polarities.csv` - site specific polarity settings which indicate when a site may have reversed polarity, or otherwise
 * `gains.csv` - site specific settings applied to individual datalogger and sensor that may impact overall sensitivities
 * `calibrations.csv` - Individual sensor sensitivity values that can be used rather than default values.
 * `components.csv` - Individual sensor elements including measurement position and responses.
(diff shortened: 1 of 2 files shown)


install/README.md
```diff
@@ -219,6 +220,47 @@ A list of _datalogger_ sampling configurations for a given _station_ and recordi
...
+
+| Field | Description | Units |
+| --- | --- | --- |
+| _Station_ | Recording _Station_|
+| _Location_ | Recording location _Site_|
+| _Sublocation_ | Recording location _Sublocation_|
+| _Subsource_ | Recording location _Subsource_|
+| _Primary_ | Whether the entry takes precedence| _"yes"_, _"no"_, or blank.
+| _Reversed_ | Whether the stream in the time window should be considered reversed or not| _"yes"_ or _"no"_
+| _Method_ | How this information was obtained| _"study"_, _"compass"_, or _"unknown"_
```
staylorofford, line 245: Some elaboration on what this means could be useful

staylorofford: i.e., "compass" shows sensor alignment, and if data are reversed then polarity...

install/README.md
```diff
@@ -219,6 +220,47 @@ A list of _datalogger_ sampling configurations for a given _station_ and recordi
 | _Start_ | Stream start time|
 | _Stop_ | Stream stop time|
 
+#### _POLARITIES_ ####
+
+Site specific times when the recorded values may have a reversed, or otherwise, polarity.
+This is often difficult to track with the known instrument responses and installation details as it can
```
staylorofford, line 226: Why couldn't this go in the sensors.csv? Alongside information like azimuth, dip?

install/README.md
staylorofford: Also, is this where the references files (https://github.com/GeoNet/delta/pull/1424/files) could be included under a citation column?

staylorofford (changes requested):
Simple review. Looks good, just a few points to clarify and consider

install/README.md
staylorofford: Method could also go in long-form but high-level documentation, as we have for orientations.

install/README.md
salichon: hello  @staylorofford  .. :) 
i would say because it doesn't always has to deal with the sensor but also the datalogger or the cables (or unknown forgotten  reason for a reversal...)

salichon (commented):
Is this polarity information unique to seismic streams ?
We shall perhaps specify what type of observations this polatiry covers in the readme ?

salichon (commented):
Note: i am  not willing to make this complicated (!)  though informative enough ... 

- The _Method_  information  may differ according to the instrumentation
- Would it be relevant to detail  more the  _Study_  information when known ( published reference)? 
- As well this might be useful  to  potentially inform if the study was _"discrete"_ (event based for instance) or spans a longer  time over a _"continuous"_ ( >1day) set of data     
- the _Primary_ part is good and shows possible undetermination. 
Would this be relevant to add  an information field as  _Confidence_ with  _"Confirmed"|"Likely"|"Weak"_ ?

install/README.md
ozym: I'm not 100% sure where to go here.

Either we keep it as is (i.e. the polarity is in the streams, as the reversed column) and have this table more of formal-ish documentation as to why.

Or we switch completely over, and the polarity table is used (which then introduces the problems related to possibly inconsistent or conflicting data).

staylorofford: It makes sense to have this separate, as @salichon says. Can we have this table as the source of truth and keep the other files free of polarity information?

staylorofford (approved):
Following discussion with @salichon and @ozym, I am comfortable with this change now that we have clarified conventions around orientations and polarities, as described in https://github.com/GeoNet/delta/pull/1409/files

ozym:
@staylorofford can you have another look, I had to fix a conflict in the README (no changes other than that).

staylorofford (approved):

PR #1428, opened 2022-08-24
add stations with multiple networks

ozym:
This is a proof of concept that would allow a single station (e.g. a station code) to be associated with multiple networks over different time windows.

This is useful for the case where a datalogger is installed at a site, but also at the site some manual data was collected, or perhaps historical data was recorded.


ozym:
I've needed to rework the station sorting, and to update the networks and stations tests.

elidana:
This one looks good, but I have a question in relation to other domains.

- geodetic stations will still need to have a unique ID. This pr is not affecting those as the table is a different one, and tests are still ensuring uniqueness. So all good from that perspective.
- what about seismic stations? I've tried to add a stationID duplicate (same entry, only a different network code), and this is now possible. Is that ok? Can we have two seismic stations with the same ID but a different network code?
- what about other domains that are using the same set of tables? (envirosensors, dart, etc).

Thanks!!

ozym:
Actually I've bumped into this before, so for instance a seismic site in the wellington network (traditionally WL) is mapped to NZ (external network) but when building stationxml it will be under a network level, and in that case it will be NZ which is the national network. This has a different meaning from WL, which sort of breaks down.

One option, for seismic stations, is to make sure the  station and network "external" codes are unique.  I.e. can only have one station X with a network Y, that has an external code Z.  (so network B, with external code C, would fail).

ozym:
The base reason for this is to allow a station to be in multiple tilde domains.

ozym:
An example:

`domains.csv` (or what makes sense)

```
Station,Network,Subnetwork,Start Date,End Date
WI227,EN,MC,2018-09-13T00:00:00Z,9999-01-01T00:00:00Z
```

This indicates the normal station (WI227 EN) is associated with the subnetwork (or domain) MC between the given times?

ozym:
This way WI227 is unique, but has the ability to be linked to other domains.

ozym:
We could also introduce "domain" as a loose grouping of stations (much looser than a pure network definition).

ozym:
This adds a "domains" construct, I haven't removed the pervious updates but that would be needed too.

rumachan:
So domains.csv would looks like this in our case? Using the domain names Tilde uses. I don't think I'm following this correctly
```
Station,Network,Domain,Start Date,End Date
WI227,EN,envirosesnor,2018-09-13T00:00:00Z,9999-01-01T00:00:00Z
WI227,MC,manualcollect,2003-06-01T00:00:00Z,9999-01-01T00:00:00Z
```

ozym:
There's two ways of looking at it, the first is to not explicitly reference tilde. This then relies on tilde knowing that to fine envirosensor meta data, it needs to look for station/sites in the EV network.

Whereas, with the manualcollect, it will need to look for domains entries, which have a domain "MC".

In this case the domains entry is purely:

```
Station,Network,Domain,Start Date,End Date
WI227,EN,MC,2003-06-01T00:00:00Z,9999-01-01T00:00:00Z
```

It's looking up the station WI227/EN and using that as the metadata related to location etc.

The second approach is to have domains actually related to tilde, in which case the MC part is redundant. So:

```
Station,Network,Domain,Start Date,End Date
WI227,EN,envirosesnor,2018-09-13T00:00:00Z,9999-01-01T00:00:00Z
WI227,EN,manualcollect,2003-06-01T00:00:00Z,9999-01-01T00:00:00Z
```

links the station to the tilde domains explicitly.
 
---

I had assumed the first method, but the second is worth exploring in as much as how does tilde fold into delta (currently it just knows what it needs from delta).

rumachan:
Now I get what you meant. Thanks.
```
This then relies on tilde knowing that to find envirosensor meta data, it needs to look for station/sites in the EV network.
Whereas, with the manualcollect, it will need to look for domains entries, which have a domain "MC".
```
For envirosensor Tilde looks for `network`, but for manualcollect it looks for `domain`. 

Its just that with this (below) the entry in delta is a bit confusing. It suggests that WI227 is in the EN network and the MC domain, which it isn't as its also in the envirosensor domain.
```
Station,Network,Domain,Start Date,End Date
WI227,EN,MC,2003-06-01T00:00:00Z,9999-01-01T00:00:00Z
```
In terms of clarity for someone editing and reading delta this is better, I think
```
Station,Network,Domain,Start Date,End Date
WI227,EN,envirosesnor,2018-09-13T00:00:00Z,9999-01-01T00:00:00Z
WI227,EN,manualcollect,2003-06-01T00:00:00Z,9999-01-01T00:00:00Z
```
If we don't use MC at all (as a network) then we can still use it as part of the location (MC01, MC02, etc) to distinguish from the LRDCP sensors at a site.

ozym:
The only caveat to that is that this will break when we have non-stations wanting to be part of a domain. e.g. marks 

So either this is excluded (and which case we go back to tilde needing to know what to do) or the domain is changed to "Code" perhaps (which encompasses either marks or stations).

elidana:
I think adding the detail around the "domain" at site level is a bit redundant. If we need to add this information here, what if use the network table instead?

this will also solve possible issues with multiple network codes that pertain to the same domain. 
Something like

```
Network,External,Domain,Description,Restricted
EN,NZ,envirosensor,GeoNet environmental data using low rate data collection platform network,false
GM,NZ,geomagnetic,New Zealand geomagnetic observatories network,false
LI,LI,gnss,The LINZ national CGPS network,false
MC,MC,manualcollect,manually sampled data,false
```

ozym:
The problem is that we have stations (marks in the future) that might be in multiple domains. Say a site on WI227 has a datalogger so it's in the EV network.  But manual sampling at the site means it's in the MC network, it keeps the same station code (which is the crux of this problem). So the network can't really match the domain as one-to-one.

ozym:
I'm heading towards keeping tilde at arms length, and letting it decide what logic is required to update its meta data.  The only issue is then having some easy way of identifying (or mapping) this logic into the delta tables.

rumachan (approved):

rumachan:
I approved the PR request. I think we are at the stage where we'll learn more by running with what we've got and seeing where the problems arise, than thinking about it any further. I can selectively add some meta-data and we can continue the discussion from there.

PR #1430, opened 2022-08-24
GNSS NLS3 adding new landslide site

meganmadley:
Same as NLS4, may want NLS4 reviewed and merged before reviewing this one as they will conflict


CBurton90 (dismissed):
For the monument I have include a photo showing the Leica tripod, we could potentially estimate the height with one at Avalon. Anyhow I think the entry is fine as it is based on the earlier Nelson Landslide Responses (1 & 2).

aleks-spes (dismissed):
All good!

aleks-spes (approved):

PR #1432, opened 2022-08-26
NLS3 and NLS4 adding monument height

meganmadley:
Adding vertical antenna heights. Not sure if we want to add "Leica Tripod" or not. 

https://github.com/GeoNet/tickets/issues/11719 and https://github.com/GeoNet/tickets/issues/11720


CBurton90 (approved):

aleks-spes (approved):

aleks-spes:
I think that we don't need to specify " Leica tripod".

PR #1446, opened 2022-09-12
Update firmware.csv

ricardojohansson:
updated CR6 04117 to CR6.Std.11.02 prior to install at TRAR1


rumachan (approved):
Yes, this is the correct firmware version in the relevant data files

PR #1040, opened 2021-04-19
site classes round two.

ozym:
this has been rolled over from GeoNet/delta#164

PR #1447, opened 2022-09-17
update classes

ozym:
This is an update of #1040 but with only an update of the README and a slight reordering of the csv columns.

I think this should be available as a reference, it will need to be updated and maintained but at least it is a start. I have added a citations column for use with references as they are converted from the notes column.  I expect this column to be removed, assuming the entries can be condensed into a set of states.

### DIFF
--- a/environment/README.md
+++ b/environment/README.md
@@ -4,6 +4,8 @@ _Lists the geological and physical environment details of collection points._
 
 ### FILES ###
 
+* `classes.csv` - Site Class Descriptions.
+
 * `constituents.csv` - Constituents Descriptions.
 
 * `features.csv` - Sensor Installation Descriptions.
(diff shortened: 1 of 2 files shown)


mnaguit (approved):
Hi @ozym . Thanks a lot for working on this, it is indeed a great start capturing the site class descriptions in our metadata. Very timely too as per recent GeoNet expansion workshop, where influence of soil conditions and/ geotechnical investigations were mentioned a couple of times for future network expansion. 

There is a recent paper on NZ Vs30 map (Foster et al, 2019), which we can refer to later. A github code repo is available for this (https://github.com/fostergeotech/Vs30_NZ). 

At this point, I'm happy with _classes.csv_ We can fine tune it as we move along.

salichon:
@samgeonet
 this is a PR with the ste class that can be good to be aware for your sites install

salichon (approved):
This  can be updated with the collaboration  of Anna and Elena 
cool

mnaguit:
> @samgeonet this is a PR with the site class that can be good to be aware for your sites install

Maybe @staylorofford ? 😄

elidana:
Hi @ozym ,

this one looks good. How is this related to site class already distributed in the data repository? https://github.com/GeoNet/data/tree/main/site-class

Shall we migrate things from that repo to here on delta?

ozym:
@elidana this code came out as part of the follow up work from that repo (has taken a while and a few iterations since then).

ozym:
@elidana the citation column is intended to have a lookup to a reference in the citations.csv file (under references) which will have a DOI (if available).

elidana:
gotcha, thanks Mark!

elidana (approved):

PR #1448, opened 2022-09-19
add doi decoding code

ozym:
This PR adds a small function which can validate DOI entries, it is currently intended for the references tables, but could also be used if DOI values are added into the main tables at some point.

### DIFF
--- a/meta/doi.go
+++ b/meta/doi.go
@@ -1,0 +1,57 @@
+package meta
+
+import (
+	"bytes"
+	"errors"
+	"regexp"
+)
+
+const doiMatch = `^((?:https?://)?(?:[^/\s]+\.)*doi\.org/10|10|DOI:10)\.([^\s]+)+$`
+
+var doiRe = regexp.MustCompile(doiMatch)
+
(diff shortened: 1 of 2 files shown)


salichon (approved):
even add with GeoNet DOI  in the test ha.  :) that is 
https://doi.org/10.21420/0S8P-TZ38

salichon (commented):
I  am curious about the "Good" test DOI formats besides the one i got .. 
NB to em: .. Not everything is seismic ...

elidana:
late to the party here, apologies @ozym . That looks really good, I wonder what tables in delta you were thinking to use for this (if already existing ones, or a new one just for DOIs?).

ozym:
@elidana the only two that sprang to mind were the citations (see other PR) and perhaps network.csv

elidana:
cool, that's where my head was heading towards as well. We might have some complications as for a bunch of data we have created a "dataset" doi, but the network.csv is a good start for sure
grazie grazie :pray: :pray: :pray:

PR #1459, opened 2022-09-29
rename calibration component to number

ozym:
In line with other changes to channels and components, this PR updates the `calibration` table to use `Number` rather than `component`. The main reason for this is that a calibration can be added for a datalogger as well as a sensor, in which case a datalogger component has no real meaning.


salichon (commented):
This is making sens to me thanks 
will need to get used to  the definition - all good :)

mnaguit (approved):

mnaguit:
good call @ozym . Thanks!

elidana (approved):
looks good, thanks Mark !!!

PR #1460, opened 2022-09-29
add component and channel docs

ozym:
This just updates the install README to add details about component and channel.

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -19,6 +19,8 @@ Meta information for the GeoNet equipment network.
 * `connections.csv` - Datalogger and sensor connection details
 * `streams.csv` - Datalogger and recorder sampling configurations
 * `gains.csv` - site specific settings applied to individual datalogger and sensor that may impact overall sensitivities
+* `components.csv` - Individual sensor elements including measurement position and responses.
+* `channels.csv` - Individual datalogger recording elements including digitiser position, sampling rate, and responses.
 * `calibrations.csv` - Individual sensor sensitivity values that can be used rather than default values.
 * `components.csv` - Individual sensor elements including measurement position and responses.
 * `channels.csv` - Individual datalogger recording elements including digitiser position, sampling rate, and responses.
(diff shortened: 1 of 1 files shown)


salichon (approved):
Great - good to me  in practice for a change/additon 
- the _sensor component  "number"_ has to be explicit  (always  0,1,2 .... )
- the _datalogger channels "number"_ can be implicit (default to 0) 
and its value should be m,most of the time   0 or 3

install/README.md
```diff
@@ -19,6 +19,8 @@ Meta information for the GeoNet equipment network.
 * `connections.csv` - Datalogger and sensor connection details
 * `streams.csv` - Datalogger and recorder sampling configurations
 * `gains.csv` - site specific settings applied to individual datalogger and sensor that may impact overall sensitivities
+* `components.csv` - Individual sensor elements including measurement position and responses.
+* `channels.csv` - Individual datalogger recording elements including digitiser position, sampling rate, and responses.
```
elidana, line 23: minor change in the README, this is a duplication of line n.26

install/README.md
```diff
@@ -245,6 +247,56 @@ For the scale factor and bias either a value can be given directly or an express
...
+| _Azimuth_   | Internal azimuth of the compnent relative to whole sensor
+| _Types_     | A shorthand reference to the SEED type labels
+| _Response_  | A reference to the nominal StationXML response
+
+Example:
+
+    Make,Model,Type,Number,Subsource,Dip,Azimuth,Types,Response
+    Guralp,Fortis,Accelerometer,0,Z,-90,0,G,sensor_guralp_fortis_response
+    Guralp,Fortis,Accelerometer,1,N,0,0,G,sensor_guralp_fortis_response
+    Guralp,Fortis,Accelerometer,2,E,0,90,G,sensor_guralp_fortis_response
```
elidana, line 276: this section to describe `components` is already present in the paragraph below (after "CALIBRATIONS"), shall one of the two be removed from the README?

install/README.md
```diff
@@ -245,6 +247,56 @@ For the scale factor and bias either a value can be given directly or an express
...
+| _Number_        | Datalogger channel offset, an empty value will map to zero
+| _Sampling Rate_ | Configured Channel sampling rate
+| _Response_      | A reference to the nominal StationXML response
+
+Example:
+
+    Make,Model,Type,Number,SamplingRate,Response
+    Nanometrics,Centaur CTR4-6S,Datalogger,,200,datalogger_nanometrics_centaur_200_response
+    Quanterra,Q330HR/6,Datalogger,0,200,datalogger_quanterra_q330_highgain_200_response
+    Quanterra,Q330HR/6,Datalogger,3,200,datalogger_quanterra_q330_200_response
```
elidana, line 298: same as above, is that section now a duplicate in the README?

elidana (changes requested):
Hi @ozym , this looks good to me, but I think a couple of minor changes are needed, as I think the info in the README seems to be now duplicated (see detailed comments below).

elidana:
uh not sure if there were new things, I'll have a quick look and raise a new pr if I notice something.

elidana:
the only difference this one was adding are the examples, but since all other descriptions in the README won't have one, I think we can leave it as is and we can always use the files in `meta/testdata` for reference, and until the channels and components tables are built.

PR #1463, opened 2022-10-03
Nanometrics Trillium Horizon TH120-2 response added

CBurton90:
This response is a replica of the original Nanometrics Trillium Horizon TH120-1. We are only adding it for naming clarity in delta so anyone out of the loop will realize the response is identical between both models (TH120-1 & TH120-2) and won't get into a panic thinking an unapproved sensor has been deployed from this point forwards.

If there is a better way to indicate and create awareness to people that the TH120-1 and TH120-2 are the same thing then I'm all ears and you can delete this PR.

More details as per linked ticket:

Model TH120-2
18437-02
ASM19106 rev. 10

### DIFF
--- a/resp/auto.go
+++ b/resp/auto.go
@@ -699,6 +699,14 @@ var SensorModels map[string]SensorModel = map[string]SensorModel{
 		Vendor:       "",
 		Components:   []SensorComponent{{Azimuth: 0, Dip: -90}, {Azimuth: 0, Dip: 0}, {Azimuth: 90, Dip: 0}},
 	},
+	"Trillium Horizon TH120-2": {
+		Name:         "Trillium Horizon TH120-2",
+		Type:         "Broadband Seismometer",
+		Description:  "Trillium Horizon 120",
+		Manufacturer: "Nanometrics Inc.",
+		Vendor:       "",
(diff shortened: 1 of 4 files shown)


salichon (commented):
thanks @CBurton90 
this is looking ok -
Would you put some version number ins the "description"
and what is the main reason of the release of a Version 2 (from the  maker side) ? 
thanks

ozym:
@CBurton90 this works, but it could be a bit more like how we handle multiple sensors that look the same, e.g. CMG-ESP and CMG-ESP3C

Either leave the name as "...TH120-1" and add the "-2" sensor in the list of sensors.

Or change the overall name to "...TH120" and make sure both sensors are in the sensors list.

CBurton90:
> @CBurton90 this works, but it could be a bit more like how we handle multiple sensors that look the same, e.g. CMG-ESP and CMG-ESP3C
> 
> Either leave the name as "...TH120-1" and add the "-2" sensor in the list of sensors.
> 
> Or change the overall name to "...TH120" and make sure both sensors are in the sensors list.

@ozym Changes made

ozym (approved):
nice

salichon (approved):
Picked up the change and happy 
to see offcially the XML snippets appearing :)

PR #1465, opened 2022-10-06
Adding WMO and Bulletin Headers for DART and TG

meganmadley:
The readme might need some touching up and we may also want to look into WMO numbers for the TGs. Would also like approval from delta for this merge.


salichon:
oopsy @meganmadley  I reckon the table change need the tests to be adapted or something like this.

ozym:
It needs talking about.

meganmadley:
Yeah I figured it would, no problems leaving it until delta next week?

meganmadley:
So this is going to be delayed by another fortnight due to a special session of delta today, shall I schedule a meeting for a few of us to chat about this? Or does it need a full delta quorum and so it better to leave for a few more weeks?

ozym:
It's much easier if this goes into a new file, rather than inserting it into an existing one (hence all the checks failing).

Maybe build an example file and see how that goes, this is mainly for gts and wmo and perhaps even what's used in the CREX headers for location (as these are often different from the "real" locations).

meganmadley:
Deleting this and will be placing it in a new file

PR #1466, opened 2022-10-09
add sampling rate to component

ozym:
Turns out there's an edge case which means that some responses fall in the middle where they don't need the response of the datalogger or sensor, but do need a derived response. The part that indicates this is that a sampling rate is needed for the sensor response to allow it to be matched to a stream.

This is mainly used for things like gain changes (i.e. height conversions etc.).


meta/component.go
```diff
@@ -130,20 +137,34 @@ func (s *ComponentList) decode(data [][]string) error {
 			azimuth = v
 		}
 
+		var samplingRate float64
+		if s := strings.TrimSpace(d[componentSamplingRate]); s != "" {
+			v, err := strconv.ParseFloat(d[componentSamplingRate], 64)
```
junghao, line 142: Maybe `strconv.ParseFloat(s, 64)`?

junghao (changes requested):
A small change.

meta/component.go
ozym: thanks @junghao , good catch.

junghao (approved):
Looks good

PR #1468, opened 2022-10-09
update connection numbers

ozym:
This is an incremental step for response generation,

This change starts to use the `connections` _Number_ field to indicate where the sensor is connected into the datalogger.

For responses this is needed as some dataloggers (Q330HR) have different characteristics depending on which digitiser bank (set of numbers) the sensor is plugged into.

It also allows checking that multiple sensors can't be plugged into the digitiser at the same time (indicating possible deployment date problems).

This will not impact any existing systems.

However, this has highlighted a couple of issues which may need addressing.

- Some sites had two streams an analogue stream (to a remote digitiser) and a local stream to the replacement digitser (coming from the same sensor).  This was handled in the CUSP system by using a different location code, the sensor was the same but the digitiser wasn't and mainly impacts the original telemetered sites.   It can be fixed for some sites where the analogue digitiser has been deployed, but other sites (e.g. in Auckland) this hasn't been so there appears to be an overlap. (In reality this may not be an issue and we should just remove this overlap as it was only an odd construct to solve a transitional problem).
-  The ordering of the `connections` file is a little bit awkward for review, it should group roles together.


install/connections.csv
```diff
@@ -25,17 +25,17 @@ AN08,10,"Mount Fleming, Antarctica",,,1999-11-09T00:00:02Z,2000-01-23T00:00:00Z
...
 ARCZ,10,Arundel,,,2014-05-13T00:05:00Z,9999-01-01T00:00:00Z
 ARHZ,10,Aropaoanui,,,2010-03-11T21:16:00Z,9999-01-01T00:00:00Z
 ARKE,10,Arakihi,Short Period,,1994-07-22T05:17:00Z,1994-12-12T23:16:00Z
 ARPA,10,Arthurs Pass,,,1995-11-11T00:00:01Z,1996-04-07T23:59:59Z
 ARPC,10,Lake Arapuni Road,Short Period,,2001-01-19T23:54:00Z,2001-07-04T07:21:00Z
 ASHV,10,Ash Pit Road,Short Period,,1995-01-11T00:56:00Z,1995-05-28T18:20:00Z
-AUCT,40,Devonport Naval Base,,,2009-03-26T00:00:00Z,9999-01-01T00:00:00Z
-AUCT,41,Devonport Naval Base,,,2009-03-26T00:00:00Z,9999-01-01T00:00:00Z
+AUCT,40,Devonport Naval Base,,0,2009-03-26T00:00:00Z,9999-01-01T00:00:00Z
+AUCT,41,Devonport Naval Base,,1,2009-03-26T00:00:00Z,9999-01-01T00:00:00Z
```
elidana, line 38: I wonder if we should add somewhere in the connections readme that "numbers" are generally
short period and broadband: group of 3 [(0,1,2)] 
strong motion: group of 3 (3-4-5)
tsunami: group of 1 (0 for primary, 1 for backup)

install/connections.csv
```diff
@@ -86,48 +86,48 @@ CARD,10,Cairn Hill,Portable,,2010-09-04T23:23:02Z,2010-09-29T00:00:00Z
 CASA,10,Glenthorne,Short Period,,1994-06-19T20:22:00Z,1994-08-06T11:15:00Z
 CASM,10,Castle River,Short Period,,1993-10-18T05:51:00Z,1994-03-11T20:39:00Z
 CAW,10,Cannon Point,,,2003-05-20T02:00:00Z,9999-01-01T00:00:00Z
-CAW,11,Cannon Point,,,2003-05-20T02:00:00Z,2003-05-20T03:00:00Z
+CAW,11,Victoria University,EARSS Network,,2003-05-20T02:00:00Z,2003-05-20T03:00:00Z
```
elidana, line 89: Shall this be 
`CAW,11,Cannon Point,EARSS Network,,2003-05-20T02:00:00Z,2003-05-20T03:00:00Z`
instead?

also, I think the line above (CAW location code 10) needs also to be changed 
either to 
`CAW,10,Victoria University,EARSS Network,,2003-05-20T02:00:00Z,9999-01-01T00:00:00Z`
or to 
`CAW,10,Cannon Point,EARSS Network,,2003-05-20T02:00:00Z,9999-01-01T00:00:00Z`

install/connections.csv
```diff
@@ -86,48 +86,48 @@ CARD,10,Cairn Hill,Portable,,2010-09-04T23:23:02Z,2010-09-29T00:00:00Z
...
 CCKO,10,Cattle Creek,Portable,,2014-10-11T02:01:00Z,2015-04-11T21:00:00Z
 CCW,10,Cape Campbell,,,2003-06-15T04:50:00Z,2003-06-19T13:30:00Z
-CCW,11,Cape Campbell,,,2003-06-15T04:50:00Z,2004-04-13T20:00:00Z
+CCW,11,Victoria University,EARSS Network,,2003-06-15T04:50:00Z,2004-04-13T20:00:00Z
```
elidana, line 92: similar to `CAW` example, should these entries be something like
`CCW,10,Cape Campbell,EARSS Network,,2003-06-15T04:50:00Z,2003-06-19T13:30:00Z`
`CCW,11,Cape Campbell,EARSS Network,,2003-06-15T04:50:00Z,2004-04-13T20:00:00Z`

?

install/connections.csv
```diff
@@ -86,48 +86,48 @@ CARD,10,Cairn Hill,Portable,,2010-09-04T23:23:02Z,2010-09-29T00:00:00Z
...
+CPIT,40,Castlepoint,,0,2009-09-16T00:00:00Z,9999-01-01T00:00:00Z
+CPIT,41,Castlepoint,,1,2009-09-16T00:00:00Z,9999-01-01T00:00:00Z
 CPWZ,10,Castlepoint Station,,,2009-07-07T01:00:00Z,9999-01-01T00:00:00Z
 CREC,10,Crestridge,Short Period,,2001-01-09T11:09:00Z,2001-06-24T11:39:00Z
-CRGZ,10,Rukuhanga Station,,,2016-09-03T23:25:01Z,9999-01-01T00:00:00Z
-CRGZ,20,Rukuhanga Station,,,2016-09-03T23:30:00Z,9999-01-01T00:00:00Z
-CRLZ,10,Canterbury Ring Laser,,,2003-03-12T00:00:01Z,2014-01-09T00:00:00Z
+CRGZ,10,Rukuhanga Station,,0,2016-09-03T23:25:01Z,9999-01-01T00:00:00Z
+CRGZ,20,Rukuhanga Station,,3,2016-09-03T23:30:00Z,9999-01-01T00:00:00Z
+CRLZ,10,Canterbury Ring Laser,,0,2003-03-12T00:00:01Z,2014-01-09T00:00:00Z
```
elidana, line 126: shall this one have "Ring Laser" added as role as well, similarly to the 1996-2003 entry for CRLZ?

install/connections.csv
```diff
@@ -140,34 +140,34 @@ CSWB,10,White Bluffs,Portable,,2013-07-23T05:20:10Z,2013-11-19T10:00:00Z
...
 DAGG,10,Dagg Sound,Portable,,2009-07-17T02:08:00Z,2009-09-14T21:34:00Z
 DCE,10,Deep Cove,Short Period,,1989-06-03T02:49:00Z,1989-06-09T18:00:00Z
 DCVF,10,Duck Cove,Short Period,,1993-03-22T04:51:00Z,1993-06-19T07:27:00Z
-DCZ,10,Deep Cove,,,2005-07-23T00:00:00Z,9999-01-01T00:00:00Z
-DCZ,20,Deep Cove,,,2005-07-23T00:00:00Z,2015-10-22T00:00:00Z
+DCZ,10,Deep Cove,,0,2005-07-23T00:00:00Z,9999-01-01T00:00:00Z
+DCZ,20,Deep Cove,,3,2005-07-23T00:00:00Z,2015-10-22T00:00:00Z
 DCZ,20,Deep Cove,Strong Motion,,2015-10-22T05:00:01Z,2021-08-09T21:08:00Z
-DIW,11,D'Urville Island,,,2003-12-10T00:00:00Z,2004-06-08T01:00:00Z
+DIW,11,Victoria University,EARSS Network,,2003-12-10T00:00:00Z,2004-06-08T01:00:00Z
```
elidana, line 154: change to:
`DIW,11,D'Urville Island,EARSS Network,,2003-12-10T00:00:00Z,2004-06-08T01:00:00Z`

?

elidana (commented):
Hi Mark, before I continue the review with possibly useless comments:
are you using "Victoria University,EARSS Network" as a `Place` and `Role` combo to be used for the edge case of analogue stream and digitiser that overlaps during a transition period?

I have made some specific comments to update some of these entries where I've noticed that place name has been replaced with "Victoria University"

ozym:
@elidana it's complicated.  Basically for an analogue system the sensor is at the site, i.e. Cannon Point, but the digitiser is at the remote site e.g. Dominion Observatory (multiple signals will be coming into that site).

It looks like a lot of this pre-digital configuration hasn't migrated over to delta here (probably a follow up). But there are times when the same sensor (or possibly a replacement as some of them will have been updated at the same time) is being recorded locally as well as remotely.  As it is considered the same sensor it can't be recorded at two places with the same code, hence the odd 14's etc.

install/connections.csv
ozym: The digitser was at Victoria University and the sensor at Cannon Point.
(at the same time the local digitiser 10 was being used, this would have been a transition from using 10 -> 11, but keeping 10 for the local Q330).

install/connections.csv
ozym: Same problem, the analogue signal was digitised at Vic,, where as the digital signal locally.

The site would have again changes from 10 -> 11 (old kit), and new sensor (perhaps) renamed to 10 locally.

install/connections.csv
ozym: Same deal, for DIW there was never a digitiser on the Island, only a sensor and an FM radio.

ozym:
This is simply some old oddities that are coming back in to bite us.  In reality due to the FM telemetry we won't know what the actual response/calibration is (just an estimate based off what we record). So we need the information to say what was where but it's not needed for building responses per se).

It also causes some confusion when the locations 10, 11 are used where sometimes it's the same sensor just different recording sites (needed to generate different names etc.)  In most of these cases though the site that was 10 was renamed 11, and the new site retained the 10.   This isn't what we do now (for this very reason of reducing confusion as to what is where) - at the time 10 was considered "primary" and 11 "secondary" rather than physical locations.

elidana:
ooooh gotcha :exploding_head: ok, nevermind my comments then, I will update the review

ozym:
@elidana your comment about the channel offsets is valid.  There are assumptions made (e.g. you know the number of channels in a sensor, and you know that when it comes to a weak + strong site, the weak is plugged into the first slot. This may not be true in reality always but it will work for creating response files.

Likewise there's an assumption that 3 similar sensors will be ordered numerically, i.e. for microphones 31, 32, 33 will be in order.

ozym:
It also means we need to be careful when sites change, i.e. a 6 channel sensor is swapped for a 3 channel sensor. We need to make sure the extra channels are handled (either they go to a different instrument, or the connection gets updated to the new offset).

elidana:
the general assumption on channel offset can be added in the README file perhaps?

Also, back to the oddities, shall we add a "Victoria University" entry to the dataloggers.csv then?

ozym:
I have some tests in place, but they rely on other information (components and channels) to work out what's valid. This will need to be merged in as well.

ozym:
@elidana just a rebase to handle the recent changes in the repo.

elidana:
and nevermind about my comment about dataloggers, there's one already
`dataloggers.csv:GNS Science,EARSS/16,209,Victoria University,EARSS Network,1998-03-02T03:48:00Z,2004-06-08T01:00:00Z`

ozym:
There are some other "problems" that will need solving.

TAZ / TARZ
KAAZ
and one near Chateau but I've forgotten.

These don't have the analogue digitisers in dataloggers.csv which are needed to split out the problem sites.

ozym:
I also added the "0" even though it isn't needed but it's a style that indicates there's other sensors connected as well.

install/connections.csv
```diff
@@ -617,32 +617,32 @@ OT001,04,Waimangu Inferno Crater,,,2019-10-11T00:50:00Z,9999-01-01T00:00:00Z
...
-OTAT,40,Dunedin Tsunami,,,2010-02-25T00:00:00Z,9999-01-01T00:00:00Z
-OTAT,41,Dunedin Tsunami,,,2010-02-25T00:00:00Z,9999-01-01T00:00:00Z
+OTAT,40,Dunedin Tsunami,,0,2010-02-25T00:00:00Z,9999-01-01T00:00:00Z
+OTAT,41,Dunedin Tsunami,,1,2010-02-25T00:00:00Z,9999-01-01T00:00:00Z
 OTKO,10,Otekaieke,Portable,,2014-09-18T01:00:00Z,2015-04-11T05:00:00Z
-OTVZ,10,Oturere,,,2004-11-18T04:00:00Z,9999-01-01T00:00:00Z
-OTVZ,30,Oturere,,,2013-10-02T00:00:00Z,9999-01-01T00:00:00Z
-OUZ,10,Omahuta,,,2002-08-30T08:00:00Z,9999-01-01T00:00:00Z
+OTVZ,10,Oturere,,0,2004-11-18T04:00:00Z,9999-01-01T00:00:00Z
+OTVZ,30,Oturere,,3,2013-10-02T00:00:00Z,9999-01-01T00:00:00Z
```
elidana, line 624: how would we handle "numbers" for a site that have
- weak motion (0-2)
- strong motion (???)
- acoustic (3)

elidana (approved):
looks good.
I have not done checks against data and data in the archive. Happy to do so if it's worth (at this stage)

install/connections.csv
ozym: they all can't connect to a single digitser (at the moment we only have a maximum of 6 channel mdels). i.e. weak + strong, or weak + something else, and so far the "else" has been unique.

install/connections.csv
salichon: need context ..  should that be a seismometer (though with fancy  lights?:))

salichon (approved):
@ozym  
this is looking good, 
The connection numbers convention  suits.
This handles the one component sensor  such as the tide gauges (see CHIT)  and provides the "1" and "0" for these TG case.

mnaguit:
Apologies to be late in this party. 😄 
I think I missed the initial discussion on this (if there was one prior to the issuance of this PR). But in any case, is there any documentation available for this update in `connections.csv`? It would be very useful, so that I can catch up with these changes. 
Thanks @ozym @salichon @elidana

ozym:
@mnaguit only real docs  is https://github.com/GeoNet/delta/blob/main/install/README.md#connections

mnaguit:
Thank you @ozym 👍

PR #1469, opened 2022-10-09
wi227mc_test

rumachan:
Three files required to set up WI227 for manualcollection data for Tilde.

No assets or install done, and not required for this.


ozym (approved):
Worth a crack Nigel.

elidana:
too late to the party. Looks good, we might  want to also add "MC" entry to the `network/networks.csv` table

PR #1470, opened 2022-10-10
TLZ sensor swap

mnaguit:
- from Guralp CMG-3ESP SN T3P74 to Trillium Horizon TH120-2 SN 2745
- for LC 10 HH(ENZ)
- no firmware info for Horizon


CBurton90 (approved):

salichon (approved):
👍  okay  donkey

PR #1475, opened 2022-10-10
update bhw bswz streams

ozym:
Some small updates to the streams entries for BHW and BSWZ.

BHW - the site was originally a short period telemetered to kelburn, it was switched to a locally recorded site in 2003. It remained running an analogue connection for a short while after (the location 14 is used here to indicate it was recorded at kelburn with location code 14 (this doesn't represent a new site). 
This has been removed as it is causing issues with generating the correct streams for the site, we may need to review this at a later date to handle this, and other, odd cases related to switching from analogue to digital.

Also, the site was switch from a short-period to a broadband between 2016 and 2020 and then switched back. This is reflected in the 1 Hz stream but not in the 100 Hz stream (i.e. E -> H -> E is missing)

For BSWZ the end time should be consistent for B as well as H for the time it switched to a standalone recorder.


salichon (approved):
all good 
cheers

salichon:
data for the .14. location are not stored afaic

mnaguit (approved):

salichon:
@elidana @mnaguit 
as a further documentation/clarification  into "some" aspects of the data/metadata alignments
> This has been removed as it is causing issues with generating the correct streams for the site, we may need to review this at a later date to handle this, and other, odd cases related to switching from analogue to digital

i m tagging it into that board.

PR #1480, opened 2022-10-12
GVZ datalogger swap

mnaguit:
- for LC 10 with HH(ENZ) streams
- Centaur firmware version 4.5.20
- from Q330/3 SN2236 to Centaur SN9443


salichon (approved):
sweet as

PR #1483, opened 2022-10-12
Creating a new file to add WMO numbers and bulletin headers

meganmadley:
Creating a new file for WMO and bulletin headers. Possible need discussions on:
- title - any ideas? Maybe GTS?
- will then need updates in the README
- also this basically has the same content as the gauges.csv, some of the details could be removed to avoid repitition, but in theory this file shouldn't have to get changed often

Generated following on from @ozym comment here: https://github.com/GeoNet/delta/pull/1465#issuecomment-1276785280


salichon (approved):

salichon:
name `gts_technical.csv` ? 
 We ll also account for this file to  change in  any new deploy :)

ozym:
Not quite what I had in mind, maybe we need some brainstorming.

meganmadley:
Hey @ozym I've trimmed it down a bit and renamed it, what do you think? I reckon we could even get rid of the lat and longs, or maybe just have the ones from the first deployment, and also leave the end dates open-ended (unless the site gets closed) so we aren't tracking individual deploys here as well, what do you reckon? Making it similar to the network/stations file (would use the same lat and longs from there)

meganmadley:
Could remove the lat/long. Technical file title will need to change to gts.csv if we are happy.

ozym:
Here is some of the meta data that's needed for processing:

_coastal crex/gts:_

These can just be system settings (not needed in table but NZKL could be useful, note NZKL means "New Zealand Kelburn")

> designator SZNZ01
> originator NZKL
> file prefix T_SZNZ01_C_NZKL_

crex header AUCT/40/-3683144/17478654 etc. (these may differ)

_coastal crex/linz:_

gauge: AUCT/40/-3683143/17478653 (these may differ)

**buoy:** 

buoy: NZA/NZ41 
dashboard: NZA/-42.3717/176.9093/2635 (can be gathered from station/site)

environment/README.md
```diff
@@ -96,6 +98,19 @@ Quality estimates are roughly
 | _Start Date_ | General date and time at which the _Gauge_ description was valid.
 | _End Date_ | General date and time at which the _Gauge_ description was no longer valid.
 
+#### _GTS_ ####
+
+| Field | Description |
+| --- | --- |
+| _Gauge_ | Code used to uniquely identify Tide Gauge _Stations_
```
salichon, line 105: few questions about "names"
Does the gts file encompass Gauges and BPRs and is specific to dart sytems /BPR?  
Shall we make the difference between gauge and BPR ? 
and 
Should we  rename "tide gauge" tsunami gauge" ?

meganmadley: The GTS and WMO numbers are just using a station so they aren't tied to a BPR and won't be updated every deployment 

We probably should move to Tsunami Gauge but as the tsunami gauges are called tide gauges throughout delta, it may need changing in a few places.

meganmadley:
Closing this PR as going with an alternative option suggested by @ozym (PR https://github.com/GeoNet/delta/pull/1647/files)

PR #1484, opened 2022-10-12
update streams after stationxml updates

ozym:
This is the first set of updates related to reworking the stationxml code (there are fewer assumptions made now which exposes some hidden issues).

- In some cases a site was switched from short period to broadband, the LH?, HH? entries may have been updated, but the EH? values still remained
- Sometimes a sensor is swapped between short period and broandband and back again. This needs to be reflected in the streams files as otherwise it may generate streams where it shouldn't

For QRZ I think there's a double negative going on. The sensor/datalogger pairing indicates it should be reversed without the stream also indicating it is reversed (this will all be handled in the future polarities file). But I haven't updated this here (essentially going from true to false).

The only updates in stationxml I'm expecting are adjustments to the WAKC end times:

```diff
6c6
<   <Created>2022-10-13T11:43:04</Created>
---
>   <Created>2022-10-13T12:02:17</Created>
3324117c3324117
<       <Channel code="BNZ" startDate="2016-06-14T00:55:00" endDate="2020-06-16T00:00:00" restrictedStatus="open" locationCode="20">
---
>       <Channel code="BNZ" startDate="2016-06-14T00:55:00" endDate="2020-07-20T22:06:00" restrictedStatus="open" locationCode="20">
3324550c3324550
<       <Channel code="BN1" startDate="2016-06-14T00:55:00" endDate="2020-06-16T00:00:00" restrictedStatus="open" locationCode="20">
---
>       <Channel code="BN1" startDate="2016-06-14T00:55:00" endDate="2020-07-20T22:06:00" restrictedStatus="open" locationCode="20">
3324983c3324983
<       <Channel code="BN2" startDate="2016-06-14T00:55:00" endDate="2020-06-16T00:00:00" restrictedStatus="open" locationCode="20">
---
>       <Channel code="BN2" startDate="2016-06-14T00:55:00" endDate="2020-07-20T22:06:00" restrictedStatus="open" locationCode="20">
```


salichon:
thanks @ozym  this is great 
For that matter my review approach would be to check against the data sets  for the periods considered
although some (like the temporary  `HD??`) might not be systematically all available. 

If the data exist then the metadata should reflect this.
It s a bit archive driven maybe /...

ozym:
@salichon the original problem is that the streams files haven't been maintained against the archive, they've been updated mainly as new things come in (and are checked). This is a more about housekeeping as it doesn't change the actual stationxml outputs.

elidana:
thanks for this @ozym ! are the time gaps included here reflecting what is provided by the current stationxml output?

ozym:
@elidana yes, there's still a couple of oddities (doesn't impact current system, just the updated one).

salichon (commented):
Hi 
I v been through most streams changes except the HD?? series (tmp not meant there) 
and been Browsing  metadata Vs data:
  All good with only apparently a few to investigate (perhaps) 

- KHEZ  has LH ( and no HH) in 2011 (most likely to be removed - or error in config for BB/shortperiod at that time )  - LH might be not releavant or a BB was there already  

- WHZ is looking like having BN data for the period 2002 -2012 as in the previous delta stream file:
`WHZ,20,50,false,false,false,2002-03-02T00:00:00Z,2014-06-05T05:00:00Z`

- WIZ has short period  data 2005/2006 SH 

Thanks - 
With regards to these and other if ever we ll patch that file when relevant and agree upon

ozym:
@salichon what does this mean? Is there a problem in the PR that needs fixing, or is there a problem in the original data that needs fixing? Otherwise this PR will wither on the vine.

install/streams.csv
```diff
@@ -1316,8 +1330,8 @@ MWFS,20,B,N,50,true,false,false,2014-11-03T00:01:00Z,2020-06-16T00:00:00Z
 MWFS,20,H,N,200,true,false,true,2006-07-21T01:00:00Z,9999-01-01T00:00:00Z
 MWZ,10,L,H,1,false,false,false,2004-02-29T00:00:00Z,9999-01-01T00:00:00Z
 MWZ,10,H,H,100,false,false,false,2004-02-29T00:00:00Z,9999-01-01T00:00:00Z
-MWZ,20,B,N,50,false,false,false,2004-02-29T00:00:00Z,2011-08-18T20:50:00Z
-MWZ,20,H,N,200,false,false,true,2004-02-29T00:00:00Z,2011-08-18T20:50:00Z
+MWZ,20,B,N,50,false,false,false,2004-02-29T00:00:00Z,2011-08-18T12:00:00Z
+MWZ,20,H,N,200,false,false,true,2004-02-29T00:00:00Z,2011-08-18T12:00:00Z
```
elidana, line 1334: shall the end timestamp for MWZ be also changed in `sensors.csv` ?
currently this is
`Kinemetrics Inc.,FBA-ES-T-ISO,3498,MWZ,20,0,,0,0,0,0,0,0,2011-03-23T02:00:00Z,2011-08-18T20:50:00Z`

elidana (approved):
Looks good to me!

I did a inexperienced review and checked this pr by comparing these changes in `streams.csv` to:
- response to fdsn channel query
- current metadata for sensors and dataloggers in `install/sensors.csv` and `install/dataloggers.csv`

a minor comment (but there might be other things that I haven't noticed like that) to update related files and use same timesteamps.

(Like the addition of some pre-2000 metadata info!)

install/streams.csv
ozym: Yes, possibly. The sensor time could be taken from when it was physically removed whereas the stream could be from when it stopped recording data. It sort of depends on a policy with which way to go (a sensor that doesn't record anything doesn't need to know when it was removed, that is we're not tracking assets anymore).

install/streams.csv
```diff
@@ -2387,8 +2407,9 @@ WHSB,2F,,,50,true,false,false,2009-04-15T00:00:04Z,2019-12-16T00:00:00Z
 WHSB,2F,,,200,true,false,false,2009-04-15T00:00:04Z,9999-01-01T00:00:00Z
 WHSB,2G,,,50,true,false,false,2009-04-15T00:00:04Z,2019-12-16T00:00:00Z
 WHSB,2G,,,200,true,false,false,2009-04-15T00:00:04Z,9999-01-01T00:00:00Z
-WHSZ,10,L,H,1,false,false,false,2020-06-19T00:00:00Z,9999-01-01T00:00:00Z
-WHSZ,10,E,H,100,false,false,false,2020-06-19T00:00:00Z,9999-01-01T00:00:00Z
+WHSZ,10,L,H,1,false,false,false,2021-11-03T22:57:00Z,9999-01-01T00:00:00Z
+WHSZ,10,E,H,100,false,false,false,2020-06-19T00:00:00Z,2021-11-03T02:58:00Z
+WHSZ,10,H,H,100,false,false,false,2021-11-03T22:57:00Z,9999-01-01T00:00:00Z
```
mnaguit, line 2412: This looks fine, with the history of sensor swaps:
(also just a note for me so I don't need to remind myself later why I have this comment 😄 )
- EH*  at LC 10 starts in 2020 with Lennartz and closes in 2021
- HH* at LC 10 starts in 2021 & still open, with Trillium 

However, I was confused why there are two location codes (10 and 11) at the same station coordinates with both broadband streams. The difference is LC 10 has non-axial components, whereas LC 11 has axial components. Does that sound confusing or I'm just confusing myself? 😄

install/streams.csv
```diff
@@ -2429,7 +2449,8 @@ WIGC,20,B,N,50,true,false,false,2010-03-25T05:00:00Z,2020-06-16T00:00:00Z
 WIGC,20,H,N,200,true,false,true,2004-11-27T00:00:00Z,9999-01-01T00:00:00Z
 WILA,10,S,H,50,false,false,true,1994-06-21T21:27:00Z,1994-06-28T23:34:00Z
 WIZ,10,L,H,1,false,false,false,2007-04-19T04:00:00Z,9999-01-01T00:00:00Z
-WIZ,10,E,H,100,false,false,false,2006-09-19T00:00:01Z,9999-01-01T00:00:00Z
+WIZ,10,E,H,100,false,false,false,2006-09-19T00:00:01Z,2007-04-19T03:00:00Z
+WIZ,10,H,H,100,false,false,false,2007-04-19T04:00:00Z,9999-01-01T00:00:00Z
 WIZ,30,L,D,1,false,false,false,2006-09-19T00:00:01Z,9999-01-01T00:00:00Z
```
mnaguit, line 2454: This looks fine, but some comments on other existing metadata
- `sites.csv` & `sensors.csv` are aligned, noting that WIZ at LC 10 was opened in 1988.
- However, metadata at `streams.csv` only started in 2006.

install/streams.csv
```diff
@@ -2402,7 +2423,6 @@ WHZ,10,L,H,1,false,false,false,2001-04-07T21:53:00Z,9999-01-01T00:00:00Z
 WHZ,10,S,H,50,false,false,true,1993-03-22T15:45:00Z,2001-04-05T14:34:00Z
 WHZ,10,H,H,100,false,false,false,2001-04-07T21:53:00Z,9999-01-01T00:00:00Z
 WHZ,11,S,H,50,false,false,true,2001-04-08T10:20:00Z,2001-10-08T17:24:00Z
-WHZ,20,,,50,false,false,false,2002-03-02T00:00:00Z,2014-06-05T05:00:00Z
```
mnaguit, line 2405: From random checking, there are available data for BN(ENZ) streams
WHZ_BN

mnaguit (commented):
I hope my comments would not open a can of worms 🐍

install/streams.csv
ozym: This is Whataroa River borehole, one sensor was at the surface and one down the hole, the one down the hole (11) would have been axial.

install/streams.csv
ozym: This will be related to the transition from analogue to digital, there's probably work to do here looking at historical connections and any early data (most likely triggered) which will not have good response information.

salichon:
Good morning @ozym 
this means  that the PR is compliant with the test though some streams are not compliant with the data as browsed .
the classic metadata/data alignement ... :( 
It does not inpact the validity of the PR  - just a few corrective patches. 

 - KHEZ will need a check  to understand the actual sensor set up  of 2011 
 - WHZ  an inclusion of the BN streams for  a period before 2012
 - WIZ an inclusion of the SH  streams for 2005/6 

upon verification
this  does not have to be a blocker for the owrk in ptorgress just flags for correction

salichon:
triage/patching to be followed into : https://github.com/GeoNet/tickets/issues/12107

PR #1485, opened 2022-10-13
update connections

ozym:
A few small changes to the datalogger installation and connections after stationxml update.

Main one is the MKAZ, having the installation of the ORION marked as portable.

The NRZ site is mainly a telemetered site rather than a locally digitied site (this still needs work but doesn't impact current response as it doesn't have one).

No changes to stationxml are expected:

```diff
6c6
<   <Created>2022-10-13T11:43:04</Created>
---
>   <Created>2022-10-13T13:27:02</Created>
```


mnaguit (approved):

salichon (approved):
here a good illustration of the number swap for the datalogger over time 
3 for a strong RPZ 20 and then 3 for a weak RPZ 11
Same datalogger used as for RPZ 10 (number 0) ?

PR #1489, opened 2022-10-15
add preamp table

ozym:
As part of the revamp of stationxml configuration the process of handling pre-amplification of datalogger signals needs to be handled.

For reference: http://docs.fdsn.org/projects/stationxml/en/latest/response.html#stage-2-the-preamplifier

Currently this is handled via a specific combination of sensors and dataloggers (that is, it is assumed that the datalogger will be configured in a consistent way when attached to a particular sensor with low voltage input, e.g. borehole, or microphones).

This change makes this more explicit, it builds a table of times when the pre-amp gains are applied:

Currently the Q330s only have a single pre-amp stage (either 30 or 32 depending on the model), where appropriate we have transferred this to the Centaur where a factor of 40 has been chosen.   This model has more options and this table allows more fine-grained control.

This table is very similar to the gain table, the only change is the absence of a bias.  The difference in practice is that the gain table will tend to reflect changes in the sensor, whereas the preamp is aimed at changes in the dataloggers. This will hopefully keep the tables cleaner and easier to read.


mnaguit (commented):
Hi @ozym . This looks neat, and as you've mentioned more explicit with `gains.csv` specifically on sensors while `preamps.csv` on dataloggers. After this change is implemented, a quick discussion on next steps to handle these enhancements as part of delta cycle would help for clarity purposes.

A very minor request for change: shall we update the `preamps.csv` header from `Gain` to `Scale Factor`?

ozym:
@mnaguit thanks for catching that, "Scale Factor" should be updated, in the file header. I think I have done this already but not in this PR.  I did have Gain (as in gain setting) but that sort of clashed with the "gains" file.

mnaguit (approved):

salichon:
hi  there thanks this is making sense 
has to be  stired into the  process of equipement change for a good understanding and correct update.  :) 
as well as compared/complemented to the gains.csv change. 
Nothing major but more knowledfge about the diverse  equipement settings to get. 
Which type of instrumentation is it affecting? : something like ~~All but not the GNSS ones
Cheers!

salichon (dismissed):

salichon:
@moorematthew  this kind of instrumentation discussion :)

rumachan (dismissed):

salichon (approved):

salichon (approved):

PR #1500, opened 2022-10-21
WVZ sensor swap

mnaguit:
- from CMG-3ESP SN T3P70 to Trillium Horizon TH120-1 SN 817
- for LC 10 HH(ENZ)
- no firmware update


salichon (approved):
okdok

PR #1501, opened 2022-10-21
heights positive

rumachan:
Delta install/README.md says for cameras.csv 

Height | Installed camera vertical offset | metres positive upwards
-- | -- | --

But we had a mixture of positive and negative values, negative implying below the reference elevation which is normally the ground surface so below ground! Negative values come from a misunderstanding that the height in this file is the same as depth for seismic sensors, but it isn't.

All negative values replaced by the equivalent positive value.


rumachan:
I believe the reference is always ground surface

staylorofford (dismissed):

ozym (approved):

PR #1502, opened 2022-10-23
update meta tests

ozym:
This is a general tidy up of the meta module testing mechanism, but so far only related to the recent updates associated with the stationxml revamp.

The main problem was the conflicts being generated in the `list_test.go` as new code is added or removed.  The changes simply add a function to replicate the existing tests (written prior to the t.Run mechanism) into a function which can be called on any given List.

This reduces changes to the main `list_test.go` file and separates each new table struct into individual files which should reduce the conflict potential.

There is no actual code "meta" code change here, just a reworking of the existing tests.

### DIFF
--- a/meta/calibrations_test.go
+++ b/meta/calibrations_test.go
@@ -1,0 +1,33 @@
+package meta
+
+import (
+	"testing"
+	"time"
+)
+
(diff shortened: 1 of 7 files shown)


CallumNZ (approved):
Looks good.

PR #1504, opened 2022-10-25
update polarities file

ozym:
This PR pulls in the combined polarity information based on equipment models (which is currently handled in the response files). Essentially it is assumed that sensor X connected to datalogger Y will be reversed (mostly due to cabling, or convention).

These are not different to the current stationxml derived dips/azimuths. Eventually a test should be developed to help maintain this.


salichon (approved):
Might require some tests and some complementary  inclusions  later on 
(As the JRistau BB polarities inversion picked up for instance ) 
All good to me so far . - great- 
Goes with the documentation:  https://github.com/GeoNet/delta/tree/eafc6bf7e3f7d6a1c34d4ed9a4481adb6e705a8e/install#polarities
needs some practice.

mnaguit (approved):
All good with me as well.
Moving forward, as @salichon mentioned, I also need to familiarize myself with these new metadata (plus other latest additions). The [documentation](https://github.com/mnaguit/delta/tree/main/install) helps a lot.

PR #1507, opened 2022-10-27
wet chem sites, stations

rumachan:
network/stations.csv and network/sites.csv for wet chemistry sample locations. Follows the WI227 test example.

Don't have time for a few days to complete environment/features.csv so will do that separately. Having this merged will make it easier to have a conversation with VMG about what sample locations can be closed, which can then be included.


ozym:
looks good. I checked via the geojson code https://github.com/GeoNet/delta/blob/geojson/manual.geojson looks as expected.

ozym (approved):

PR #1514, opened 2022-11-02
DART Constituents

meganmadley:
Adding the tidal annual and monthly constituents from voyage 5 (July-August 2022). Further details and files in this ticket https://github.com/GeoNet/tickets/issues/12188


ozym:
don't merge any of this in until it's tested

ozym:
I think we need to look at this again, we have short term values overriding old data which may be better.  The original case (i.e. coastal data) never moved so the constituents only ever got better with time.  With dart the opposite happens, we get less accurate values overriding better values.  We may need to think about times or location codes.

meganmadley:
Yes we will have short term data going over the longer term stuff, this will be for the most recent values (B, F, G, H, I) as we get the previous years values and then a month which is the month with the new deployment. We could decide just to use the annual values? And then for those just deployed they would be using the annual values from the last deployment (FYI @JonoHanson)

salichon:
yeap - it may look a bit confusing to update with new data :) .
What would be the practical way  to log the best constituents per deploy over time?  

It s also pretty hard to review. 
There might be some easier ways to do so. 

I would be keen on targeting annual series only for instance.

PR #1517, opened 2022-11-03
add preamp code

ozym:
This is the follow up code to go with the newly merged preamp data.

It essentially handles the csv file, encode/decode and testing.

It is also added to the Set construct to allow for embedded  access.


meta/preamp.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+}
+
+func (g *PreampList) decode(data [][]string) error {
+	var preamps []Preamp
+	if !(len(data) > 1) {
+		return nil
+	}
+
+	for _, d := range data[1:] {
```
CallumNZ, line 141: Why is first ignored? Maybe add a comment

CallumNZ (commented):
Looks good, just a small bit that wasn't clear to me.

meta/preamp.go
ozym: yes, it's the comment line in the file. will add a comment about the comment.

CallumNZ (approved):

PR #1521, opened 2022-11-06
add an absolute bias to the gain table

ozym:
This is a response to another edge case.

In this instance the gain bias needs to be applied to the scaled gain factor and then have an absolute bias added.

The general case is given in the comments, but the short case is, given a sensor with a gain of `f`, then adding a scale factor of `a`, with a scale bias `b` and an absolute bias of `c` will result in the equation:

> value = volts * `a` * `f` + `f` * `b` + `c`

where the original equation is simply:

> value = volts * `f`

If the original sensor has a bias (`g`), then the initial equation will be

> values = volts * `f` + `g`

and will become

> value = volts * `a` * `f`  + `a` * `g` + `f` * `b` + `c`

after the gain has been applied.


elidana (approved):
looks good, thanks Mark!

ozym:
This will need some co-ordination once merged.

ozym:
This was tested via

```
go get  github.com/GeoNet/delta@absolute-bias
go mod tidy
go mod vendor
```

then test against the actual `absolute-bias branch of delta

There results are:

```
 diff origin.main origin.abs
20825a20826
>     <Absolute>0</Absolute>
20835a20837
>     <Absolute>0</Absolute>
20845a20848
>     <Absolute>0</Absolute>
20855a20859
>     <Absolute>0</Absolute>
```

elidana (approved):

PR #1522, opened 2022-11-06
whiteis soilgas sites

rumachan:
stations.csv, sites.csv, and features.csv for White Island "soil gas" sample sites. These collect CO2 flux rates and temperatures.

Have made an assumption that these locations are "closed". If it is decided to recommence these observations (and that is not clear) then new locations will likely be used. If old locations can be found, then they can be opened on an "as required" basis.


ozym (approved):
passes the checks!

PR #1528, opened 2022-11-10
Update constituents.csv

meganmadley:
Trialing the idea we discussed yesterday for only NZA. 
- Keeping all of the Z0 values 
- Using the rest of the constituents with the batch of constituents that has the longest time period


meganmadley:
Bump @salichon and @ozym are we still happy with this idea?

meganmadley:
Bump @salichon and @ozym, we should probably look into this again some time soon. I think it is failing travis because something needs to be updated in the travis checks (but I could be wrong)

meganmadley:
Bump @salichon and @ozym we should probably try and get this sorted at some point. Or at least decide if we just want to put the values in from the last deploy as they are, and then do more of a change for the next deploy.

PR #1535, opened 2022-11-17
add fumarole temperature locations

rumachan:
Added all locations for which we have fumarole temperature data for MC network in FITS. Most of these locations have unavailable fumarole chemistry data, but temperatures themselves are available and are used for monitoring purposes. In some cases they are the long-term equivalent of new observations using LRDCP.

Most stations/sites are left open, but one or two have been closed where the locality is known to no longer exist or only one temperature was measured many years ago.


ozym (approved):

elidana (approved):
Hi @rumachan ,

this looks good to me and happy for it to be merged.
I've been checking this using data currently available via FITS and looked at:
- start/end times 
- measurement types 
- station codes
i.e.: 
https://fits.geonet.org.nz/plot?type=scatter&sites=TO006,TO006A,TO006B&typeID=t
https://fits.geonet.org.nz/plot?type=scatter&sites=TO015&typeID=t

Haven't looked at coordinates but can double check if valuable
Assume that (when is not the same as in FITS) station-sublocation is:
- TO006,MC01 = TO006
- TO006,MC02 = TO006A
- TO006,MC03 = TO006B
-

rumachan:
Thanks. I used FITS data to construct start/end times. And your MC01, 02, etc is correct.

PR #1537, opened 2022-11-17
add collection code to set

ozym:
This code joins the dataloggers, the sensors, and the streams for the times when they overlap.  There is also provision for times which may need adjustments (i.e. gains, polarities) which may not relate to the physical connections between the various parts.

This is the engine for building stationxml like meta data.

### DIFF
--- a/meta/collection.go
+++ b/meta/collection.go
@@ -152,6 +152,49 @@ func (s *Set) Collections(site Site) []Collection {
 					continue
 				}
 
+				var gains []Gain
+				for _, g := range s.Gains() {
+					if g.Station != stream.Station {
+						continue
+					}
+					if g.Location != stream.Location {
+						continue
+					}
+					if g.Subsource != component.Subsource {
+						continue
+					}
+					if !span.Overlaps(g.Span) {
+						continue
+					}
+					gains = append(gains, g)
+				}
+				sort.Slice(gains, func(i, j int) bool {
+					return gains[i].Span.Start.Before(gains[j].Span.Start)
+				})
+
+				var sensors []Calibration
+				for _, c := range s.Calibrations() {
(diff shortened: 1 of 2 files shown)


meta/collection.go
```diff
@@ -0,0 +1,410 @@
...
+						}
+						if !span.Overlaps(c.Span) {
+							continue
+						}
+						dataloggers = append(dataloggers, c)
+					}
+					sort.Slice(dataloggers, func(i, j int) bool {
+						return dataloggers[i].Span.Start.Before(dataloggers[j].Span.Start)
+					})
+
```
bpeng, line 226: don't know if I fully understand the logic, but seems the `gains`, `sensors` and `dataloggers` are not dependent on `channels`, and can be calculated outside the channels loop L155

meta/collection.go
```diff
@@ -0,0 +1,410 @@
...
+									continue
+								}
+								if !span.Overlaps(c.Span) {
+									continue
+								}
+								sensors = append(sensors, c)
+							}
+							sort.Slice(sensors, func(i, j int) bool {
+								return sensors[i].Span.Start.Before(sensors[j].Span.Start)
+							})
```
bpeng, line 361: similar to the above, check if it is possible to calculate `gains` and `sensors` out side the channels loop

bpeng (changes requested):
Code looks tidy, little check

ozym:
@bpeng I did a recheck and found there should be an extra constraint for the sensor/datalogger calibrations for the Number. These files should have the Number explicitly given (as it's a calibration).

I've also adjusted the order of the loops to make the time span get evaluated as soon as possible, and to be consistent for both sets.

bpeng (approved):
OK

PR #1541, opened 2022-11-22
update resp generation

ozym:
This PR does a few things:

- It updates the automatic generation of response XML snippets from the current system to handle odd datalogger names.
- It adds a set of "combined" responses, these are for when there is no sensor+datalogger pair, but usually a digital sensor+datalogger combination.
- Switches to returning a byte slice for responses, rather than a decoded response. This avoids the dependency on the version of stationxml.

What needs reviewing is the two go files.


CallumNZ (approved):
Looks good

PR #1542, opened 2022-11-24
removed gain/calibration from Collection

ozym:
This PR catches up with the addition of the Correction application to Collection. The Correction code will discover the various options for ad hoc corrections, so that the Collection struct doesn't need to have these determined.

### DIFF
--- a/meta/collection.go
+++ b/meta/collection.go
@@ -16,11 +16,6 @@ type Collection struct {
 
 	InstalledSensor
 	DeployedDatalogger
-
-	Gains []Gain
-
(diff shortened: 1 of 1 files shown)


ozym:
@bpeng sorry I left some code in the previous PR that I'd moved into the Correction work and had forgotten about it. This just removes the gain  and calibration fields, these are found by calling the Corrections method for any Collection. It returns the adjustments if required to any given sensor/datalogger/stream set.

junghao (approved):

PR #1543, opened 2022-11-24
update the response snippets and initial channel + component files

ozym:
This is an initial commit to update the channel and component files.

I don't expect this to be the final version but simply a starting point and in some ways an illustration of how it will work.

I have also added a simple check to make sure the response file names exist as files to avoid typos as much as possible. This isn't used for anything currently, but is the major part of building future stationxml files.


salichon:
Great 

_Channels - > data loggers
Components -> sensors + recorders_ 
This will help for some instrumental  "functional sketches" @elidana   ;)

salichon (approved):
so far so good aye.

elidana (approved):
haven't looked much into details, and happy to do so if helpful, but looks pretty good!

ozym:
We will need to update this as we work through names and configurations.

PR #1550, opened 2022-12-02
create a separate file to list stations for manually collected data

elidana:
How does this look?
I've used "samples" as file name (using the GeoNet data glossary of terms as inspiration, https://doi.org/10.21420/XQS0-0Z48) but happy to change to a better name if there are different opinions.


rumachan (approved):
I'm quite happy with that. Happy to use any filename that works. 

This will have downstream consequences for Tilde meta-data I assume, so you'll want to let that team (Callum especially?) know.

ozym:
@elidana looking good, could I get you to restore the stations.csv file to the original state. This should allow the current code to work and tests passing downstream etc.  I can then look at adding the backend support.

elidana:
How about now? not sure if you wanted the `samples.csv` to also include the data, but if not I can update that file too and remove the MC records for now

ozym (approved):
Looks good, there will be some follow up PRs

elidana:
nice! thanks!!

PR #1552, opened 2022-12-03
add samples tables, cleanup testing and csv headers

ozym:
This is a follow up PR to add the surrounding code for the new `samples.csv` table.  This table is currently a copy of the structure of the `stations.csv` file but that may change in the future so I have implemented an actual `Sample` type.

This PR also takes some updates to address a couple of outstanding issues. The primary one is to remove the constraint on the csv columns, and what rows a configured as which. This is to allow current applications to remain working even if the file format changes, so long as the required columns retain the same names. This is the `Header` code and has been designed to be as least disruptive as possible to the existing code.

For help with testing, I've added an `Id` and an `Overlaps` function to  the `Sample` and `Feature` functions. I've also updated the tests for these two tables using the `Set` mechanism. 

These updates, and related `meta` testing code will be progressively inserted into the current table code.


meta/headers.go
```diff
@@ -0,0 +1,60 @@
+package meta
+
+import (
+	"sort"
+)
+
+// Header is used to manage the encoding and decoding of csv formatted files by using their column headers.
+// This allows adding or removing columns in the file without breaking existing applications that may not
+// understand or expect extra columns.
```
CallumNZ, line 9: Is this feature covered in the test case? From what I can see it's just testing difference in order.

ozym: Doesn't look like it. Will add one.

ozym: That took longer, than I thought.  Have to swap one of the maps around and add an existence check to avoid overwriting the default (0) map values if not found.

CallumNZ (approved):
Looks good

PR #1558, opened 2022-12-08
add points to separate manully collected data from sites.csv

elidana:
This pull request is completing the separation of manually collected data from continuously recording stations.

Some preliminary pull requests have introduced the
`network/samples.csv` file and associated checks to list all locations where manually collected data can be sampled.

In each of these locations multiple samples can be collected at slightly different locations. This was previously handled in the common file `network/sites.csv`, that is used for seismic stations and alike to distinguish different continuously recording sites at the same location.

To avoid conflicts and downstream issues, the same concept is here applied to manually collected data.

I have created a file named `network/points.csv`.

I'm not 100% convinced about the name, and keen to get further suggestions on that. The only other one that came to mind was "subsamples.csv" but wasn't really sold on that.

Have tried to look at NEMS (https://www.nems.org.nz/documents/) for inspiration, but haven't found a one to one match (and haven't spend enough time on that)

- [ ] once/if approved we will need to create another pr to remove "points" from "sites" :exploding_head:


elidana:
ok, that one didn't work. I'll split this pr into two parts
This one will be introducing "points.csv" and I'll generate a second one to only remove manually collected data from stations.csv

ozym:
This sounds good, something that may need some thought is whether when there's the same code for samples.csv and stations.csv, that the locations (i.e. lat/lon) should match.  This wouldn't hold at the points/sites level as they are more specific to the measurement.

elidana:
oh thanks @ozym !! would you have some suggestion for a better filename?

elidana:
and on the coordinate side of things, would that mean we should add a test to ensure that lat/lon of stations and samples that have the same "site" code are the same?

rumachan (approved):
This will be great!

ozym:
@elidana @rumachan this will needs some work, which I'll do, to fix the conflicts and get it up to date.

The main question in my head was still the names of things, is "Points" the right name (and I also don't like subsamples).

In conversation I can imagine using phrases like:

"The sampling point", "The sample point", "The measurement point", "The measurement place", 

But "point" keeps coming back, so unless there's a better suggestion I'll go with that.

elidana:
thanks @ozym ! I'm also happy to resolve conflict and update this pr (unless you're thinking of different stuff). if we're happy with points let's go with it.
Otherwise (just to keep them connected), my only other idea was a composite name like `samplesites.csv`
but I wasn't 100% convinced

ozym:
composite names always look bad as they miss the hyphen, I'm happy with points at the moment.

I will need to write the wrapper go code but that doesn't have to be in this PR, so if you want to handle the data I can look at the code (and downstream impacts, i.e. tilde manual collected etc.).

elidana:
ok, thanks Mark, let's stick with points then. I will update this pr and ping you and @rumachan again once ready. THANKS for looking into this!!

elidana:
I've tried to remove the "points" from `sites.csv` but the samples.go test is failing. So will need a step by step approach

1. add points
2. modify test
3. remove points from sites.

I will restore this pr and only add the points.csv file with this

elidana:
I've messed up this one, was quicker to create a new one. 
will close , and have moved the changes here: https://github.com/GeoNet/delta/pull/1856

PR #1563, opened 2022-12-09
update responses for Q330S 10Hz and derived responses

ozym:
Another couple of updates found while working on stationxml updates.

- The 10Hz response for Q330S  incorrectly had an input sampling rate of 1Hz, should be 10Hz
- The derived responses were not being limited by their sampling rates (too many responses).


CBurton90 (approved):

mnaguit:
Oooops, I'm late for the party. Thanks for sharing @ozym & for reviewing @CBurton90 😄

PR #1564, opened 2022-12-11
update tests to remove exec call, replace with cmp

ozym:
This is a bit of housework prior to go 1.19 update.

It replaces the original exec to diff checks with the use of the google cmp module.


Mossman1215 (approved):
much tidier.

PR #1566, opened 2022-12-12
update to go 1.19

ozym:
The main change outside of the 1.19 update (travis docker vendor) etc. is the deprecation of the `strings.Title` function which is picked up in the linter.

The replacement was added after 1.16 so I can slip in a change prior to this one without updating to go 1.17 etc.

There are a couple of other updates mainly related to the linter (e.g. pointers on loops) and a few smaller cleanups.


Mossman1215 (approved):
Module changes make this one a little scary looking but the actual changes are minimal

PR #1567, opened 2022-12-13
Add individual GNSS antenna calibration files from GA

aleks-spes:
Calibration files from Geoscience Australia for brand new GNSS antennas.


meganmadley (approved):
LGTM!

PR #1571, opened 2022-12-18
Add PTX-1830-LAND w/ Centaur for derived fresh water depth

CBurton90:
This is the only difference I can see between the Centaur and Q330 configurations with PTX-1830-LAND sensors. Added in preparation (or caution) for the Horomatangi Reefs tsunami gauge deployment (likely Thurs/Fri this week).


meganmadley (approved):
LGTM

salichon (changes requested):
hmm maybe needs a bit of recheck for me 
but whats the diofference with GLKZ ? 
and the   "Nanometrics Dataloggers Connected to Fresh Water Depth Sensors": 

I reckon everything is already setup for some of the datalogger  types

salichon:
Also the "spirit" might be  to mention the combo with the fresh water label ?

CBurton90:
> hmm maybe needs a bit of recheck for me but whats the difference with GLKZ ? and the "Nanometrics Dataloggers Connected to Fresh Water Depth Sensors":
> 
> I reckon everything is already setup for some of the datalogger types

When integrating the Centaur response into delta I believe @ozym already updated the combination with the freshwater druck ptx-1830. As I said in the comment this is the only discrepancy I can find between how we do it with the q330 vs the Centaur so to be safe I have added it in. There is a decent chance its not even needed as I can't find the `"WaterDepth1"` or `"WaterDepth10"` filters used. I'd say we should just let Mark confirm when he returns from leave.

CBurton90:
> Also the "spirit" might be to mention the combo with the fresh water label ?

`PTX-1830-LAND` is used for fresh water installations of the `PTX-1830`

mnaguit (approved):

ozym:
This is only used for the "derived streams", these are not generated on the instrument. These are built as part of the collection system, basically they will build a new stream "LTH" / "BTH" from the incoming streams (LTZ/BTZ) where the conversion factors for the sensor are applied to convert into water depth (10th mm).

ozym:
As a consequence, it doesn't mater whether it's freshwater or not as that conversion has already happened.

ozym (approved):

salichon (approved):
cool 
thanks for monday explanations!

CBurton90:
@salichon @ozym @mnaguit please merge if you are happy (I cannot) otherwise if deemed unnecessary you can close!

salichon:
@CBurton90 I am happy, hence i merge|
image

PR #1574, opened 2022-12-19
HRMT Network Delta details

meganmadley:
Hi all, 
Used network/marks.csv for TGHO coordinates and elevation. 
Wasn't sure about depth in stations, so just used '0', possibly might have left it blank in sites, so can update both to a specific number or just add '0' to the sites one. 
Also with network/sites.csv wasn't sure about the "External GPS Device" bit


elidana (dismissed):
thanks for doing this @meganmadley !

- for the depth, I think we should be consistent in both stations and sites.csv and either use `0` or empty value. I can see that normal tide gauges has empty value for depth in stations and sites, so can you please leave it empty in both files for HRMT as well?

- for start date, can you please change to the beginning of the UTC date (we can use the same "convention" used for GNSS sites, and it will surely be before the Friday morning installation time (NZ time)

- for the "Survey" column, I had to dig a bit and that column is used to show how coordinates where obtained. Since we used those from the co-located GNSS TGHO, I think we can use  "Site Survey" (instead of external clock)

meganmadley:
Thanks @elidana have fixed as you've suggested :)

elidana (dismissed):
thanks @meganmadley !
looks good to me

meganmadley:
@CBurton90, @elidana and @ozym, I think we only have one sensor https://github.com/GeoNet/tickets/issues/12456#issuecomment-1358768431, so have removed line to `41` in network/sites.csv

elidana (approved):

CBurton90 (approved):

CBurton90:
Just a note the gps timing antenna on the Centaur will eventually give us a site location from `Internal GPS Clock`. But it looks fine using site survey as well.

PR #1577, opened 2022-12-20
OTNZ RCET New seismic  station BB+Strong

salichon:
RCET BB station with a strong motion "special Xmas ;) "

    for LC 10 (LH/HH-ENZ) aand LC 20 HN - ENZ 
    Sensor: Trillium Compact PH SN9367 and SM Fortis TF1119
    Datalogger: Centaur SN9441
    Centaur firmware v4.5.20


mnaguit (commented):
Hi @salichon . Looks all good to me, but I have these questions:

- can you remind me what is the Number entry in `connections.csv` (i.e. 0 for LC 10 & 3 for LC20)?
- for the station/site coordinates, do we want it consistent & precise up to the sixth decimal place?

salichon:
Kia ora morena @mnaguit 
a- To me :)  and from tall the delta reworks (still inl process for parts) 
 https://github.com/GeoNet/delta/pull/1325  and https://github.com/GeoNet/delta/pull/1324
 
- The datalogger's associated sensors via connection records and the pin numbers
- This might be not mandatory yet -  based upon this https://github.com/GeoNet/delta/pull/1324#issuecomment-1166423873
- Although it is good to get into the routine of checking and doing the pins allocations when several sensors are plugged onto  a datalogger. 

and indeed this deserves a bit of scrutiny :) .

b- About or the loc - i used the figures as they were and to make a difference between loc 10/20 -
 Indeed  they can handle same coordinates (4 digits)  most likely sensors are about same place.

mnaguit:
Thanks for sharing the link & wow the discussions are very extensive and too long to read. 
Nevertheless, I'll learn it slowly to incorporate in routine delta process.
All good now. Thanks much! @salichon

mnaguit (approved):

salichon:
@mnaguit ii gonna push th esite lat/lon  digits  to 5 digits roughly meters precision ;)

mnaguit (approved):

CBurton90 (approved):

salichon:
Kia Ora @ozym 
Adding the pin numbers in the connection.csv file seems to be the best practice.
It looks like it is optional when at least  dual sensor onto single datalogger   (user need to know if bank gains are different) 
The test.sh does not complain about numbering  the "pin number"
Is it all right like this upon your review? 
thanks!

ozym:
Thanks @salichon , yes it's a good idea. It's not needed right now, but it will be for the next round of stationxml generation. It's also only really needed for dataloggers which have different characteristics on different channels (i.e. HR models) but it's a good habit to get into (to avoid multiple sensors connected into the same "slot") as a check.

ozym (approved):

PR #1579, opened 2022-12-20
add stationxml response decoding and snippet tests

ozym:
This PR adds the code to manipulate parts of a StationXML response type ,  it can join and normalise responses from different sensors and dataloggers.

It can adjust for different calibration values, gains, or preamps. 

It can also handle scaling of polynomial responses in a similar mannner.

This code only impacts the current system by adding a test in the `resp` directory to check that the generated or edited xml response files can read and decoded.

The next PRs will relate to encoding of this type into actual StationXML versions, as with the full StationXML root elements.

### DIFF
--- a/internal/stationxml/response.go
+++ b/internal/stationxml/response.go
@@ -83,9 +83,7 @@ func NewResponse(opts ...ResponseOpt) *Response {
 		GainFactor:  1.0,
 		Preamp:      1.0,
 	}
-	for _, opt := range opts {
-		opt(&r)
-	}
+	r.Config(opts...)
 	return &r
 }
 
--- a/internal/stationxml/response.go
+++ b/internal/stationxml/response.go
@@ -1,0 +1,468 @@
+package stationxml
+
+import (
+	"encoding/xml"
+	"errors"
+	"fmt"
+	"strings"
+)
+
+var ErrInvalidResponse = errors.New("attempt to correct the wrong type of response, biases require a polynomial response")
+
+type ResponseOpt func(*Response)
+
+// Response is used for building an instrument response based on sensor and datalogger pairs. It makes no assumption about
+// the StationXML version, ideally it should encompass all required elements. The conversion from a bas Response to a
+// particular version is done via encoding interfaces.
+type Response struct {
+	Prefix       string
+	Serial       string
+	Frequency    float64
+	ScaleFactor  float64
+	ScaleBias    float64
+	GainFactor   float64
+	GainBias     float64
+	GainAbsolute float64
+	Preamp       float64
+
+	sensor     *ResponseType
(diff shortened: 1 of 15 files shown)


internal/stationxml/response.go
```diff
@@ -0,0 +1,468 @@
...
+					}
+				case c > 0:
+					stage.Polynomial.Coefficients[0] = PolynomialCoefficient{
+						Number: stage.Polynomial.Coefficients[0].Number,
+						Value:  r.ScaleBias,
+					}
+				}
+
+				// update the overall instrument polynomial
+				sensor.InstrumentPolynomial.Coefficients = append([]PolynomialCoefficient{}, stage.Polynomial.Coefficients...)
```
bpeng, line 196: This may only get the last stage, is it right?

internal/stationxml/response.go
```diff
@@ -0,0 +1,468 @@
...
+// NewResponse builds a Response with the given options.
+func NewResponse(opts ...ResponseOpt) *Response {
+	r := Response{
+		ScaleFactor: 1.0,
+		GainFactor:  1.0,
+		Preamp:      1.0,
+	}
+	for _, opt := range opts {
+		opt(&r)
+	}
```
bpeng, line 88: can use the `Config` func below

internal/stationxml/response.go
```diff
@@ -0,0 +1,468 @@
...
+					}
+				case c > 0:
+					stage.Polynomial.Coefficients[0] = PolynomialCoefficient{
+						Number: stage.Polynomial.Coefficients[0].Number,
+						Value:  stage.Polynomial.Coefficients[0].Value*r.GainFactor + r.GainAbsolute,
+					}
+				}
+
+				// update the overall instrument polynomial
+				sensor.InstrumentPolynomial.Coefficients = append([]PolynomialCoefficient{}, stage.Polynomial.Coefficients...)
```
bpeng, line 230: same as above

internal/stationxml/response.go
ozym: Yes,  there can only be one polynomial stage by definition, so the last one (or none) will be fine.

internal/stationxml/response.go
ozym: Hadn't thought of that, easy enough.

internal/stationxml/response.go
ozym: as above, can only be one.

bpeng (changes requested):
code looks good, minor checks

ozym:
Thanks @bpeng have updated the Config bit, the other parts are okay.

bpeng (approved):
OK

PR #1553, opened 2022-12-06
GNSS site 2004 information

aleks-spes:
Hey @meganmadley! Could you please review it?

PR #1580, opened 2022-12-21
Update info for GNSS site 2004

aleks-spes:
The update to keep metadata about the pre-event measurements.

Also, there were performed survey-type measurements at the site. For more information about measurements contact LINZ (https://www.linz.govt.nz/)

The site information available on the LINZ Geodetic Database web page: https://www.geodesy.linz.govt.nz/gdb/index.cgi?mode=gmap&sessionid=10160578311891670209733&code=2004&action=+setfoundmarks+updatelist&foundmarklist=2004&listaction=clear+add&mark=


meganmadley (approved):
LGTM. 

Just adding this merged 2004 PR link as it helped me check dates https://github.com/GeoNet/delta/pull/1553/files

ozym:
Is this real, or should the test be updated:

```
sessions_test.go:76: unknown satellite system: GPS+GLO+GAL
```

ozym:
> Is this real, or should the test be updated:
> 
> ```
> sessions_test.go:76: unknown satellite system: GPS+GLO+GAL
> ```

Currently have "GPS", "GPS+GLO", and "GPS+GLO+GAL+BDS+QZSS"

meganmadley:
@aleks-spes? Not sure if it could have been because of some of Sigrun's specifications, so it could be right

network/monuments.csv
```diff
@@ -1,5 +1,5 @@
 Mark,Domes Number,Mark Type,Type,Ground Relationship,Foundation Type,Foundation Depth,Start Date,End Date,Bedrock,Geology
-2004,,Forced Centering,Steel Mast,0,Reinforced Concrete,0,2000-06-19T00:00:00Z,9999-01-01T00:00:00Z,,
+2004,,Forced Centering,Steel Mast,-1.35,Reinforced Concrete,0,2022-10-26T00:00:00Z,9999-01-01T00:00:00Z,,
```
elidana, line 2: I think the monument height should be 0 in this case, since the geodetic mark is on the ground?

elidana (changes requested):
Hi @aleks-spes , all looks good to me, but I think the monument height should now be set to `0`?

elidana:
for the session info, I think the info is real but unusual as it is a setting only used for GNS Science survey style recording. 
@aleks-spes , I suggest we can use `GPS+GLO` here, rather than changing the code...this measurements are going to be unusual anyway

network/monuments.csv
aleks-spes: But what type of the mark type should we use in this case?
Shoud we use different Start Date (2000-06-19T00:00:00Z) to be aligned with the mark.csv file?

CBurton90 (dismissed):

network/monuments.csv
elidana: oh good question!! definitively not `steel mast`, you're right. 

and sorry I don't have an answer... currently allowed options in delta to describe a monument are:
```
Mark Type,Type
Forced Centering,Pillar
Forced Centering,Shallow Rod / Braced Antenna Mount
Forced Centering,Steel Mast
Forced Centering,Wyatt/Agnew Drilled-Braced
Iron Rod,Unknown
Not Specified,Steel Mast
Not Specified,Unknown
Unmarked Mark,Steel Mast
Unmarked Mark,Unknown
```


in the [LINZ database 2004 page](https://www.geodesy.linz.govt.nz/gdb/index.cgi?mode=gmap&sessionid=10160578311891670209733&code=2004&action=+setfoundmarks+updatelist&foundmarklist=2004&listaction=clear+add&mark=), this geodetic mark is described as:
Mark type: Pin
Beacon type (monument type): Not Beaconed

we currently don't have `pin` as an allowed Mark Type so not sure what we should use here.

ozym: and in the appropriate tables.

elidana: oh nice, thanks @ozym

aleks-spes: If we add the  new mark type - pin, what monument type should we use?

elidana: I have been looking around comparing what we have in delta with descriptions in the LINZ db (did some random checks, so a deeper and more robust check might be useful). 

Maybe we can use "unkown" for now maybe? It has been used for sites like WRA1 (that was a forced centering mark type, but still something like a "metal thing in the ground".

Alternatively we can have a look at what IGS would use for similar use cases (even though IGS should only have continuous type monuments rather than survey marks)

ozym:
This will not pass the tests unless either:

- the session code is changed to `GPS+GLO`
- the list of possible session codes is updated to add `GPS+GLO+GAL`

@aleks-spes @elidana

meganmadley:
Have closed the loop on the "GPS+GLO" stuff but @aleks-spes can you make the final fixes to the network/mounument file (from @elidana comments above), then we can get the loop closed on this

aleks-spes:
Thanks everybody for your help!
I still don't want to push these changes into Delta. So, will try to figure out how to fit this information in accordance with Delta's requirements in the best way. I am closing this PR for now.

PR #1586, opened 2022-12-22
HRMT new tsunami gauge site equipment install

meganmadley:
Got the start time from @CBurton90's image in slack 2022-12-22T00:31:15

image


elidana (commented):
hei @meganmadley , not 100% sure but I wonder if we also need an entry in the `install/connections.csv` file with something like
`HRMT,40,Horomatangi Reefs,,0,2022-12-22T00:31:15Z,9999-01-01T00:00:00Z`
?

will continue review other bits now

salichon (changes requested):
Ciao there indeed 
You re   missing the bit of virtual "wire"  between the datalogger and the sensor
as in the `connections.csv` 
all good to have  `HRMT,40,Horomatangi Reefs,,0,2022-12-22T00:31:15Z,9999-01-01T00:00:00Z` (pin number set to `0` or `' '` to default either ) -

CBurton90:
@meganmadley I think we can put the sensor dip as `90` (similar to NBRT/TAUT) as Jackson confirmed they are a vertical mount. Thanks for using my screenshot (haha).

salichon (approved):
sweet as a mince pie to me

meganmadley:
Thanks @elidana and @salichon, have added the connections file

meganmadley:
Have added the sensor dip @CBurton90 :)

CBurton90 (dismissed):

salichon:
Sensor dip should not have too much importance since we are on a  pressure measurement

meganmadley:
Oh well, it's been changed anyways

elidana:
have checked with @salichon the scale factor, good to keep 1 as it is freshwater

elidana (dismissed):
looks good, thanks @meganmadley !

ozym (dismissed):
It shouldn't have a dip as it's a pressure sensor (i.e. the sensor points in all directions) so we usually use 0.

meganmadley:
Changed sensor dip back to 0 (@CBurton90 and @ozym)

ozym (approved):

elidana (approved):
thanks for changing sensor dip back to 0!
I think is now good to be merged? (after squashing all commits)

meganmadley:
Thanks everyone! :)

CBurton90:
@meganmadley Sorry about that!

@ozym might be worth considering an updated PR in the new year then to revert all to 0 as there seems quite a mix between 0 and 90

salichon:
@CBurton90 you mean a general  Gauge dips deep house cleaning in delta  i guess

CBurton90:
> @CBurton90 you mean a general Gauge dips deep house cleaning in delta i guess

@salichon now you are speakin ma language

PR #1589, opened 2023-01-18
NOVZ Q330 to Centaur - 3+3 channels (10/30)

salichon:
- Short Period+ Infrasound datalogger swap


CBurton90 (changes requested):
The change is a q330 to Centaur

No TitanSMA has been installed?

PR #1590, opened 2023-01-18
MTVZ  datalogger swap to Centaur

salichon:
EH short period 3 channels


CBurton90 (changes requested):
Can we add the `4.5.20` firmware for Centaur `9730`?

salichon:
Done @CBurton90  thanks

CBurton90 (approved):

PR #1595, opened 2023-01-18
Add a telemetry table

ozym:
This PR tries to handle an edge case whereby the datalogger and the sensor are not at the same location and they are connected via analogue telemetry.  Either using FM radio or via leased-line modems.

This should add an extra stage in the response with notionally a gain of 1.0 (it's usually quite hard to work out the actual gain, but if this is done then the adjustment can be made here).

The main part is to add the `telemetries.csv` file and an extra related "Correction" option (very similar to the Preamp stage).

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -24,6 +24,7 @@ Meta information for the GeoNet equipment network.
 * `components.csv` - Individual sensor elements including measurement position and responses.
 * `channels.csv` - Individual datalogger recording elements including digitiser position, sampling rate, and responses.
 * [`preamps.csv`](#preamps) - site specific settings applied to individual datalogger pre-amplification that may impact overall sensitivities.
+* [`telemetries.csv`](#telemetries) - site specific settings applied to datalogger and sensor connections that may use analogue telemetry.
 
 * `cameras.csv` - Installed field cameras.
 * `doases.csv` - Installed field DOAS (Differential Optical Absorption Spectrometer) equipment.
@@ -360,6 +361,19 @@ and the expected response. Some digitisers have different nominal responses for 
 | _Location_ | Recording sensor site _Location_ |
 | _Subsource_ | The sensor channel _Subsource_ which has the preamp configured (e.g _"Z"_). An empty value indicates all channels have this setting for the provided _Location_.
 | _Scale Factor_ | The datalogger pre-amp scale factor used for this time span. These tend to be integer steps and may be referenced as **gain** settings.
+| _Start_ | Gain start time|
+| _Stop_ | Gain stop time|
+
+#### _TELEMETRIES_ ####
+
+Sometimes the datalogger and the sensor are not at the same location. Usually this means there is some form of analogue link between the two, either a dedicated
+telephone line, or an FM radio link. This table allows this to be documented, and provides a mechanism to adjust the signal gains if known.
+
+| Field | Description | Units |
+| --- | --- | --- |
+| _Station_ | Datalogger recording _Station_|
+| _Location_ | Recording sensor site _Location_ |
+| _Gain_ | The telemetry gain factor for the analogue link, this represents the amplification of the signal if appropriate, an empty value is assumed to be 1.0
 | _Start_ | Gain start time|
 | _Stop_ | Gain stop time|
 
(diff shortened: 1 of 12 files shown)


staylorofford (approved):
A lot of work for one site, but LGTM

ozym:
@staylorofford that's only the guinea pig site, there are quite a few old ones to be added (original wellington network etc.).

install/README.md
```diff
@@ -363,6 +364,19 @@ and the expected response. Some digitisers have different nominal responses for
...
 
+#### _TELEMETRIES_ ####
+
+Sometimes the datalogger and the sensor are not at the same location. Usually this means there is some form of analogue link between the two, either a dedicated
+telephone line, or an FM radio link. This table allows this to be documented, and provides a mechanism to adjust the signal gains if known.
+
+| Field | Description | Units |
+| --- | --- | --- |
+| _Station_ | Datalogger recording _Station_|
+| _Location_ | Recording sensor site _Location_ |
```
elidana, line 375: other tables (preamps, gains) also include subsource (and even sublocation...)
shall we add the column here to, even if not used?

install/telemetries.csv
```diff
@@ -0,0 +1,2 @@
+Station,Location,Gain,Start Date,End Date
```
elidana, line 1: same as in comment above, shall we add susource/sublocation? (not a big fan of that but thought was worth mentioning)

elidana (commented):
Hi @ozym ,

thanks for that, it looks good and clarify very well what was discussed in https://github.com/GeoNet/delta/pull/1468.

I have on question and a comment (but mostly due to me being still quite a bit lost in the seismo-world)

- use of `gains`. We have been using "Scale Factor", "Scale Bias" in other tables (see e.g. preamps.csv, gains.csv). How does it fit with the "Gain" terminology used here for telemetry?
- additional channel/sensor info: again, in other places we also use subsource and sublocation (mostly to capture envirosensor). I don't think we need it here, but wanted to raise the point just to confirm.

Thanks!!

install/README.md
ozym: I was thinking of that too, swaying back and forwards related to what it does.  As this will not have a bias, should it be simply "Scale Factor" rather than bias.

(if a bias is added then the response becomes a polynomial and will make life tricky with responses).

elidana: oh great!! As I was suggesting to change it to "scale factor" but then got scared that I was missing something obvious :D

install/telemetries.csv
ozym: My thinking about Gain was it being a physical setting (i.e. a switch), e.g. preamp.

This would imply flicking over to "Scale Factor".

ozym: turns out my confusion was that I'd updated the docs for the preamp, but not the code (it didn't matter as the columns were the same).

I'll put in a PR with the name changes for preamp now.

elidana: thanks!!

install/README.md
```diff
@@ -363,6 +364,19 @@ and the expected response. Some digitisers have different nominal responses for
...
+| _Gain_ | The telemetry gain factor for the analogue link, this represents the amplification of the signal if appropriate, an empty value is assumed to be 1.0
```
elidana, line 376: change this to `Scale Factor`

meta/correction.go
```diff
@@ -37,26 +38,34 @@ func (set *Set) Corrections(coll Collection) []Correction {
...
+						for _, datalogger := range set.DataloggerCalibrationCorrections(coll) {
+							span, ok := span.Extent(datalogger.Span)
+							if !ok {
+								continue
+							}
+
+							corrections = append(corrections, Correction{
+								Span:                  span,
+								Polarity:              polarity.Polarity,
+								Gain:                  gain.Gain,
```
elidana, line 62: does this need to change to ScaleFactor as well? (there might be other places in the codes, only commenting here)

elidana (changes requested):
one small change in the README, sorry to be a pain!

elidana:
preamp updated in: https://github.com/GeoNet/delta/pull/1610

meta/correction.go
ozym: No, that one's okay. It's linking to the full Gain list (i.e. the gains.csv input), there will be a gain.Gain.ScaleFactor.

install/README.md
ozym: Done.

meta/correction.go
elidana: ah gotcha!! THANKS!!!

install/README.md
elidana: :pray:

elidana (approved):
looks good!

PR #1602, opened 2023-01-24
use lowercase j for all ServoTech T16 sensors

elidana:
Hi @rumachan ,
I noticed while reviewing https://github.com/GeoNet/tickets/issues/12320 that the model name for Servotech T16 sensors was not consistent.
This pr is using the nomenclature with lowercase `j` for all such sensors.
Hope is not going to cause downstream effects


rumachan (dismissed):
Well spotted. I simply copied what was in the ticket.

elidana:
thanks @rumachan !
Can I have another thumbs up please? Had to resolve a conflict and add a further commit to this one

rumachan (approved):

PR #1606, opened 2023-01-27
New Nanometrics Titan Accelerometer TACCL-N1 (standalone) and Titan Borehole Accelerometer TACCL-BH1

CBurton90:
Planned for deployment at the Christchurch Botanical Gardens Borehole. I'll begin adding response information next week, should be the same paz as a TitanSMA minus the digitiser response.


salichon (changes requested):
ola l @CBurton90 
- This needs a "," line 641  you re missing a field 
- i would swap  the 2 lines 

have a good WE

CBurton90:
@salichon thanks, updated

mnaguit (approved):
Looks good. Thanks @CBurton90

ozym (changes requested):
Can we drop the "Borehole" that's more a description rather than a model name/number unless it says that on the spec sheet of course.

CBurton90:
> Can we drop the "Borehole" that's more a description rather than a model name/number unless it says that on the spec sheet of course.

@ozym done

mnaguit (approved):

ozym (approved):

salichon (approved):

ozym:
For reference in the spec sheet.

```
AVAILABLE MODELS
TACCL-N1: Standard Model
TACCL-V1: Vertical Mount Model 
```

ozym:
@CBurton90 think this can be merged now, this is just updating the assets which we have and will install at some point.

PR #1608, opened 2023-01-30
RIZ sensor-recorder swap

mnaguit:
- from FBA-ES-T-ISO (7457) to TitanSMA (001913)
- TitanSMA firmware v4.5.22
- for LC 20 (HN-ENZ)

Note: heads up on `streams.csv`, from triggered to continuous. I used the closing date corresponding to FBA (instead of end of triggered files supposed to be in 2020).


salichon (approved):
haha just saw  WAKS turned in to the correct RIZ sensor  eventually 
all good as then! 
thanks

mnaguit:
> haha just saw WAKS turned in to the correct RIZ sensor eventually all good as then! thanks

Hahaha. True, I was scared I messed up delta on my 1st day at work! 🤣

salichon (changes requested):
you may have missed the RIZ strong motion datalogger !

install/connections.csv
```diff
@@ -739,7 +739,7 @@ RIVH,11,Lathrope Road,Portable B,,2007-03-27T00:00:01Z,2007-05-14T03:40:00Z
 RIVO,10,Riverton,Portable,,2014-03-19T21:30:00Z,2014-09-13T00:00:00Z
 RIZ,10,Raoul Island Met Station,,0,2009-05-18T00:01:00Z,9999-01-01T00:00:00Z
 RIZ,20,Raoul Island Met Station,,3,2009-05-18T01:00:00Z,2016-09-17T13:00:00Z
-RIZ,20,Raoul Island Met Station,Strong Motion,,2016-09-17T13:00:01Z,9999-01-01T00:00:00Z
+RIZ,20,Raoul Island Met Station,Strong Motion,,2016-09-17T13:00:01Z,2023-01-28T21:16:00Z
```
salichon, line 742: Perhaps would  you 
change for :
**"RIZ,20,Raoul Island Met Station,Strong Motion,3,2016-09-17T13:00:01Z,2023-01-28T21:16:00Z"**

just adding the 3 for the pin 3  of the related  datalogger with multiple instruments ...
thats all :)

salichon (approved):
i  m all good finally besides the connection pin number 
as a new best practice?

install/connections.csv
mnaguit: 3 added 😄 
another thumbs up/approval please @salichon

salichon (approved):
supa dupa

install/connections.csv
```diff
@@ -739,7 +739,7 @@ RIVH,11,Lathrope Road,Portable B,,2007-03-27T00:00:01Z,2007-05-14T03:40:00Z
 RIVO,10,Riverton,Portable,,2014-03-19T21:30:00Z,2014-09-13T00:00:00Z
 RIZ,10,Raoul Island Met Station,,0,2009-05-18T00:01:00Z,9999-01-01T00:00:00Z
 RIZ,20,Raoul Island Met Station,,3,2009-05-18T01:00:00Z,2016-09-17T13:00:00Z
-RIZ,20,Raoul Island Met Station,Strong Motion,,2016-09-17T13:00:01Z,9999-01-01T00:00:00Z
+RIZ,20,Raoul Island Met Station,Strong Motion,3,2016-09-17T13:00:01Z,2023-01-28T21:16:00Z
```
CBurton90, line 742: @mnaguit @salichon I'm not sure if we need to worry about the `3` pin as the EpiSensor is not connected to the Q330S+ at site rather it is connected to an Obsidian 4X datalogger/digitiser. So set up has gone from:

Trillium 120QA -> Q330S/6 (only using pin 0 i.e. channels 1,2, and 3)

Episensor FBA-ES-T-ISO -> Obsidian 4X datalogger (there is only 1 bank/pin so I think we leave it blank)

Let me know if this makes sense?

CBurton90 (changes requested):

install/connections.csv
mnaguit: Hi @CBurton90 . Thanks for double-checking.
To be honest, I haven't had a clear understanding yet of the 'pin' number. From related tickets (https://github.com/GeoNet/delta/pull/1324#issue-1222031439), does that mean `3 is for SM` channels and `0 for broadband` OR the 'pin' depends on the number of banks available in the datalogger? I'll tag @ozym here.

salichon: @CBurton90  Thansk ! 
Oh okay because this was not too  clear form the install in the ticket 
so  - no pins since the datalogger/connection is not very relevant/.  
and 
- `Kinemetrics | Obsidian 4X Datalogger | 656 | Raoul Island Met Station | Strong Motion | 2016-09-17T13:00:01Z | 9999-01-01T00:00:00Z`
 should updated I reckon @mnaguit

mnaguit: Yeah, a bit complex. This is easy to change now, but understanding what the entry means matters so that we can carry it on in the next delta PRs. So (pls correct me if I'm wrong), from what i gauge, this means that `0` is the entry for broadband and `3` for strong motion, if both sensors are plugged in the same datalogger. Then we leave the entry blank if only one sensor is plugged in a datalogger? @salichon @CBurton90

CBurton90: @mnaguit Yes that is correct. Applicable to dataloggers/digitisers with two banks (A for channels 1,2, and 3 = pin 0, B for channels 4,5, and 6 = pin 3).

Typically for us when two sensors are plugged into both digitiser banks the weak motion (either SP or BB) will be on A (pin 0) and strong motion on B.

The real benefit here is historically we had some tricky digitisers (Q330HR/HRS) that had a different sensitivity on bank A vs bank B (24 bit vs 26 bit) and that made problems accounting for this in response files.

and yeah I agree with @salichon for this PR I wouldn't worry about pins (two dataloggers operating separately)

ozym: @mnaguit what @CBurton90 said.

It's all about multiple sensors into the same datalogger, in some cases we have 4 sensors (e.g. microphones) so the datalogger will have a number of physical digitisiing channels (or pins), a Q330/6 has 6, and a Q330/3 has 3. (Obsidian 4X has 4 etc).  The numbering starts at 0, and if no value is given it is assumed to be 0.  

Each sensor may have multiple components, e.g. a strong motion sensor will have 3 (Z12) etc. and it is assumed that these are installed in a standard order, i.e. 0,1,2 say.  these will usually be plugged into a datalogger at a fixed offset (usually 3), so if plugged into the "second" bank in a Q330/6 they will be given a pin offset of 3 (and will add up to 3,4,5 when expanded).  Whether it is 0 or 3 will depend on convention and how the site was set up.

The main thing is to map the datalogger channel to the sensor channel, as we can either have different responses on them, or we can actually calibrate a channel if we want to.   (It also means we can do things like build a response for the mass positions, these would be extra pins on the end of a Q330/6, perhaps 6,7,8 for the first bank).

mnaguit: Thanks a lot @ozym @CBurton90 for the explanation. I'll take note so that I can be guided in the next PRs. 👍 👍

salichon: yup ;)

mnaguit:
@ozym @salichon @CBurton90 
`connections.csv` modified. PR now ready for your checking & approval.

ozym:
Has the "Obsidian 4X Datalogger"  been removed as well (i.e. the sensor is no longer there). If so, this should be removed from the `dataloggers.csv` file as well.

mnaguit:
> Has the "Obsidian 4X Datalogger" been removed as well (i.e. the sensor is no longer there). If so, this should be removed from the `dataloggers.csv` file as well.

Ah yeah, good pickup @ozym, many thanks. The combo Obsidian-FBA is now replaced by a recorder (TitanSMA). And the existing sensor for LC 10 (Trillium) is plugged to Q330S/6.

salichon (approved):
looks good :) thanks!

CBurton90 (approved):

mnaguit:
Thanks @CBurton90 @salichon @ozym ... merging this now.

PR #1610, opened 2023-01-31
update preamp

ozym:
Turns out the documentation was updated, but not the code, probably just waiting on a PR to be made.

Here's the updates to change over from Gain to "Scale Factor"


elidana:
thanks @ozym ! 

looks great, and awesome to have name alignment between code and documentation!

elidana (approved):

PR #1611, opened 2023-01-31
add quality table

ozym:
An proof of concept ... only a "Fault" column to indicate it shouldn't be used. Need to think about whether to have a specific set of faults to avoid the free form fields.


elidana:
this is a nice PoC but I'd like to give it some proper thoughts, in view of many "quality and data quarantine" discussions we had for seismic, envirosensors and other types...sorry @ozym !!

ozym:
Will revisit.

PR #1612, opened 2023-01-31
Response added for Nanometrics Titan TACCL-N1 and TACCL-BH1

CBurton90:
This is a new configuration of two new strong motion sensors connected to a Nanometrics Centaur. The deployment will be at Christchurch Botanical Gardens and is planned for next week (beginning 06/02/23).


salichon:
Thanks @CBurton90 
- the filter is already present for the titan - so all good 
This is looking good so far 
- i ll go through the PAz and details at some point. 

Combo will be for sometime  `Centaur/Titan` all nanometrics strong at borehole unicorn CBGS :)

ozym (commented):
I feel we should drop the "Borehole" part in the name of the sensor, i.e. "Titan TACCL-BH1" should be enough (the BH means borehole).

I'm trying to get the naming conventions between the various parts standardised.

e.g.
"Nanometrics Titan TACCL-BH1" is easier to handle as an asset name.

with the assumption that all of these will be 4g  (otherwise we would need that as a suffix or we would need to manage gain settings).

I think the "go generate" should have created some other files in "resp/auto" which can be used in the new system.  These need to be linked to the components file.

CBurton90:
> I feel we should drop the "Borehole" part in the name of the sensor, i.e. "Titan TACCL-BH1" should be enough (the BH means borehole).
> 
> I'm trying to get the naming conventions between the various parts standardised.
> 
> e.g. "Nanometrics Titan TACCL-BH1" is easier to handle as an asset name.
> 
> with the assumption that all of these will be 4g (otherwise we would need that as a suffix or we would need to manage gain settings).
> 
> I think the "go generate" should have created some other files in "resp/auto" which can be used in the new system. These need to be linked to the components file.

@ozym changes made. XML files now in resp/auto/ additionally.

ozym:
Sorry, just noted. Could you switch out "Nanometrics Inc." for simply "Nanometrics" (part of the consolidation work).

ozym (dismissed):
looks good.

salichon:
sorry ticked the wrong  tick for review ...

salichon (approved):
looks good to me  wrt to the docs of https://github.com/GeoNet/tickets/issues/12600#issue-1557389044 and IRIS nrl

@CBurton90 eventually only a name change for the same sensor Titan  - only the casing change as i understand here 
thanks !

mnaguit (approved):
Two of the best delta champions in GeoNet already approved this PR, so I'm pretty confident this is all good. 😄 
To share, what I observed in the files:
- looks like all entries for response are the same for both equipment
- just to note: the difference I spotted was the channel code: ZNE for TACCL-N1 and Z12 for TACCL-BH1
Thanks @CBurton90  👍

salichon:
Does  a change  for "Nanometrics" for  "Nanometrics Inc."  affect  the test with regards to the assets.csv ?

ozym:
> Does a change for "Nanometrics" for "Nanometrics Inc." affect the test with regards to the assets.csv ?

Not for the response part, but it may cause confusion when we have dual names for the same company (actually kicks in with Kinemetrics between Sensors and Dataloggers).

ozym (approved):

mnaguit (approved):

CBurton90:
@ozym @mnaguit would one of you mind merging please : )

PR #1616, opened 2023-02-02
add role to co-located RLGLK and GLKZ

elidana:
this will allow us to indicate that the two dataloggers are co-located

@ozym , @rumachan can you please confirm that:
- you are happy to add "Broadband" role to GLKZ
- you are happy to use the newly introduced "Envirosensor" as a role category


rumachan (approved):
Excellent solution. Clear and simple

PR #1617, opened 2023-02-02
Envirosensors cheatsheet

elidana:
starting some documentation for envirosensors. Still a draft
- [x] need to double check file list
- [ ] need to check General and Field requirements
- [ ] need to write what Delta checks are in place
- [x] need to discuss with @rumachan any additional useful "reminder"

### DIFF
--- a/docs/sheets/ENVIROSENSOR.md
+++ b/docs/sheets/ENVIROSENSOR.md
@@ -1,0 +1,83 @@
+# COMMON ENVIROSENSORS TASKS
+
+* :hammer: [Combined editing steps](#steps)
+* :scroll: :one: [Creating updating or closing an __Envirosensor__ site](#openclose) :dragon: 
+
+
+Envirosensors are installed at monitoring __stations__ and measure data from a given __feature__. A __station__ can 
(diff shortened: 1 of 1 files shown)


elidana:
this is still WIP, but my benefit from a second set of eyes!

rumachan (changes requested):
A couple of things and a quick glance - its difficult to read md like this.

`The sampling rate needs to be within the expected list: 0.1, 1, 10, 50, 100, or 200 Hz.` For this platform, the only (so far) used sampling rate is one sample per 10 minutes. This is indicated by a sampling rate of `-600`.

Lots of comments about failing the tests. Mine never fail PR tests as I always repeatedly run test.sh locally as I make changes to the various files. Suggest this practice be included as a strong recommendation.

I'm sure there other things we can add, but lets get this out therenow.

elidana:
thanks for looking at that Steve! I have updated the document with your comments, and if you're happy agree that we can start adding this and build on it if useful.

rumachan (dismissed):
I added a few small things. Think we should merge this and then think about how we get others (than me) involved in these meta-data changes.

elidana:
thanks @rumachan ! I've just changed a typo (sesnor to sensor) and rebased.

Can you please give another 👍 to this up and then we can merge?

rumachan (approved):

PR #1618, opened 2023-02-03
add stationxml decoder elements

ozym:
This PR follows on from the encoder work.

Given a byte representation of an XML file, the decode will, depending on required version, extract the appropriate parts and return a cleaned byte slice. This code is primarily used to compare two different StationXML files and attempts to make the options uniform allowing comparisons.

### DIFF
--- a/internal/stationxml/decoder10.go
+++ b/internal/stationxml/decoder10.go
@@ -1,0 +1,171 @@
+package stationxml
+
+import (
+	"encoding/xml"
+	"math"
+	"sort"
+
+	"github.com/GeoNet/delta/internal/stationxml/v1.0"
+)
+
+type Decoder10 struct{}
+
+func Decode10(data []byte) ([]byte, error) {
+	return Decoder10{}.Decode(data)
+}
+
+func (d Decoder10) Decode(data []byte) ([]byte, error) {
+
+	var root stationxml.FDSNStationXML
+	if err := xml.Unmarshal(data, &root); err != nil {
+		return nil, err
+	}
+
+	for n := range root.Network {
+		net := &root.Network[n]
+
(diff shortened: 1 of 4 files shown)


internal/stationxml/decoder10.go
```diff
@@ -0,0 +1,171 @@
...
+					stage.Coefficients.ResourceId = ""
+				}
+				for v := range cha.Response.Stage {
+					stage := &cha.Response.Stage[v]
+					if stage.Decimation == nil {
+						continue
+					}
+					stage.Decimation.Delay.Value = 0.0
+					stage.Decimation.Correction.Value = 0.0
+				}
```
bpeng, line 138: the above 5 loops can be combined into 1, use something like
```
if stage.Polynomial != nil {
	stage.Polynomial.Name = ""
	stage.Polynomial.ResourceId = ""
}
...
```

internal/stationxml/decoder10.go
```diff
@@ -0,0 +1,171 @@
...
+			}
+
+			for i := range sta.Comment {
+				sta.Comment[i].Id = stationxml.CounterType(i + 1)
+			}
+
+			net.StartDate = stationxml.DateTime{}
+			net.EndDate = stationxml.DateTime{}
+		}
```
bpeng, line 150: channels are not sorted as version11&12, is this intentional?

bpeng (changes requested):
code looks good, small suggestions
separate structs for 3 versions have duplicate code, not sure if it is feasible to use one struct

ozym:
The duplicate code is related to importing different modules. Will update other comments.

ozym:
@bpeng have updated code.

bpeng (approved):
OK

PR #1621, opened 2023-02-07
add CUSPM and CUSP-Me

ozym:
This PR adds the two CUSP building array datalogger/sensor combinations. It's assumed a CUSP-3C style response.


mnaguit (approved):
Hi @ozym . Very happy to see this metadata enhancement, especially for structural array 👍

PR #1627, opened 2023-02-17
fox crackmeter reset values

rumachan:
Crack meters at FOXKP, FOXVF, and FOXVV have a limited displacement range. If a crack expands/contracts too far, it will exceed the range of the meter and it will need to be reset to zero to continue to record further expansion/contraction. If the logger code is changed, then a zero reset is also required. In essence, this is similar to a re-centre for a seismic sensor.

We need to see the total displacement across a crack from when a crack meter was installed. At installation time, the crack meter output is zeroed, so the meter measures offset from the initial length. We therefore need to apply an offset to any data after a zero reset, so that the subsequent observations remain relative to the install time, not to the most recent zero reset time. Each zero offset value will need to be added to the previous offset value so that all data values remain relative to the install time. This ticket is for the first zero resets. There are, as yet, no procedures for this. Jackson and I agreed that we needed to correct data after a zero reset and this ticket documents our first attempt at doing that.

The following steps are required:

1. Examine the data in Tilde (https://tilde.geonet.org.nz/) and determine the time and magnitude of the data offset
1. Using a local copy of the data, apply that offset to be sure the value is correct
1. Add that offset to Delta, in `install/gains.csv` as _Scale Bias_ from the zero reset time. This will result in subsequent values being corrected for the zero reset, but some data will remain in Tilde for which the zero offset value still needs to be applied
Retrieve data from Tilde and manually apply the zero reset correction to data already in Tilde before the Delta change, and then re-upload to Tilde. 
There are also potentially situations where the zero reset process does not complete before the logger takes a sample (the reset procedure takes 3 mins). In that case, there is a "data spike" and that sample value should not be used; we should remove it from Tilde.

The data in Geonet raw archive will not be touched, just the derived data in Tilde. Once Delta contains the zero reset values in `install/gains.csv`, anyone who accesses the raw archive would be able to work out what to do. 

The changes proposed for `install/gains.csv`, _Scale Bias_ column are below (and explanation in the first comment of this pull request). 
There will be no existing values for these site/sensor combinations as this is the first zero reset.

```
FOXVV 05 is -4.625630 from 2023-02-09 03:50:01
FOXVV 04 is -8.570246 from 2023-02-09 03:50:01
FOXKP 04 is 9.177308 from 2023-02-07 23:50:01
FOXVF 05 is -5.494182 from 2023-02-09 02:50:01
FOXVF 04 is -2.176395 from 2023-02-09 02:50:01
```


elidana:
**FOXVV**
_Sensor code 05_
newplot (2)

last value before zero reset 2023-02-09 03:50:00 -4.625630
_there are no values between_
first value after zero reset 2023-02-09 04:30:00 -0.027247* 
*Jackson says first value will not be exactly zero, this represents 0.027 of one millimetre.
**Offset to apply to all values after 2023-02-09 03:50:00 is -4.625630. This is value to go in Delta.**

If we apply the offset, the result looks good.
foxvv_05

_Sensor code 04_
newplot (3)
last value before zero reset 2023-02-09 03:50:00 -8.570246
_there are no values between_
first value after zero reset 2023-02-09 04:30:00 -0.038722 
**Offset to apply to all values after 2023-02-09 03:50:00 is -8.570246. This is value to go in Delta.**

If we apply the offset, the result looks good.
foxvv_04

elidana:
@rumachan , thanks for doing that!! Lots of good work in here.

The general idea looks pretty good to me. I will now review individual data and add comments/questions here as I go (apologies for that, but I need to work on that in between other commitments)


**FOXKP** 
- Happy to leave the non-accounted offset  (Feb7 to Feb 8) for now and possibly : is it worth comparing with the GNSS site? The signal might be too and within GNSS processing noise. or check against other recorded data (crack displacement, for example) to see if there is anything that can explain it? 
- By looking at the time series, there is another "jump" back in december. Shall we include that one as well? 
newplot (1)

elidana:
@jacksonshanks , if you look at the comment above (https://github.com/GeoNet/delta/pull/1627#issuecomment-1447118527), you might see that we observe another FOXKP-04 offset in early December. Do you know the origin of that? Thanks!

rumachan:
@elidana @jacksonshanks I thought about this more after talking with Elisabetta. The early-December offset cannot be a zero reset, simply because the apparent length increases (by 10 mm). By definition, a zero reset resets the measured distance to zero. Pre-zeroing values can be positive or negative, but in both cases they must move towards zero. There must have been some unknown source of this step change. We therefore should not consider it in this ticket, but instead consider what yo do with such signal in terms of data QC/

elidana:
review checklist
- [ ] FOXKP-04
- [ ] FOXVV-05
- [ ] FOXVV-04
- [ ] FOXVF-05
- [ ] FOXVF-04

elidana (approved):
@rumachan , it took a while and apologies for that, but this one looks good and is a nice use of the `gains.csv` files. I did review the offsets you propose (https://github.com/GeoNet/delta/pull/1627#issuecomment-1450918761) and what you propose looks good to me. It is an arbitrary value we are imposing to the time series, so we might potentially miss some ongoing deformation while the crackmeter is reset to `0`, but I think that's the best we can do with this tricky issue. Thanks a lot for all your work on it!!

/cc @ozym , for your awareness

jacksonshanks:
thanks for the awesome work here everyone looking really good.
my two cents for the ceoment above @elidana and @rumachan  is that i was definately not a zero reset. no one on site for many montths before the visit last month. 

Its' worht pointing out that the jump at KP came from one of the less trust worthy, DP 9 the sisgeos) that we recently replaced two of and has a v complex code to make work so its possible that it was at a calculcation threshold and is it stretched out it jumped too much. 

i wont og into too much detail around this but woth keeping in myind if we continue to see this kind of change. 

J

elidana:
thanks @jacksonshanks , that's really useful to know.

@rumachan , I think we can merge this one now, shall I leave you the honors (since you did all the work!)?

PR #1632, opened 2023-02-23
GNSS ALS2 new temporary site antenna and reciever install

meganmadley:
Few things:
- install/antennas - SN needs checking off photos
- environment/visibility - made this up, so want to check photos
- network/monument - wasn't too sure? Basically a 2m pole driving into the ground with 0.5m above ground right?
- network/marks - "Auckland Landslide 2" or "Auckland Landslide Response 2", went with former to match Nelson, also got the coordinates off network/data/localities (https://github.com/GeoNet/network/pull/2931/files) this will need checking against receiver when reviewed (@aleks-spes could you check this please when you review?)

FYI have assigned three people as is a new site


aleks-spes (approved):
All good, if the information from the ticket matches with photos.

CBurton90 (approved):

PR #1633, opened 2023-02-24
setting up ALS2R

rumachan:
Setup as per https://github.com/GeoNet/tickets/issues/12747

Delta changes to the following

- network/stations.csv
- network/sites.csv
- data logger already in assets
- assets/sensors.csv, Pronamic rain gauge has no serial number, I don't believe they have serial numbers but @jacksonshanks may be able to confirm, but not required to proceed. Used 003 as previous two I used 001 and 002.
- install/dataloggers.csv
- install/sensors.csv
- install/streams.csv
- install/connections.csv
- install/gains.csv, 0.2 Scale Factor applied to sensor 01 for rain gauge bucket tip size of 0.2 mm, sensor 02 is 1mm size so no factor required
- environment/features.csv


CBurton90 (approved):

elidana (changes requested):
it all looks good but I think it is missing the firmware information.

Can you please add this entry in the `install/firmware.csv` (might be worth double checking with @CBurton90 or @jacksonshanks or @ricardojohansson if that's correct)

in install/firmware, change the following (already existing) line:
`Campbell Scientific,CR6,11770,CR6.Std.xx.xx,1901-01-01T00:00:00Z,9999-01-01T00:00:00Z,`

to
`Campbell Scientific,CR6,11770,CR6.Std.12.01,2023-02-23T03:59:00Z,9999-01-01T00:00:00Z,`

rumachan:
@elidana firmware added

elidana (approved):
looks good, thanks for adding the fw info!

CBurton90 (approved):

PR #1634, opened 2023-02-24
GNSS ALS1 new temporary site - antenna and receiver install

meganmadley:
Not sure about some details in monument file (length of rod). 
Coordinates off receiver. 
Antenna SN verified from photos from Dave and are in main ticket.


CBurton90 (dismissed):

aleks-spes (approved):
Looks good!

PR #1637, opened 2023-02-24
update ALS1R and ALS2R coordinates based on co-located GNSS

elidana:
this one is updating the envirosensors coordinates using the same as recorded on the co-located GNSS receiver


elidana:
just noticed this one from a while ago.
@Camasher or @CBurton90 , maybe you can review this on behalf of Steve?

CBurton90 (approved):

PR #1638, opened 2023-02-28
add stationxml-compare

ozym:
This application is used to verify that two StationXML files are the same within a set of encoding/decoding constraints. It's main use is to check update code returns the same values.

### DIFF
--- a/cmd/stationxml-compare/main.go
+++ b/cmd/stationxml-compare/main.go
@@ -64,6 +64,6 @@ func main() {
 			log.Fatal(err)
 		}
 	default:
-		log.Fatalf("requires two files to compare, or one to display")
+		log.Fatalf("requires two files to compare, or one to verify")
 	}
 }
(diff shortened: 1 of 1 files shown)


cmd/stationxml-compare/main.go
```diff
@@ -0,0 +1,69 @@
...
+			log.Fatal(err)
+		}
+
+		fmt.Println(cmp.Diff(string(first), string(second)))
+	case n > 0:
+		if _, err := decode(version, flag.Args()[0]); err != nil {
+			log.Fatal(err)
+		}
+	default:
+		log.Fatalf("requires two files to compare, or one to display")
```
wilsonjord, line 67: Should be "or one to verify", as per the help message?

wilsonjord (dismissed):
Nice clean utility, works for me (tested equivalence and non-equivalence)

cmd/stationxml-compare/main.go
ozym: Yup @wilsonjord , that was a case of updating the README after the code, I'll fix this and you'll need to give it another tick.

wilsonjord (approved):

PR #1640, opened 2023-03-01
update titan components

ozym:
For the new XML generation code, the component table needs to be updated. In this case the borehole titans.


mnaguit (dismissed):
Can this be possibly associated on why FDSN query does not return the new LCs for CBGS? Approving this now...thanks @ozym

ozym:
@mnaguit no, this is the new system. But it would have the same symptoms!

CBurton90 (dismissed):

salichon (dismissed):

mnaguit (approved):

salichon (approved):

PR #1648, opened 2023-03-10
Update firmware.csv

ricardojohansson:
new logger installed at TOUEM (no comms yet)


elidana:
have added the datalogger to the asset file as the build was failing

elidana (approved):
thanks @ricardojohansson !

also /cc @rumachan for awareness

PR #1649, opened 2023-03-12
Update firmware.csv

ricardojohansson:
New logger installed at rotomasodaspring


elidana:
@ricardojohansson have added the info in asset as the test was failing

elidana (approved):

PR #1652, opened 2023-03-16
Update firmware.csv

ricardojohansson:
New logger installed at TO063


ozym:
The new logger needs to be added to assets.

PR #1663, opened 2023-03-30
GNSS AUKT Alloy Swap

meganmadley:


CBurton90 (dismissed):

install/sessions.csv
aleks-spes, line 1: Hey @meganmadley ! Everything looks good, except session.csv file. I believe we don't need to update the string, because we haven't changed receiver type or session settings. What do you think?

meganmadley:
Only found one example of no change apart from dates in https://github.com/GeoNet/delta/blob/main/install/sessions.csv:
AUCK - lines 19 and 20

So will delete the change I made to the install/sessions file (FYI @aleks-spes)

aleks-spes (approved):
Looks good!

PR #1666, opened 2023-03-31
NNZ sensor swap

mnaguit:
- from CMG-3ESPC (T36195) to TH120-2 (2838)
- for LC 10 (HH/LH-ENZ) streams

Note: I didn't find any gaps in the data within the time period of the sensor swap. Let me know if the suggested closing-opening times are all good with you. 😄


CBurton90 (approved):

salichon (approved):
good as

PR #1677, opened 2023-04-27
add geomag APIM metadata into delta

elidana:
add APIM (Apia) geomag metadata.

For the sensor calibration, I have not used the -38000 absolute correction used for geomag stations installed in Antarctica.
Is that correct?


merijnthornton (approved):
Looks good to me.

elidana:
thanks for reviewing @merijnthornton ! Let me know if you would like me to produce the stationXML response file to have a look at the results

merijnthornton (commented):
The observatory grade proton is a gsm-90. Gsm-19 magnetometers are used in the absolutes hut, but are not part of the network. We may want to differentiate between gsm-90 and gsm-90F. The F indicates fast sampling (1sec). The non F model has been used in the past (5 sec sampling).

merijnthornton:
Hi Elisabetta, 

Thats correct, the Z Bias switch is just for higher latitudes:

Since the adjustable BIAS-circuit compensates the magnetic field components in 255 steps of 150 nT, it is
necessary to add a fixed bias (38000 nT) at higher fields like the Z-field at mid and high latitude.
This is done using the two switches on the circuit board (one for X and one for Z)
At high latitude (Z>38000) the Z-switch must be ON.
At lower latitude (Z<38000) the Z-switch must be OFF.

elidana:
message from @merijnthornton :
The proton was replaced in Apia (~8 May 2019). Old GSM90F1 was   8012579)

elidana:
@merijnthornton , all metadata should now be correctly included here (I hope!)

can you please have another look?

ozym (commented):
not sure this is the right model for apia Q330S/6

elidana:
thanks @ozym ! I was just about to update that bit.
I have two serial numbers for the two q330 installs: 3869 and 3781
would the model be Q330/6 then?

elidana:
that's the only pic I could find 
image
(from 2022)

elidana:
changed datalogger model to be Q330/6 and added the two install that I know about

tanjapetersen (dismissed):
looks good to me

elidana:
@tanjapetersen , this one had some conflicts to resolve once the SBAM pr was merged.
Can i please have another quick approval from you so that I can merge?

@ozym , hope that's ok for you but please if something looks odd let me know !

tanjapetersen (approved):

PR #1679, opened 2023-04-28
fix EYWM proton model

elidana:
installed proton at EYWM is GSM-90 model. Also adding the driver serial number


ozym (approved):
looks okay,

I think the -19 model (or similar) would have been installed prior to this but it may need more details (e.g. ADAM digitiser etc).

elidana:
thanks @ozym ! ok will keep that in mind when we get there

PR #1685, opened 2023-05-05
add EYWM geomag recorder

elidana:
the GSM recorder was missing for EYWM.
Assuming here only one installation since 2013. Using model and serial number as photo sent on 2023/05/01 by geomagnetic obs observer.

image


elidana:
@ozym can I bother you with a thumbs up for this one please? Thanks!!

ozym (approved):

PR #1690, opened 2023-05-12
GBIH datalogger swap

mnaguit:
- from Q330/3 (946) to Q330S/6 (6895)
- for LC 41 to 43, with `200hz HDH` streams

Hi @CBurton90 @salichon 
I noticed there are `1hz (LDH)` & `10 hz (BDH)` streams for this station. Shall we add that to delta?


CBurton90:
If they were not in delta before then I would say leave them out @mnaguit

@gva6500 Applied config to this q330 based on the old q330 which must of had LDH/BDH streams but unsure why they were not collected/metadata recorded further downstream...

@ozym any comments 😆 ?

ozym (approved):

ozym:
It was probably configured like a coastal gauge logger, but was only useful at 200 Hz (or higher!) so I'd ignore these channels.

CBurton90 (approved):

PR #1691, opened 2023-05-13
add changed option to stationxml-build

ozym:
A small update to the stationxml builder application. This adds a flag "changed" which when set will only update the output stationxml file if the contents change other than the timestamp.

The stationxml schema requires a "Created" timestamp and is needed to pass the StationXML validators. However, as part of the CI/CD processes this file will be updated each time an update is run, with a new creation time but most likely otherwise the same contents.   This interferes with the processes used to detect changes in the network and adds extra processing and validation.

This PR allows a given output file to only be updated when the contents change, this is similar to how the single files are built as well.

### DIFF
--- a/cmd/stationxml-build/main.go
+++ b/cmd/stationxml-build/main.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"flag"
 	"fmt"
+	"io/fs"
 	"log"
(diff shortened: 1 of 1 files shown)


junghao (approved):
Looks good.

PR #1696, opened 2023-05-16
PNUZ new RCET station

mnaguit:
- TC120PH-2 (SN 3816) for LC 10 (LH/HH-ENZ streams)
- Fortis (TF1126) for LC 20 (HN-ENZ streams)
- Centaur firmware v4.5.20

Is the network code `RA`? @CBurton90 @salichon


CBurton90:
@mnaguit Is it worth including @staylorofford in the reviews for the RCET sites? I know the majority should have been installed in a shallow posthole vault so we might need some consistency in regards to depth (but perhaps `0` is fine and that's ok with me 😄 )

CBurton90 (dismissed):

mnaguit:
Hi @staylorofford . We will include you in all RCET delta reviews, so you can have a good look of the delta nitty-gritty.

staylorofford:
Hold fire on merging this - I will check the depth with Brendong

CBurton90:
@staylorofford Was there a standardize depth planned for all RCET sites or were they designed to vary slightly location dependent (or how the contractor actually built the vault)? Just wondering if there is a "blanket" value we can put in for all the sites so it represents them nicely.

staylorofford:
SM sensor is at 0 m (loc 20) while WM sensor is at 1.2 m (loc 10)

staylorofford (changes requested):
See above

staylorofford:
@CBurton90, all weak motion sensors should be at 1.2 m install depth, all strong motion will be at 0 m. The contractors drilled to specification 1.2 m on all sites.

mnaguit:
Sensor depth for TC120PH-2 updated to 1.2m

staylorofford (approved):
I approve these depths.

install/connections.csv
salichon, line 1: I reckon it would be better to  set the connection numbers when using  the dataloggers with multichannels now 
 to me this can be good practice
(but need to remember .. :) ) 
 so far i don think there are any impacts though  for this mount

salichon (changes requested):

install/connections.csv
mnaguit: entry corrected

mnaguit:
Waiting for 3 👍 on this one, before I issue another PR for the other RCET (KBRZ). FYI Gentlemen!

CBurton90 (approved):

salichon (approved):
cool  as ;)

PR #1697, opened 2023-05-16
NOVZ sensor swap

mnaguit:
- sensor swap (like-for-like)
- from Lennartz MKII (L-1019) to Lennartz MKII (F-0374)
- for LC 10 (EH-ENZ) streams


CBurton90 (approved):

salichon (approved):
like4like sensor

PR #1701, opened 2023-05-21
add Titan N1 to components list

ozym:
Testing indicated that the site CBGS wasn't building stationxml files using the latest code, turns out the `Titan TACCL-N1` was missing an entry in the `components.csv` file.


elidana (approved):
ouch, good catch, thanks @ozym !

mnaguit (approved):

salichon:
@CBurton90 that s somnehting we d need to think about each time we get a new instruments (ooch :) , woudln't we? 
It does not look allways the same - so that we can inform the pins 0,1,2 - cheerios

CBurton90 (approved):

PR #1704, opened 2023-05-22
use Guralp as manufacturer (Make)

elidana:
This is part of the work to update the codes and metadata to generate stationXML.
A first example is here (for one manufacturer) to use the "Make" column to match the _Manufacturer_ element in the _stationXML/Equipment_  field.

hope I covered all the parts where this is needed, but please let me know if I've missed something!


ozym (approved):
Very nice.

elidana:
cool! I'll keep going then! `Make` is the easy one....thanks for the quick review :)

PR #1705, opened 2023-05-22
update manufacturer (Make) from CSI to Cant Seism Instr

elidana:
continuing the work related to `Make` delta info to be used as StationXML manufacturer name (see also https://github.com/GeoNet/delta/pull/1704)

and /cc @mnaguit


salichon (approved):
All good ... 
OMG!  i hope it s some `sed -i` magic ;) 
Cheers

elidana:
of course it is!! :D

elidana:
will wait for @ozym as well on this specific one, as I'm not sure he thought to go to full name rather than CSI only

ozym (commented):
I think the company make is okay, but the equipment model, should probably be removed, or at least thought about.

i.e the model should be CUSP3A   (rather than CSI CUSP3A, or Canterbury Seismic Instruments 
CUSP3A).

ozym (approved):
I was looking in pod.go which is no longer being used!.

elidana:
so do you think is better we use "CSI" instead of "Canterbury Seismic ..." as `Make`? 
happy to apply this change if CSI that's the preferred manufacturer name

elidana:
ah no nevermind, you approved! cool, will merge this one!

PR #1706, opened 2023-05-22
update manufacturer (Make) General Electric

elidana:
another manufacturer update

@ozym , not 100% sure what "GESystems" was used for and if the update here is correct


ozym (commented):
again, maybe the Sensor needs to remove Druck now (as it's in the Make name now).

elidana:
thanks @ozym ! will update sensor name accordingly

CBurton90:
@elidana @ozym What about just:

`General Electric,Druck PTX-1830,10236617,,`

elidana:
please just suggest a preferred name and I'll update accordingly!

ozym:
Only comment is around the use of Description, i.e. it will be made up of Maker + Model so should go together without duplicate parts (e.g.   GE Druck  Druck PX-...).

elidana:
cool, that helps. didn't realised that Description was going to be composed by Make+Model.

Will go with @CBurton90 suggestion then...update in the making :)

elidana:
done. @ozym or @CBurton90 , can I have another (hopefully short) round of review on this one please?

CBurton90 (approved):

ozym (approved):

PR #1707, opened 2023-05-22
ROTS recorder swap

mnaguit:
- from CUSP3D (42006) to EQR120 (46503)
- for LC 20 (HN-12Z) streams

Hi @CBurton90 , is the firmware version for EQR120 v2.2.2? Thanks!


Thomas-Benson (dismissed):
Looks good to me

CBurton90:
@mnaguit yes 2.2.2

CBurton90 (dismissed):

mnaguit:
Hi @CBurton90 @Thomas-Benson 
Travis error fixed. Another thumbs up please? Thanks!

CBurton90 (dismissed):

install/streams.csv
salichon, line 1: ROTS is no more triggered since 202.040  (february  2020) 
it s a bit  better for consitsency  withthe other ones (information)

install/recorders.csv
salichon, line 1: Out of precaution and since the azimuth is 90 and the SM boxes are different
would you check that the orientation has been kept the same and not shifted  90deg between components. 
Thanks

salichon (changes requested):
just a bit opf polishing :)

install/streams.csv
mnaguit: Yes that is right. But because of the 20 hour gap from this recent recorder swap, I opted to use this gap in updating the entries. Nevertheless, if consistency is the key, will use the 'standard' 200hz end-start.

install/recorders.csv
mnaguit: Hi @CBurton90 . Can you please double checking the azimuth? Thanks!

CBurton90 (approved):
Approving in anticipation azimuth is the same (will wait for Cameron's comment in the ticket).

salichon (approved):

PR #1712, opened 2023-05-23
revert manufacturer (Make) name of GNS to remove macron

elidana:
the macron in "Te Pū Ao" might not be fully compatible with automations, so moving the name of GNS back to "GNS Science" only


CBurton90 (approved):

elidana:
see also https://github.com/GeoNet/delta/pull/1704

PR #1722, opened 2023-05-28
update manufacturer (Make) Sercel

elidana:
continuation of work from #1704 
also removing a couple of entries in assets (Galitzin model), that I don't see used in any installation.
please check if that's ok!


CBurton90 (dismissed):

ozym (changes requested):
Could we add back the Galitzin,  we don't know the serial number, but it was installed at WEL prior to geonet and will have been used to generate picks/magnitudes etc.

This is a tricky one as it relates to historical data.

elidana:
Galitzin readded, thanks @ozym !

ozym (approved):

salichon:
all good

PR #1725, opened 2023-05-29
update stationxml configuring of equipment details

ozym:
This update adds in the Equipment fields for the stationxml encoding. It's using a Description method to generate the name (based on the Make, Model, and Type).

### DIFF
--- a/cmd/stationxml-build/main.go
+++ b/cmd/stationxml-build/main.go
@@ -319,8 +319,8 @@ func main() {
 
 								Datalogger: stationxml.Equipment{
 									Type:             collection.Channel.Type,
-									Description:      strings.Fields(collection.Channel.Model)[0],
-									Manufacturer:     strings.Fields(collection.Channel.Make)[0],
+									Description:      collection.Channel.Description(),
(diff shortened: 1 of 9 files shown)


wilsonjord:
```golang
return fmt.Sprintf("%s %s %s", c.Make, strings.Split(strings.Fields(c.Model)[0], "/")[0], c.Type)
```
There are some new functions that generate a description...were they used to generate the `auto.go` changes here? https://github.com/GeoNet/ac-tools/pull/271/files#diff-33771997d625ef056361b7d7478b1e56af26cc6588ab844b816cd9e3855d3f28
I notice some decriptions have leading spaces, presumably because `c.Make` is empty...this may be known and acceptable behaviour, just wanted to double check.

ozym:
This is part of a transition so the original description was taken from a yaml file, e.g.:

https://github.com/GeoNet/delta/blob/main/resp/responses/sensors/gns.yaml

Which does not match the new mechanism where this is generated from the elements, which is part of this change.  Eventually the auto.go file will be removed.

wilsonjord (approved):

PR #1726, opened 2023-05-29
updated test files to use Set

ozym:
The final update to the test code, switch to a universal formatted call using the `*meta.Set` construct.


wilsonjord:
`go test` works. Notice `loadListFile` has be refactored out...need README update?

wilsonjord (approved):
LGTM, thanks

PR #1729, opened 2023-05-31
WCZ datalogger swap

Thomas-Benson:
- From Q330/3 (6071) to Centaur (10234)
- for LC 10 (HH-ENZ & LH-ENZ) streams
- Centaur firmware v4.8.5


Thomas-Benson:
@CBurton90 Double checking that this Centaur was deployed with the new 4.8.5 firmware version?

salichon (approved):
waitng for the conrad  :)

CBurton90 (approved):
Yes `4.8.5` is correct, thanks @Thomas-Benson

PR #1733, opened 2023-06-02
Update dataloggers.csv for ARHZ

mnaguit:
modify recent entries


salichon (approved):
oui oui  have a good ouikend!

PR #1739, opened 2023-06-06
update manufacturer (Make) PEL

elidana:
continue work from #1704 

This should be the last manufacturer name that is used for StationXML generation

Have found this reference to update manufacturer name to something that end users can more easily find references to.
https://authors.library.caltech.edu/33738/1/12%283%290256.pdf

/cc @mnaguit


salichon (approved):
strong motion inst  from NZ - are these - okay!
thanks all good.

PR #1741, opened 2023-06-07
update model name for Guralp group

elidana:
update model name for Guralp sensors.

- CMG-3ESP-Z to CMG-3ESP/Z
- CMG-3TB-GN to CMG-3TB/GN
- CMG-40T-30S to CMG-40T/30S
- CMG-40T-60S to CMG-40T/60S 

Should now align  with models listed in the IRIS Nominal Response Library https://ds.iris.edu/ds/nrl/sensor/guralp/


salichon (approved):

elidana:
on hold, waiting for backend system to be updated first so that downstream processes are properly handled

elidana:
Delta meeting:
check if response files of, say, Basalt and Basalt/8 , are the same or not.
If the same, use `Basalt` instead of `Basalt/8`

ozym:
@elidana this PR hasn't aged well.  There are a lot of changes to files that have been removed, I'll have another go but probably not a single big bang change.  I'll try and pick out the changes.

elidana:
hei @ozym , no please ignore this one. I've restarted it once back on a plane, but haven't even pushed it. I'm keeping this mostly for the list of models to change and the new names, but is just easier to restart from scratch (I did that for another one). sorry as I've started a couple of months ago and never got back into it

elidana:
also, I think I got distracted by the fact that some Basalt response files are the same, so I was thinking how to handle that, rename so that all map into the same one (instead of having copies with different names)

elidana:
Guralp,CMG-3ESP-Z and Guralp,CMG-3TB-GN done in https://github.com/GeoNet/delta/pull/2457/

elidana:
Guralp,CMG-40T/30S and Guralp,CMG-40T/60S done in https://github.com/GeoNet/delta/pull/2459

PR #1746, opened 2023-06-19
geomag: update EYWM coordinates

elidana:


ozym (approved):
punched in location into google -43.474,172.393 and it looks good (within limits of maps etc).

PR #1747, opened 2023-06-19
geomag: add SMHS partial metadata

elidana:
this is adding some initial metadata for the geomag sensor connected to SMHS.

There are a few things we need to clarify before this can be merged.

Test is also failing, pending on those decisions. Remaining tasks and questions are detailed below

---
- [x] this is adding a new geomag sensor. How should we handle its response? 
- [x] based on that, we need to update relevant info in the following `install` files: `channels.csv, components.csv, calibrations.csv, install/gains.csv`
- [x] how do we treat the "datalogger" here? based on my current understanding, the datalogger is the same used for the strong motion station SMHS, a `FBA-ES-T-BASALT`. This is modelled in delta as a _recorder_, for SM station. Whereas for the geomag data it _should_ (??) be treated as a datalogger. 
   - [x] How shall we reconcile this? Can a possible option add a specific "geomat" role this one, in `install/dataloggers.csv`?
   - [x] or shall we model it, for this use case, as a Basalt+Mag670 recorder? 

this pr is failing the automatic test as well, (possibly related to all missing info?) details below:

```
+ go test ./tests
--- FAIL: TestDataloggers (0.49s)
    --- FAIL: TestDataloggers/check_for_missing_datalogger_assets (0.49s)
        dataloggers_test.go:89: unable to find datalogger asset: FBA-ES-T-BASALT [1196]
        dataloggers_test.go:89: unable to find datalogger asset: FBA-ES-T-BASALT [1275]
        dataloggers_test.go:89: unable to find datalogger asset: FBA-ES-T-BASALT [1677]
        dataloggers_test.go:89: unable to find datalogger asset: FBA-ES-T-BASALT [556]
FAIL
FAIL	github.com/GeoNet/delta/tests	3.926s
FAIL
++ error_handler
++ echo 'Trapped error - Unknown Error'
Trapped error - Unknown Error
++ ((  errcount++  ))
```


install/dataloggers.csv
```diff
@@ -406,6 +406,10 @@ Kinemetrics,BASALT 8X DATALOGGER,1626,Richmond Hill Borehole,,2013-01-28T00:00:0
 Kinemetrics,BASALT 8X DATALOGGER,1626,Redcliffs Borehole,,2013-09-03T00:00:01Z,9999-01-01T00:00:00Z
 Kinemetrics,BASALT 8X DATALOGGER,1627,Redcliffs Borehole,,2013-01-28T00:00:01Z,2013-09-03T00:00:00Z
 Kinemetrics,BASALT 8X DATALOGGER,1627,Richmond Hill Borehole,,2014-07-22T03:00:01Z,2014-10-30T00:30:00Z
+Kinemetrics,FBA-ES-T-BASALT,1196,Summerhill,,2013-04-18T00:10:00Z,2014-10-14T00:00:00Z
+Kinemetrics,FBA-ES-T-BASALT,1275,Summerhill,,2014-10-14T00:00:05Z,9999-01-01T00:00:00Z
+Kinemetrics,FBA-ES-T-BASALT,1677,Summerhill,,2013-01-16T00:30:00Z,2013-04-18T00:00:00Z
+Kinemetrics,FBA-ES-T-BASALT,556,Summerhill,,2012-10-16T04:30:00Z,2013-01-15T11:00:00Z
 Kinemetrics,Obsidian 4X Datalogger,139,Deep Cove,Strong Motion,2015-10-22T05:00:01Z,2020-06-23T00:48:00Z
```
CBurton90, line 413: This is a bit tricky. From what I understand the "datalogger" is actually a strong motion recorder `Kinemetrics | BASALT` acting as a datalogger whereby we are using the spare 4th channel for geomag (the first 3 channels are for strong motion data).

So we probably need to add `Kinemetrics | BASALT` as a datalogger asset in delta that also matches the datalogger install entries above. But these assets will already exist in recorders (as that's what they are and are also being used for) so I'm not sure if this will create a Travis CI nightmare with the same serial numbers used across multiple asset types...

cc @ozym

ozym: Keep the assets as is, install recorder, and extra sensor. The add a connection, and an offset to the 4th channel.

It will probably break but that would be the starting point to avoid duplicate kit.

ozym:
@elidana @CBurton90 this PR should help, #1750 after that is merged you need to add somethings:

`channels.csv`
 > Kinemetrics,BASALT,Strong Motion Recorder,,1,datalogger_Kinemetrics_Basalt_26bits_1sps

`components.csv`
> Bartington,Mag670 L-100,Fluxgate Magnetometer,0,,F,0,0,G,,

`connectsions.csv`
> SMHS,50,SMHS,20,3,2012-10-16T04:30:00Z,9999-01-01T00:00:00Z

The response for the Bartington can be appended to the components file once known, also the names/labels can be adjusted as needed.

ozym:
The key bit is that I've added the ability to use the strong motion installation (SMHS/20) as the Place/Role in the connections file.

elidana:
thanks @ozym , finally got some time to do a bit of work on this. Have tried to implement your suggestion but still no luck, I need to look into more detail 

for the `connections.csv`, I've tried your suggestion and the one in the most recent commit. both did not work...

elidana:
Hi @ozym , I have tried to capture your comments, but I can't get this pr to pass the tests.

I have tried different combinations of SMHS,50 entry in the `install/connections.csv` file, but all I've tried so far is failing a test. 

I'm running out of ideas, can you please have a look and point me to what is wrong?

I _think_ the current pr is reflecting your suggestions , but have copied below some of the other trial/errors I've attempted 

Thanks!


```diff
SMHS,20,Summerhill,,0,2012-10-16T04:30:00Z,9999-01-01T00:00:00Z
+SMHS,50,SMHS,20,3,2012-10-16T04:30:00Z,9999-01-01T00:00:00Z
```

```diff
SMHS,20,Summerhill,,0,2012-10-16T04:30:00Z,9999-01-01T00:00:00Z
+SMHS,50,Summerhill,20,3,2012-10-16T04:30:00Z,9999-01-01T00:00:00Z
```

```diff
-SMHS,20,Summerhill,,0,2012-10-16T04:30:00Z,9999-01-01T00:00:00Z
+SMHS,50,SMHS,20,3,2012-10-16T04:30:00Z,9999-01-01T00:00:00Z
```

ozym:
Yup, this is looking a bit trickier than expected. This two problems are, by connecting a sensor to the basalt the basalt no longer looks like a recorder, but a datalogger instead (we may be able to do a hybrid setup).

The second issue is that we haven't setup the response file for 1sps for a basalt (judging by the error message at the end).

These are edge cases on the edge case, so may need some thinking.

elidana:
oh ok that makes sense Mark. Glad I asked as I started going around in circles. This is not elegant at all and possibly the wrong solution (and might need to check with the geomag team), but would using a different side code solve this?

ozym:
First you need to remove the entry for Summerhill in the connections file (only one line needs to be removed).

I'm working on the response now.

elidana:
@ozym done

ozym:
you need to change the name of "datalogger_Kinemetrics_Basalt_26bits_1sps" to "datalogger_Kinemetrics_Basalt_26bits_1s"

ozym:
You also removed the wrong line in the connections file, the one that should have been removed has the place "Summerhill" in it.

elidana:
ouch, thanks @ozym , and apologies for my confusion

done and all test now pass! Can you and @CBurton90 please have a look?

(also, for reference, response added in https://github.com/GeoNet/delta/pull/1803)

install/channels.csv
```diff
@@ -19,6 +19,7 @@ GEM Systems,GSM-90,Proton Magnetometer,,1,
 GEM Systems,GSM-90F1,Proton Magnetometer,,1,
 GNS Science,EARSS/3,Datalogger,,50,datalogger_GNS-Science_Earss_22bits_50sps
 GNS Science,EARSS/3,Datalogger,,100,datalogger_GNS-Science_Earss_22bits_100sps
+Kinemetrics,BASALT,Strong Motion Recorder,,1,datalogger_Kinemetrics_Basalt_26bits_1s
```
elidana, line 22: @ozym shall we change this to be something like
`Kinemetrics,BASALT-1Hz,Geomag datalogger,,1,datalogger_Kinemetrics_Basalt_26bits_1s` ?

ozym: no, it's still a "Strong Motion Recorder" just capable of extra things but so far only geomag uses 1hz.

elidana: cool, thanks for clarifying

ozym:
@elidana could you have another go with this ticket, I've updated the code to fix the BNF issues. Hopefully the only extra channels now for SMHS will be LFF.  Otherwise this looks good.

elidana:
have tried this again, and I can generate a stationxml file for the site.
attached here
[smhs.zip](https://github.com/GeoNet/delta/files/12653955/smhs.zip)

I can't really review the content, but if it looks fine I think we can merge this one?

CBurton90 (approved):
StationXML looks good : )

elidana:
thanks @CBurton90 !! Did you also managed to have a quick look at the metadata? If they looks ok I think we can merge?

junghao:
The station xml has empty Response for `SMHS LFF 50`, which caused the service crash because in text output for channel level, it's printing response data, too.

https://github.com/GeoNet/tickets/issues/14280

Not sure if it's because of incorrect metadata, or it's just empty at this moment.

@elidana

ozym:
note, the response part of the meta data is optional. 

having said that, it should be filled in when it's known, I think this one is still being worked on (geomag sensor).

PR #460, opened 2019-01-09
GNSS marks additional informations (geology and monument desc)

elidana:
Hi @ozym , this pr supersede https://github.com/GeoNet/delta/pull/196.
I'm assigning you for now as I was inspired by your "site classes" pull request, but this will require a wider audience. 

It is a proof of concept for now, aimed at  salvaging useful GNSS mark notes (related to site geology and monument additional descriptions) that were not imported from the previous GeoNet metadata database.

Files need to be reviewed and worked on before merging.

Geology description is based on the IGS Sitelog and GeodesyML definitions, as per (_ftp://igs.org/pub/station/general/blank.log_) and  GeodesyML schema definition for the [monument](https://github.com/GeoscienceAustralia/GeodesyML/blob/master/schemas/geodesyml/0.4/geodeticMonument.xsd) and the [surrounding area](https://github.com/GeoscienceAustralia/GeodesyML/blob/master/schemas/geodesyml/0.4/monumentInfo.xsd)

```
1.   Site Identification of the GNSS Monument

     Site Name                : 
     Four Character ID        : (A4)
     Monument Inscription     : 
     IERS DOMES Number        : (A9)
     CDP Number               : (A4)
     Monument Description     : (PILLAR/BRASS PLATE/STEEL MAST/etc)
       Height of the Monument : (m)
       Monument Foundation    : (STEEL RODS, CONCRETE BLOCK, ROOF, etc)
       Foundation Depth       : (m)
     Marker Description       : (CHISELLED CROSS/DIVOT/BRASS NAIL/etc)
     Date Installed           : (CCYY-MM-DDThh:mmZ)
     Geologic Characteristic  : (BEDROCK/CLAY/CONGLOMERATE/GRAVEL/SAND/etc)
       Bedrock Type           : (IGNEOUS/METAMORPHIC/SEDIMENTARY)
       Bedrock Condition      : (FRESH/JOINTED/WEATHERED)
       Fracture Spacing       : (1-10 cm/11-50 cm/51-200 cm/over 200 cm)
       Fault zones nearby     : (YES/NO/Name of the zone)
         Distance/activity    : (multiple lines)
     Additional Information   : (multiple lines)
```

some more details and reasoning are in the main comment of the closed pull request.

PR #1751, opened 2023-06-20
add miscellanoues notes inherited from old GNSS marks delta DB

elidana:
this one will require a bunch of improvements, but is hopefully a good start to replace https://github.com/GeoNet/delta/pull/460

We might add start/end date if we want, but in this case and for GNSS marks we only have notes that are not time dependent, but more focused on geology background, scope of the site and additional details for the monument

Much more and useful details in the original pull request


elidana:
replaced by https://github.com/GeoNet/delta/pull/2340

PR #1753, opened 2023-06-20
WTAZ sensor-logger swap

mnaguit:
- from Centaur (9289) to Centaur (10125) fw v4.5.8
- from TC 20PH-2(8268) to TC 120PH-2(9366)   ---> is this correct @CBurton90 , thanks!
- WTAZ has these streams
  - for LC 12 (HH/LH-ENZ)
  - for LC 13 (HH/LH-12Z)


install/firmware.csv
Thomas-Benson, line 1: Do you need to close the firmware for the old Centaur?

mnaguit: Good catch! Centaur 9289 firmware now closed! Thanks @Thomas-Benson 👍

salichon (approved):

CBurton90 (approved):

PR #1755, opened 2023-06-20
refactor(headers): add headers to network tables

ozym:
This is a refactor to get all the table code to line up, it basically adds a Header struct to each table and then uses it to encode/decode each row.  I've also standardised the variable naming a little more.  This is a no change update and the tests pass as before.


junghao (approved):
Looks good.

PR #1771, opened 2023-07-05
HAFS Cusp3D to EQR120 SM sensor swap

Thomas-Benson:
- Removed CSI Cusp3D (42152) at location 20
- Replaced with CSI EQR120 (46660) at new location 21
- New orientation N(0) - changed from axial to non-axial
- EQR120 latest firmware 2.3.2
- Streams updated to reflect that site has been continuous since 2020-02-09
- I've left the location the same since I believe that the orientation is the only thing that has changed, please correct me if I am wrong here @CBurton90 .


CBurton90 (approved):
Happy for the coords to remain as is as they reflect the gps antenna's position on the roof rather than the actual recorders position!

mnaguit (commented):
Hi @Thomas-Benson , this one looks good. But can we change the `LC 21` coords to `-43.14891`,`170.5561` just so to distinguish the difference with `LC 20`
Apologies for the fussiness, thanks in advance!

mnaguit (approved):
Thanks @Thomas-Benson 👍

CBurton90 (approved):

PR #1772, opened 2023-07-06
WHSZ Trillium Compact weak-motion sensor swap

Thomas-Benson:
- Swapping Trillium compact 8785 at location 11 with Trillium compact (9313)
- Streams affect HH-ENZ and LH-ENZ at location 11
- Using same depths & orientation as previous sensor


CBurton90 (approved):
Orientation will need an update (I'll do the analysis soon) but happy to merge as is for now!

mnaguit (approved):

PR #1773, opened 2023-07-06
PYZ Guralp CMG-3ESP to Trillium Horizon TH120-2 weak-motion sensor swap

Thomas-Benson:
Removing Guralp CMG-3ESP (T3P68) from location 10
Replacing with Trillium Horizon TH120-2 (2584) at same location
Streams affected HH-ENZ and LH-ENZ
Azimuth remains N (0)
Do I need to add anythng regarding the UWV mode?


mnaguit (commented):
Looks good @Thomas-Benson 
Just in case we want to be proactive, shall we update LC 20 from triggered to continuous?

Thomas-Benson:
@mnaguit I missed that it was still triggered, will update thanks!

mnaguit (approved):
thank you @Thomas-Benson 🎉

CBurton90 (approved):

PR #1778, opened 2023-07-11
MQZ Q330/6 to Centaur datalogger swap

Thomas-Benson:
- Removing Kinemetrics Q330/6 (3386)
- Replacing with Nanometrics Centaur CTR4-6S (9880)
- Latest centaur firmware version 4.8.5
- Streams affected HH-ENZ and LH-ENZ


mnaguit (approved):
Looks like instead of `McQueen's Valley`, the place that matches the datalogger with open `Q330/6` in `install/dataloggers` is Glencoe Station. 😄 I had to check, coz the serial number is not indicated in the ticket?
Anyway, all good! 👍

CBurton90 (approved):

PR #1780, opened 2023-07-12
Chaparral Physics M64Vx2 - add response

CBurton90:
From: 

http://service.iris.edu/irisws/nrl/1/combine?instconfig=sensor_Chaparral_64Vx2_LP50_HF200_SG0.0275_STairPressure&format=stationxml

Related to deployment of new infrasound sensor for the GeoNet infrasound network.
This will replace aging sensors currently deployed (a mixture of absolute barometers (Setras) and low cost infrasound (InfraBSU)). 
Following an internal evaluation the Chaparral M64Vx2 was selected as the most capable infrasound sensor for volcanic monitoring purposes within the GeoNet network in terms of frequency ranges and sensitivity.


rumachan (dismissed):
Happy with that. Not that familiar with response xml files, but is just copied from the online resource. Compared and match. The other two files look good. Am assuming the serial number, etc are correct.

ozym:
Looks pretty good, just a couple of things. We should remove the resourceId entry, this will be overwritten. We also may have a different protocol on the numbering of poles and zeros (i.e. start from 0 or start from 1).

elidana (approved):
looks good! great we can use the NRL for this one :)

CBurton90:
> Looks pretty good, just a couple of things. We should remove the resourceId entry, this will be overwritten. We also may have a different protocol on the numbering of poles and zeros (i.e. start from 0 or start from 1).

@ozym Resource ID removed.

CBurton90:
@ozym Do you mind having a glance at this? From memory it was all in order to be merged a wee while back, I've just tidied up some conflicts.

It is still failing one test above but unsure if that matters?

CBurton90:
Closing, using NRL xml copy directly instead:

https://github.com/GeoNet/delta/pull/2132

PR #1791, opened 2023-07-26
TOUTL replace CR6

rumachan:
Changes to three files due to changing CR6 digitiser at TOUTL


elidana (approved):
this looks good. 
I have a question around confirming timestamps of the swap as the first data in the raw archive start from `2023-07-25T09:00:00Z` with this new datalogger, but this might be related to site being still ofline / data not collected at the time of the swap

PR #1801, opened 2023-08-08
WTMC&PRPC updating streams to continuous

Thomas-Benson:
- Updating streams for WTMC and PRPC to reflect that they have been continuous since `2020-02-09`
- Streams affected WTMC LC20 HN-12Z and PRPC LC20 HN-12Z


mnaguit (approved):
👍

PR #1803, opened 2023-08-09
feat: add response for basalt at 1hz sampling

ozym:
This is to allow collection of 1Hz geomag data via a basalt recorder.


ozym:
As a check here's the NRL reference:

http://service.iris.edu/irisws/nrl/1/combine?instconfig=datalogger_Kinemetrics_Basalt_PG1_FV10Vpp_FR1_FPNon-causal&format=stationxml-resp

CBurton90 (approved):

elidana:
too late but this looks good, thanks @ozym !

PR #1804, opened 2023-08-09
Update classes.csv

mnaguit:
- modify entries for VanHoutte2012a and VanHoutte2012b


environment/classes.csv
```diff
@@ -158,7 +158,7 @@ Station,Site Class,Vs30,Vs30 Quality,Tsite,Tsite Method,Tsite Quality,Basement D
 941A,D,272,Q2,0.85,I,Q2,98.5,Q2,,McVerry2011 Semmens2010,
 ADCS,D,270,Q3,>2,I,Q3,990,Q3,,Lee2015 Perrin2015,
 AICS,D,210,Q3,1,I,Q3,80,Q3,,Perrin2015,
-AKSS,C,435,Q1,0.3,Mw,Q1,10,Q2,,Kaiser2013 VanHoutte2017,
```
ozym, line 161: this appears to be 2017? was that an original mistyping?

mnaguit: We probably did a mismatch for that entry. The reason behind is there is another unpublished reference (Unpublished HVSR which is VanHoutte2016). 
AKSS is linked to Unpublished MASW (VanHoutte) which from our current reference corresponds to `VanHoutte2012b`

image

ozym (approved):

PR #1806, opened 2023-08-13
THQ2 WM sensor and Centaur install

Thomas-Benson:
- Removing Duke Malin Seismometer (601) from THQ2
- Replacing with LE-3DliteMkIII (G-0559) at new site LC11
- Removing Q330/3 (3625) from `Tauhara North`
- Replacing with Centaur (9506) at `Tauhara East`
- connections.csv updated to match change in place name
- streams.csv updated to reflect historical channel naming as axial (12Z)
- connections.csv and streams.csv also updated to reflect data on incorrect station code - to be changed after archive correction.


install/connections.csv
```diff
@@ -845,7 +845,9 @@ TEU,10,Te Uri,Short Period,,1990-02-20T02:23:00Z,1990-05-17T17:11:00Z
 TGRZ,10,Welcome Bay Road,,,2003-02-18T08:00:01Z,9999-01-01T00:00:00Z
 THLX,10,Tinakori Hill,,0,2014-08-25T00:05:00Z,2015-03-24T00:00:00Z
 THLX,20,Tinakori Hill,,3,2014-08-25T00:05:00Z,2015-03-24T00:00:00Z
-THQ2,10,Tauhara North,,,2012-12-18T01:00:00Z,9999-01-01T00:00:00Z
+THQ2,10,Tauhara East,,,2023-08-11T02:44:00Z,2023-08-13T21:14:53Z
```
CBurton90, line 848: I think we don't need this entry as that is what requires a cleanup/rename in the archive? @mnaguit Thoughts?

install/streams.csv
```diff
@@ -2138,7 +2138,9 @@ THOB,22,H,N,200,true,false,false,2011-01-01T00:00:00Z,9999-01-01T00:00:00Z
 THOB,23,B,N,50,true,false,false,2011-01-01T00:00:00Z,2019-12-16T00:00:00Z
 THOB,23,H,N,200,true,false,false,2011-01-01T00:00:00Z,9999-01-01T00:00:00Z
 THPS,20,H,N,200,true,false,true,2001-11-27T00:00:00Z,2005-02-07T00:00:00Z
-THQ2,10,,,100,false,false,false,2012-12-18T01:00:00Z,9999-01-01T00:00:00Z
+THQ2,10,E,H,100,true,false,false,2012-12-18T01:00:00Z,2023-08-07T23:26:00Z
+THQ2,10,E,H,100,false,false,false,2023-08-11T02:44:00Z,2023-08-13T21:14:53Z
```
CBurton90, line 2142: Same with this one. Dates should be merged into the latest install with archive cleanup/renaming...

mnaguit:
Is there a change in place name or we are sticking with Tauhara North?

install/connections.csv
mnaguit: We are closing LC 10? if yes, then we need a new entry for LC 11?

CBurton90:
@Thomas-Benson @mnaguit It should be now called Tauhara East (same site though).

mnaguit:
I'll post my thoughts on this in the main ticket. FYI.

install/connections.csv
Thomas-Benson: Extra entry removed

CBurton90 (approved):

mnaguit (approved):
👍

PR #1808, opened 2023-08-14
fix: bug with testing responses and do a double encode check

ozym:
Caused by the Parallel call, this then only checked the last response file.

The auto responses are generated using a different technique so will not match, unless the same mechanism is used. This will likely be fixed


jajera (approved):
lgtm

PR #1811, opened 2023-08-17
fix: update build workflow

ozym:
Just removing the dispatch for now.


BobyMCbobs (commented):
To not have it be callable manually?

BobyMCbobs (approved):
make it run workflows!

PR #1812, opened 2023-08-17
Migrating sites from streams.csv to polarities.csv

Thomas-Benson:
Found six sites marked as reversed in `install/streams.csv`:
```
BFZ | 10 | 2003-07-30T05:00:00Z | 2008-11-09T22:13:00Z
BKZ | 10 | 2004-02-11T02:00:00Z | 2013-10-09T00:00:00Z
HIZ | 10 | 2004-02-03T02:00:00Z | 2014-01-30T01:00:00Z
KHZ | 10 | 2003-08-06T00:00:00Z | 2011-02-23T04:00:00Z
ODZ | 10 | 2003-09-11T06:00:00Z | 2008-07-08T00:00:00Z
OUZ | 10 | 2003-08-26T05:00:00Z | 2010-02-24T00:00:00Z
```
Transfered into `install/polarities.csv`
Assuming Primary = true since every other entry uses it?


elidana:
thanks @Thomas-Benson ! 
the changes looks pretty good to me, but I think the code might need some changes? 
(@ozym to confirm)

I did the following:
- downloaded your branch/pull request, built a stationXML file for BFZ
- used the main GeoNet branch, built a stationXML file for BFZ

the dip and azimuth info do not match, i.e.
```
./stationxml-build -station BFZ -version 1.0 -location 10 -base /home/elidana/git/delta/ -output main.xml
./stationxml-build -station BFZ -version 1.0 -location 10 -base /home/tmp/delta/ -output thomas.xml
```

here's some examples of differences that I've found:
image
image
image

Thomas-Benson:
@elidana I had a look at that, and it seem like the reversal flag in `streams.csv` is not implemented because I do not see a difference when building the stationxml with only the modifications to `streams.csv`. Only reversals listed in `polarities.csv` appear to have any effect. (if you want to test, you can use this branch: https://github.com/Thomas-Benson/delta/tree/Polarities-only-modified-streams)

If this is right, the difference we are seeing is the change in polarity being correctly applied to the stations now that they are listed in `polarities.csv`

elidana:
oh, nice, good call to also test by only applying the change to `streams.csv`.
I also see that the current FDSN service is providing the info that match the status prior to your pull request, i.e.
```
[elidana@hutl24201 install]$ curl 'https://service.geonet.org.nz/fdsnws/station/1/query?network=NZ&station=BFZ&level=channel&format=text' | grep HH | grep 2003

#Network | Station | Location | Channel | Latitude | Longitude | Elevation | Depth | Azimuth | Dip | SensorDescription | Scale | ScaleFreq | ScaleUnits | SampleRate | StartTime | EndTime
NZ|BFZ|10|HHE|-40.679647|176.246245|283.000000|0.000000|90.000000|0.000000|Broadband Seismometer|629145600.000000|1.000000|m/s|100.000000|2003-07-30T05:00:00|2007-11-01T00:00:00
NZ|BFZ|10|HHN|-40.679647|176.246245|283.000000|0.000000|0.000000|0.000000|Broadband Seismometer|629145600.000000|1.000000|m/s|100.000000|2003-07-30T05:00:00|2007-11-01T00:00:00
NZ|BFZ|10|HHZ|-40.679647|176.246245|283.000000|0.000000|0.000000|-90.000000|Broadband Seismometer|629145600.000000|1.000000|m/s|100.000000|2003-07-30T05:00:00|2007-11-01T00:00:00
```

I have another question now. 
By looking at differences, it seems to me that the reversal is applied to all 3 components now (i.e. when data are moved to `polarities.csv`). Is that what we want? 

For BFZ, I was assuming that we only wanted to revers the vertical component. I'm not sure that's what is expected/reflecting what we know 

based on stationXML definition for `Dip`
> Dip of the component in degrees, positive is down from horizontal. For horizontal dip=0, for vertical upwards dip=-90
and for vertical downwards dip=+90.

and if I get it right, if `dip=-90` , for HHZ component , then the component is marked as reversed?

Or am I getting this totally wrong?

salichon:
Hi there, 
As pickedup by @elidana 
This may  need some clarification perhaps to me 
- _Polarities.csv_ "polarities" is for the (almost always) Z component that is for some reason inverted (Z sign changes) - other components being correct
- in _streams.csv_  "reversed" is for the entire  3 components to be inverted. (all signs are changed) 
The migration therefore does not look straight forward as this 

As far as i remember,  the polarity correction for the Z component can be informed by the metadata in addition to the cartesian system orientation, though the Z sign correction itself is generally  operated by the user.

elidana:
thanks @salichon ! whatever we end up with, I think might be better if we only use one table for the reversal information (i.e. move the info out of streams). Perhaps we can add a new column in `polarities.csv` to indicate the vertical reversal only, or one field per component? also not 100% sure how streams.csv is currently used, might be worth a double check from expert eyes

PR #1813, opened 2023-08-18
GEOMAG: APIM datalogger(s) swap

elidana:
This is capturing a fluxgate datalogger change at Apia Geomagnetic Observatory.

Previous installation was a datalogger with 6 channels.
This has been replaced by two dataloggers with 3 channels each.

Some notes:
- The current pr is only providing info for one of them, connected to the fuxgate magnetometer components.
- The fluxgate temperature components are connected to a different datalogger (`Q330/3 sn 3614`) that I currently don't know how to correctly represent in delta. Any suggestion appreciated! 
- I am assuming here that there is no change in the fluxgate specific calibration to align to the magnetic north, so I have not included any change in the `gains.csv` file.


CBurton90 (dismissed):
I think this is fine to merge then we can tackle the second datalogger next week?

CBurton90 (approved):

mnaguit (approved):
I don't know much about Geomag, but since you requested a reviewer @elidana 😁 , this looks fine (like the datalogger has an entry in assets, the Apia Geomag Obs has entry (and open) in stations and in connections. Not sure if there should be sensor/s connected to the logger? Other than that, LGTM

mnaguit:
There was a sensor connected! (`Danish Meteorological Institute,FGE,S0330-E0397`) 😄

elidana:
awesome, thanks a lot @mnaguit !! merging!

PR #1817, opened 2023-08-22
fix: update the way the stationxml networks get their start/end times

ozym:
This is an attempt to be more consistent with how the start/end dates for the networks are managed, previously each network was based on the times of the stations included. This worked except for when single station xml was being generated and each xml had a different network code based on the single station.  This method finds the times prior to building the xml, it also only uses stations which have had a sensor installed at some time in their past

### DIFF
--- a/cmd/stationxml-build/main.go
+++ b/cmd/stationxml-build/main.go
@@ -218,6 +218,44 @@ func main() {
 		exts[n.External] = append(exts[n.External], n.Code)
 	}
 
+	// keep the external network operational times constant even if not all stations are used.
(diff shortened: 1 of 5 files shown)


ozym:
A run via this code and the existing code gives:

```diff
6c6
<   <Network code="NZ" startDate="1884-02-01T00:00:00" restrictedStatus="open">
---
>   <Network code="NZ" startDate="1916-01-01T00:00:00" restrictedStatus="open">
```

jajera (approved):
lgtm

PR #1818, opened 2023-08-23
feat: extend matcher to negative checks

ozym:
This will allow matching the current checks which use negative features (i.e. not these networks etc.)


jajera (approved):
lgtm

PR #1819, opened 2023-08-23
feat: switch to using a settings structure

ozym:
This is a no-op change, it simply moves the configuration into a struct


jajera (approved):
lgtm

PR #1820, opened 2023-08-23
feat: add operational checks

ozym:
For building single files there are often extra criteria needed to decide which stations and channels to include based on start/end times.


jajera (approved):
lgtm

PR #1822, opened 2023-08-24
fix: update stationxml single file build parameters

ozym:
This is used for building seiscomp3 config files in single mode, needed to match the existing output


jajera (approved):
lgtm

PR #1823, opened 2023-08-24
fix: update ABAZ streams

ozym:
The sensor was swapped from a short period (EH?) to broadband (HH?) so the streams listing needs updating


Thomas-Benson (approved):
LGTM, but noting that the site is still streaming EH?

ozym:
@CBurton90 see above, does any of this make sense?

ozym:
@Thomas-Benson @elidana has been confirmed that the sensor is broadband, but the channels weren't renamed. Once merged, this will need to be added to the "fixup" ticket to rename the channels that have been created since the sensor swap.

moorematthew:
@ozym @Thomas-Benson Also needs a location code update. Sensor has moved from vault to direct bury. 
Apologies this was missed on install

moorematthew:
Update from centaur 

Capture

ozym:
@Thomas-Benson could you check the newly installed sensor location (lives in sites.csv under  ABAZ/11) and update if needed.

Thomas-Benson (approved):
lgtm

PR #1824, opened 2023-08-25
feat: add DART persistent identifiers for WMO bulletins

elidana:
Hi @ozym (and /cc @JonoHanson )

This is a mock up to add WMO Bulletin headers and persistent DART stations identifiers to delta.

I have also added start and end times to cover the use case when a DART station is redeployed at a location for which its previous WMO identifier is no longer valid.  
(based on my understanding of https://www.ndbc.noaa.gov/faq/staid.shtml)
Would such a use case imply we change the station code, or would we use the same?
(E.g if we redeploy site NZA to a location so different that it's bulletin header will change, will we still be calling it NZA?)

Not sure if it is covering all that is needed, but hopefully is a start?


ozym (changes requested):
Looking good, only minor changes

- remove the SZ from the start of the WMO Identifiers (this just means "Sea Level" data).
- remove NZKL from the WMO identifiers, this is a GTS code for the organisation that puts the message onto the system rather than related to the site (this code means the Metservice at Kelburn).

elidana:
Hi @ozym ,
changes now implemented, can you please do another review?
And feel free to merge away and delete my branch if it looks good!

ozym (approved):
we can add more entries as needed.

PR #1828, opened 2023-08-31
feat: expand dart table to be a replacement for dart values in stations

ozym:
Added in the code to wrap around the newly added dart file,  in the scheme of things this can be treated much like the monuments file.


elidana (commented):
Hi @ozym , 
this looks good but not sure if we need to add some flexibility to the code to allow multiple entries for different time windows in `network/darts.csv`?
(for the use case of a relocated dart with different WMO identifier...if that's a possibility, which I'm still not sure about)

I've tried to add this and the test will fail
```
[elidana@fedora delta]$ git status
On branch feature-expanded-dart
```

```diff
diff --git a/network/darts.csv b/network/darts.csv
index 82ee0c70..ccbffa03 100644
--- a/network/darts.csv
+++ b/network/darts.csv
@@ -1,5 +1,6 @@
 Station,Pid,WMO Identifier,Start Date,End Date
-NZA,5501002,NZ41,2019-12-21T17:00:00Z,9999-01-01T00:00:00Z
+NZA,5501002,NZ41,2019-12-21T17:00:00Z,2023-04-20T23:59:59Z
+NZA,5501002,NZ41,2023-04-21T00:00:00Z,9999-01-01T00:00:00Z
 NZB,5501003,NZ42,2020-09-18T00:00:00Z,9999-01-01T00:00:00Z
 NZC,5401000,NZ43,2019-12-12T18:00:00Z,9999-01-01T00:00:00Z
 NZD,5501004,NZ44,2021-07-23T00:00:00Z,9999-01-01T00:00:00Z
```

and when running the test

```
ok  	github.com/GeoNet/delta/tides	(cached)
+ go test ./tests
--- FAIL: TestDarts (0.00s)
    --- FAIL: TestDarts/check_for_duplicated_darts (0.00s)
        darts_test.go:21: dart duplication: NZA
FAIL
FAIL	github.com/GeoNet/delta/tests	3.832s
FAIL
++ error_handler
++ echo 'Trapped error - Unknown Error'
Trapped error - Unknown Error
++ ((  errcount++  ))
```

ozym:
Think of these as GNSS monuments, they won't change unless there is a complete rename which won't happen.

elidana:
ah ok, I thought that the "NZx" code would have been retained for a wide geographic area.
thanks for clarifying that's not the case

elidana (approved):

PR #1831, opened 2023-09-03
chore: remove old tools

ozym:
This includes tools/stationxml which has been replaced with cmd/stationxml-build, and pod which is no longer needed.


jajera (approved):
lgtm

PR #1833, opened 2023-09-04
feat: update bnc and ntripcaster config apps

ozym:
These have been updated to use an internal module which loads the config files and allows for better testing


jajera (approved):
lgtm

PR #1834, opened 2023-09-04
feat: update actions tests order

ozym:
This swaps the testing actions to avoid overloaded testing debug, first check delta, then check the apps. This is to avoid confusing users with application code rather than delta file problems


jajera (approved):
lgtm

PR #1839, opened 2023-09-07
manualcollect needed for Tilde

rumachan:
As per https://github.com/GeoNet/tickets/issues/14165


elidana (approved):
Hi @rumachan , this one looks good, though I'd like at some point to resume https://github.com/GeoNet/delta/pull/1558 and separate out the manually collected data from `network/sites.csv`

also, as a general comment:
I see that there is a huge diversity of  _XXelement-conc_ in the `property` column.
Would it be possible to separate out in two different columns the XXelement from the "type of observation", and do something like:
- `property` column = `concentration`
- `aspect` column = `XXelement`
?

We have a further column that can be leveraged and might minimize the number of items users will see from the "Name" key in tilde

rumachan:
Thanks @elidana. Some good points there.

Yes, its possible to separate but doesn't that just move the problem from A to B?
If I understand correctly we'd have:
Name = spring-conc, spring-temperature, etc
Aspect = Al, As, B, Br, etc if name is spring-conc
Fewer Names but more Aspects.

The plan was to have Aspect used for "duplicate samples" in this case. If we made these changes then Aspect would become very important, which isn't what I'd thought we intended.

If we can deal with the display issue in Tilde DD GUI, both in the selection dropdown and the map popup, then the problem you rightly identify would have less impact. Am going to note that we are now ready to deal with that issue, when I find the relevant ticket.

sue-h-gns:
@rumachan is this done?

PR #1841, opened 2023-09-09
fix: required to split out common ntrip files

ozym:
The config files may be split into two directories, common and otherwise. This adds the xtra requirements to find these files. It also updates the code into the settings structure.


jajera (approved):
lgtm

PR #1846, opened 2023-09-11
fix: added serial number of bpr sensor to ndbc-update

ozym:
Pulling in some extra details into the format to handle the BPR sensor serial number.


jajera (approved):
lgtm

PR #1848, opened 2023-09-12
GLKZ added Broadband role to past dataloggers and connections

Thomas-Benson:
Adding `Broadband` roles to GLKZ connections and dataloggers to fix missing elements in stationxml
 - The `Broadband` role was added to one datalogger & connection when a second datalogger was installed at the same place, but due to overlapping sensors and dataloggers, this role need to be applied to all linked entries to avoid losing channel from the stationxml
Solution to help ticket: https://github.com/GeoNet/help/issues/110


mnaguit (approved):

salichon (commented):
@Thomas-Benson @mnaguit 
I m not quite sure about the relevancy of the Role  as "broadband"
to connections that are 40/80/and 81
"Broadband" for broadband seismometer
"short period" etc....
Hence does not sound right to me 
Should the Role be removed from the change instead ? 
Does it work ? 
thnaks

Thomas-Benson:
@salichon I agree that the `broadband` role is possibly not the most apt description. The role needs to be the same for all of locations 10/40/80/81, since these were running off of the same datalogger until a new system was installed on `2023-01-30`. This new system involves a second datalogger with the place name `Raoul Island Green Lake`, necessitating the addition of roles to the two dataloggers and their connections.
I went with the `Broadband` role since it had already been partially implemented when the new datalogger was installed and it reflects the current use of the datalogger, but I'm happy to change it if you think a different role fits better. Current roles in use: https://github.com/GeoNet/tickets/issues/12649#issuecomment-1412919975

mnaguit:
Does this PR help? https://github.com/GeoNet/tickets/issues/12649#issuecomment-1412927057

gva6500:
Delta: @ozym to have a look.

salichon:
Sorry for the late - it looks good to me :)

salichon (approved):

PR #1853, opened 2023-09-15
MNMZ station installation

Thomas-Benson:
- Opening new site MNMZ as replacement for RITZ
- Installing sensor Trillium Compact 120PH-2 (9328) at LC 10, depth of 10m
- Installing datalogger Centaur CTR4-6S (9882) at place Mangamawhitiwhiti
- New streams MNMZ LC 10 HH-ENZ and LH-ENZ
- Using orientation of  0 & non-axial stream names for now, to be updated once orientation is determined
- Using network code of `TP` to be the same as RITZ


mnaguit (approved):
Looks good, @Thomas-Benson thanks!
A few comments though:
- is there a need to insert the Broadband role in `connections`?
- the station/site name is quite long aye? 😄 
- noting RITZ is still operational and azimuth for MNMZ has not been measured

Thomas-Benson:
@mnaguit I didn't add a role since there is only one datalogger at this site currently, although noting that if the recent disucussion in the parent ticket https://github.com/GeoNet/tickets/issues/13940#issuecomment-1722687621 about adding a second sensor+datalogger to calibrate the borehole goes ahead, we will need to add roles to `connections` and `dataloggers`. I believe in that case using the `Borehole` and `Surface` roles would be the most apt.

CBurton90 (approved):

CBurton90:
@Thomas-Benson @mnaguit For the surface sensor to orient the borehole the best practice is to add it on the second bank of the digitiser for ~24hrs and add the channel code of MNMZ.XX.HHZ|N|E. Then we can just grab the data stored locally on the digitiser. The XX channel code should prevent any archiving of the seedlink stream for those channels therefore no need to worry about delta.

PR #1854, opened 2023-09-15
fix: correct the handling of single files

ozym:
An error checking bug meant that files which were the same, other than their creation dates, were being treated as different, this removes that problem and also now checks for an output directory


jajera (approved):
lgtm

PR #1856, opened 2023-09-15
feat: add points to separate manually collected data from sites.csv

elidana:
this replace https://github.com/GeoNet/delta/pull/1558

This pull request is the next step to complete the separation of manually collected data from continuously recording stations.

We have introduced the `network/samples.csv` file and associated checks to list all locations where manually collected data can be sampled.
In each of these locations multiple samples can be collected at slightly different locations. This was previously handled in the common file `network/sites.csv`, that is used for seismic stations and alike to distinguish different continuously recording sites at the same location.
To avoid conflicts and downstream issues, the same concept is here applied to manually collected data.

For that, a new file is created: `network/points.csv` so that we have the following analogy for table names:

continuously recording station | manually collected data station
---|---
stations.csv | samples.csv
sites.csv | points.csv

Once this one is approved, the points information will need to be removed from `network/sites.csv`


elidana:
@ozym this is now updated and hopefully compatible with https://github.com/GeoNet/delta/pull/1859

ozym (approved):

rumachan (commented):
I'm adaptable and for data entry and modification can work with points.csv instead of sites.csv very easily. This all seems very sensible.
Do we have to change the code that does Tilde meta-data to look at points.csv instead of sites.csv, i.e. downstream impact? I'm not sure what code or how it works.

ozym:
Yes, we'll need to update that code, but it will come after this PR is merged.

PR #1858, opened 2023-09-17
chore: swap point with position

ozym:
This is to make way for the future table called point.  It's a straight s/Point/Position/g  s/point/position/g.


jajera (approved):
lgtm

PR #1859, opened 2023-09-18
feat: add points table

ozym:
This is based on sites but has the same role as samples has to stations


bpeng (approved):
Looks good

elidana:
is there any reason why this one can't be merged? (besides updating the branch...)

PR #1878, opened 2023-10-04
add tilde method to references

rumachan:
Explanation of meaning of `method` in Tilde application. There is currently no explanation, and for `manualcollect` its a bit complicated, to say the least.

Tossed up between CSV file and README for this, and went with CSV as its compatible with existing file here, and potentially allows future referencing from Tilde GUI if that is ever required.

Descriptions are those used in FITS, but updated and clarified to be easier to understand and give more information.

Included `envirosensor` as well as `manualcollect` and imagine we can add others if/when we deem it necessary.


elidana (approved):
thanks @rumachan !
This looks very good to me.

AFTER this one has been approved/merged, we can add `dart` and `coastal` related methods
and (before they get lost) those related to GNSS data currently in FITS.

Not suggesting to do that now.

Would be good to also get @ozym 's opinion on this one

ozym (commented):
This looks fine from a content point of view, just wondering whether it needs to be attached to tilde though.  It comes across as a "glossary" of methods which are used in tilde (manual collect) but could also be used in other "domains" not in tilde.

I'd be happy with simply "methods.csv" as the file name, the content of the README and file can stay as is.

elidana:
We were wondering the same, but right now we cannot find anything in the TILDE API that can be used for that.

would be nice to have a new field in the _dataSummary_ endpoint (https://tilde.geonet.org.nz/v3/api-docs/endpoint/dataSummary) for that (something like Method-Description)

rumachan:
I agree with you both. Its very Tilde specific material but with no way to connect it to Tilde. What I hope is that once Elisabetta and others add other methods then we can ask for it to be linked to Tilde somehow.

PR #1879, opened 2023-10-09
update placenames based on latest LINZ NZ Suburbs and Localities

elidana:
This is updating the delta placenames, that are based on the "LINZ NZ Localities", now deprecated and repleaced by the newly released "NZ Suburbs and Localities" dataset.

The new dataset is available on the LINZ Data Service (https://data.linz.govt.nz/layer/113764-nz-suburbs-and-localities/)


ozym (approved):
ok

PR #1882, opened 2023-10-10
WMGZ SP sensor swap

Thomas-Benson:
- Removing LE-3DliteMkII (I-0650) from WMGZ LC 10
- Replacing with LE-3DliteMkII (J-0940) at same location
- Streams affected LC 10 EH-ENZ


CBurton90 (approved):

mnaguit (approved):
Looks good, just a side comment that the serial numbers are not arranged alphabetically (or arranged alphanumeric) in both tables.

PR #1888, opened 2023-10-19
HALS Cusp3D to EQR120 SM Recorder Swap

Thomas-Benson:
Removing Cusp3D (42035) from HALS LC 21
Replacing with EQR120 (46669) at same location
Using latest EQR firmware version 4.3.2
Using same orientation as previous sensor 272
Streams affected HALS LC 21 HN-12Z


CBurton90 (dismissed):

mnaguit (changes requested):
Thanks @Thomas-Benson , a few changes:
- should the firmware for EQR120 start when the recorder starts?
- for the streams, we can update the triggered to continuous streaming for LC 21?

Thomas-Benson:
@mnaguit fixed both the firmware timing and the stream status

CBurton90 (approved):

mnaguit (approved):
Thanks @Thomas-Benson 👍

PR #1889, opened 2023-10-20
STAS Cusp3D to EQR120 SM Recorder Swap

Thomas-Benson:
- Removing Cusp3D (42012) from STAS LC 20
- Replacing with EQR120 (46668) at same location
- Using latest EQR120 firmware version 2.3.2
- Using same orientation as previous recorder, 180
- Updated stream to continuous from usual time of 2020-02-09T00:00:00Z
- Streams affected STAS LC 20 HN-12Z


salichon (approved):
sweet have a great WE !

CBurton90 (approved):

PR #1899, opened 2023-10-31
Lennartz LE-3D/BH(s) MkIII - add response

CBurton90:
Response info retrieved from the ES NRL > https://service.iris.edu/irisws/nrl/1/combine?instconfig=sensor_Lennartz_LE-3D-BH-MkIII_LP1_SG800_STgroundVel&format=stationxml-resp

- (s) refers to stainless steel AFAIA so no need to include in naming
- placeholder has been entered for SN
- Normalization Frequency is 5Hz

Assigning a bunch of people for "enlightenment" of the new response addition process 😆


salichon:
- will locally import and test 
- Generate Response
- Validate fdsnxml rsposne output

CBurton90:
@salichon @elidana @ozym This has now passed (woohoo). The sensor has now been deployed so a review would be great and if you have merge permissions please merge after review.

@salichon You may want to do some local testing by pulling this branch before the review/merge?

elidana (commented):
Thanks @CBurton90 , this looks good but I'll let the expert @salichon to review/approve before merging.

FWIW, I did check how the response file compares to the SAGE Nominal Response Library.

the differences are 
- reference to the schema 
- Instrument Sensitivity Value units
- PolesZeros resourceId

- For the schema, we don't link to the NML in any other response file, so that's ok to keep as in your pr
- Not sure if the difference in unit/format using for Value is significant or not
- same as above for the resourceId 

in the pic below: left = SAGE NML, right = delta pr
image

CBurton90:
@elidana Nice, thanks! For the differences I removed a lot of the extra bits from the SAGE NRL that @ozym has previously said are unneeded/unwanted. For your specific points:

- Schema reference: I removed the NRL one and added our standard one as per the other response files in delta (happy to change this back)
- Instrument sensitivity: this should be fine and is identical (its just 8e+2 vs 800), unless there is a problem with formatting etc. (I think that's what your suggesting?).
- PolesZeros resource ID: @ozym added this but perhaps it should be next to StageNumber rather than PolesZeros?

elidana:
oh sounds good thanks @CBurton90 !!

I should have done before, apologies for that. 
Based on the FDSN specs for stationXML, if I read it correctly (https://docs.fdsn.org/projects/stationxml/en/latest/reference.html#response-instrumentsensitivity-value)
the value as you filled in should be fine

specs states that:
> <Value> (Response → InstrumentSensitivity → Value)
> content type: [double](https://docs.fdsn.org/projects/stationxml/en/latest/appendices.html#glossary-double)
> A scalar value representing the amount of amplification or diminution, if any, the current stage applies to the input signal.

and [double](https://docs.fdsn.org/projects/stationxml/en/latest/appendices.html#glossary-double) type is 
> A number between 2.23 × 10^-308 and 8.98 × 10^307 (rounded), along with positive and negative infinity and NaN.
(with no specific requirements on number format, as far as I understand) 


IMO: happy to keep as is, but I'd still like to have @salichon approval before merging!



Also, FWIW for next time, in the NRL you can also directly download the response snippet (they provide a full one, and then only the response part [StationXML-Response](https://service.iris.edu/irisws/nrl/1/combine?instconfig=sensor_Lennartz_LE-3D-BH-MkIII_LP1_SG800_STgroundVel&format=stationxml-resp), though for some reason if you "curl" that URL, it will download the full stationxml.

image

resp/files/sensor_Lennartz_LE-3D-BH-MkIII.xml
salichon, line 1: Hi @CBurton90  form https://www.lennartz-electronic.de/wp-content/uploads/2021/04/Lennartz-SeismometerManual.pdf 
Is it a LE-3Dlite MKIII cased for boreholes? 
and it says +/- 4.440 , not 4.444 in the ref (but i am being pickish hey ! ...)

salichon (dismissed):

salichon:
@CBurton90  this is the first part prior to `auto.go` and the association to a datalogger  right ?
So far i am good :)

install/components.csv
```diff
@@ -95,6 +95,9 @@ Kinemetrics,FBA-ES-T-ISO,Accelerometer,2,,E,0,90,G,,sensor_Kinemetrics_FBA-ES-T-
 Kinemetrics,SBEPI,Accelerometer,0,,Z,-90,0,G,,sensor_Kinemetrics_SBEPI
 Kinemetrics,SBEPI,Accelerometer,1,,N,0,0,G,,sensor_Kinemetrics_SBEPI
 Kinemetrics,SBEPI,Accelerometer,2,,E,0,90,G,,sensor_Kinemetrics_SBEPI
+Lennartz Electronic,LE-3D-BH-MkIII,Short Period Seismometer,0,,Z,-90,0,G,,sensor_Lennartz_LE-3D-BH-MkIII
+Lennartz Electronic,LE-3D-BH-MkIII,Short Period Seismometer,1,,N,0,0,G,,sensor_Lennartz_LE-3D-BH-MkIII
+Lennartz Electronic,LE-3D-BH-MkIII,Short Period Seismometer,2,,E,90,0,G,,sensor_Lennartz_LE-3D-BH-MkIII
```
Thomas-Benson, line 100: Shouldn't the orientation of the 2/E component here be 0,90 not 90,0?

CBurton90: @Thomas-Benson Yes indeed, good catch! We can blame my fat fingers for that one!

CBurton90: I've now updated it.

CBurton90:
> @CBurton90 this is the first part prior to `auto.go` and the association to a datalogger right ? So far i am good :)

@salichon Yes this is just the sensor portion of the response. @ozym Any indication on how the the "combination" or "connection" should be matched with the correct digitiser/datalogger in the new method? I couldn't re-generate auto.go with the usual go generate (but perhaps my golang version is again behind).

ozym:
auto.go is legacy and not needed for the newer code. The equivalent output is in the auto dir, but this is only from updates to the yaml files.

CBurton90:
@ozym Great! So after the correct install entries are entered into delta (install/dataloggers.csv, install/sensors.csv, install/connections.csv) this should link the two response together? Or is there another step before that?

@salichon If your happy with the PR please fee free to merge when ready!

salichon:
relates to https://github.com/GeoNet/tickets/issues/14574

salichon (approved):

PR #1904, opened 2023-11-05
COVZ TitanSMA SM Recorder Installation

Thomas-Benson:
- Installing TitanSMA (1257) at new site LC 20 for existing station COVZ
- Using most recent TitanSMA firmware version 4.5.22
- New site uses same coordinates as all other sites at this station.


mnaguit (commented):
Hi. Double checking: was the instrument oriented to the north? If that is the case, the streams should be non-axial? @Thomas-Benson @CBurton90

CBurton90:
@mnaguit I'd tend to agree. Chucked a comment in the ticket. We will need to update the instrument config if we go with non-axial (easily done).

mnaguit:
> @mnaguit I'd tend to agree. Chucked a comment in the ticket. We will need to update the instrument config if we go with non-axial (easily done).

Thanks @CBurton90

ozym:
I think this site is expected to become part of the national network (NZ) rather than the chateau network (CH).  That is, it has a strong motion oriented North/South and suitable comms upto national network specs.

Thomas-Benson:
`streams.csv` updated to axial

mnaguit (approved):

CBurton90 (approved):

PR #1905, opened 2023-11-06
TEST - NEW Lennartz MKIII Borehole - KBAZ metadata building - 

salichon:
Test 
- Validate New sensor entry into Delta "new generation"
- Update in implementation of change  (usage of XML snippets and/or auto.go)


CBurton90:
Depth is 160m.

resp/responses/sensors/lennartz.yaml
```diff
@@ -88,5 +99,16 @@ filter:
     delay: 0
     inputunits: m/s
     outputunits: V
+  LE-3D-BH-MkIII:
+  - type: paz
+    lookup: LE-3Dlite
+    frequency: 15
```
CBurton90, line 105: Is frequency 15Hz or 5Hz?

resp/responses/configurations/shortperiod.yaml
```diff
@@ -66,6 +66,12 @@ response:
       - LE-3DliteMkIII
       channels: ZNE
       reversed: false
+    - sensors:
+      - LE-3D-BH-MkIII
+      filters:
+      - LE-3DliteMkIII
+      channels: ZNE
```
CBurton90, line 73: Channels are `Z12`

CBurton90:
Few comments but looks good for the most part.

@ozym Can you confirm if we are still using the old yaml files with the "new" method.

resp/responses/sensors/lennartz.yaml
salichon: @CBurton90  will need to check that (since it s derived from the default delat information for LE3D_lite_ as opposed to BH

resp/responses/configurations/shortperiod.yaml
salichon: okay - might need to change this too

ozym:
We've stopped using those files, unless there is a fault which needs rebuilding.

salichon:
Clean PR will be performed -this one was a practice

PR #1912, opened 2023-11-16
PHHS Basalt to EQR120 SM Recorder Swap

Thomas-Benson:
- Removing Basalt (1277) from PHHS LC 20
- Replacing with EQR120 (46661) at same location
- Using latest EQR120 firmware version 4.3.2
- Updated streams to continuous from 2020-02-09
- Streams affected: PHHS LC 20 HN-12Z


install/firmware.csv
mnaguit, line 1: use opening time of recorder? thanks @Thomas-Benson

mnaguit (changes requested):
tweak opening time of firmware, to match opening time of recorder

install/firmware.csv
Thomas-Benson: Fixed, my bad for using the closing time instead.

mnaguit: All good @Thomas-Benson 😄

mnaguit (approved):

gva6500 (approved):

PR #1942, opened 2023-12-13
KBAZ Q330S/6 to Centaur Datalogger swap

Thomas-Benson:
- Removing Q330S/6 (6897) from place Karaka Road Borehole
- Replacing with Centaur CTR4-6S at same place
- Using latest Centaur firmware version 4.8.5
- Streams affected: KBAZ LC 10 EH-12Z


CBurton90 (approved):

mnaguit (approved):
Looks good, thanks @Thomas-Benson ... although I got confused with the Q330S/6 because the serial number indicated in the ticket was `6844` instead of `6897`.

PR #1950, opened 2023-12-21
GRAS adjust LC 22 closing time to align with recovered data

Thomas-Benson:
Adjusting closing time of EQR120 (46508) at GRAS LC 22 in delta to match new end data of data after recovering and uploading data from removed recorder.


mnaguit (approved):
Thanks @Thomas-Benson 👍

PR #1952, opened 2024-01-12
chore: upgrade go version to 1.21

CallumNZ:
Reference: https://github.com/GeoNet/tickets/issues/13645


ozym (changes requested):
Please wait until fedora has released 1.21 as its base package. Otherwise I'll need to maintain multiple versions.

ozym (approved):

PR #1953, opened 2024-01-14
Add time correction table, and an initial application for DART bouys

ozym:
This PR adds a new table:

__timings.csv__

This is meant to be used when the archived data will have a time issue that can't easily be corrected, due to formats etc.  It can also be used when a running correction is needed to be applied.

I have also added this to the list of possible corrections which makes building config files easier, an example is the dart config file. In this case it adds a correction duration to NZB for the 42 deployment.


cmd/dart-config/config.go
```diff
@@ -8,15 +8,16 @@ import (
...
-	Start     time.Time `json:"start"`
-	End       time.Time `json:"end"`
+	Network    string        `json:"network"`
+	Buoy       string        `json:"buoy"`
+	Location   string        `json:"location"`
+	Latitude   float64       `json:"latitude"`
+	Longitude  float64       `json:"longitude"`
+	Depth      float64       `json:"depth"`
+	Detide     *Detide       `json:"detide,omitempty"`
+	Correction time.Duration `json:"correction,omitempty"`
```
CallumNZ, line 18: Suggestion: `TimingCorrection`

why: The existence of the `Correction` struct that contains a `Timing` field is confusing.

meta/timing.go
```diff
@@ -0,0 +1,136 @@
...
+func (t *TimingList) decode(data [][]string) error {
+	if !(len(data) > 1) {
+		return nil
+	}
+
+	var timings []Timing
+
+	fields := timingHeaders.Fields(data[0])
+	for _, row := range data[1:] {
+		d := fields.Remap(row)
```
CallumNZ, line 88: Not clear what `d` stands for.

meta/timing.go
```diff
@@ -0,0 +1,136 @@
...
+			return err
+		}
+
+		end, err := time.Parse(DateTimeFormat, d[timingEnd])
+		if err != nil {
+			return err
+		}
+
+		var correction time.Duration
+		if s := d[timingCorrection]; s != "" {
```
CallumNZ, line 101: Not clear what `s` stands for.

salichon:
Super thanks -  ill go through the review 
and remoraks

cmd/dart-config/config.go
ozym: Yup, fixed as suggested.

install/README.md
```diff
@@ -421,6 +422,34 @@ A list of _doas_ installations, these include values for:
 |  _Start_ | Installed DOAS start time | 
 |  _Stop_ | Installed DOAS stop time | 
 
+#### _TIMINGS_ ####
+
+There are some data sources which may have timing issues and the type of recording format means
+that the archived raw data will include the bad timing and a correction will need to applied.
```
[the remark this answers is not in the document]
ozym: Done

meta/timing.go
ozym: have updated variable name, it's just the same as row except the order of the entries can be adjusted based on the labels of the underlying csv.  i.e. it knows where to find values based on the csv label, rather than the order of the entry.

meta/timing.go
ozym: s is simply a string,  a bit convoluted but saves on typing.

meta/correction.go
```diff
@@ -75,6 +84,81 @@ func (set *Set) Corrections(coll Collection) []Correction {
...
+
+	// no timings found process an empty timings correction
+	if !(len(timings) > 0) {
+		return []Correction{{Span: coll.Span}}
+	}
+
+	var res []Correction
+
+	// check prior to the first timing
+	if v := timings[0]; v.Start.After(coll.Span.Start) {
```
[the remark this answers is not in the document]
ozym: Yes, as that case is handled normally.  This is just looking for any parts prior to the first timing being set, which means within the span there shouldn't be any initial time corrections until the first one appears.

CallumNZ (approved):

JonoHanson (approved):
LGTM

jajera (approved):

PR #1954, opened 2024-01-16
CPFS Basalt to EQR120 SM recorder swap

Thomas-Benson:
- Removing Basalt (557) from CPFS LC 21
- Replacing with EQR120 (46594) at same location
- Using EQR120 firmware version `2.2.0`
- Updating recorder orientation to `49`
- Streams affected: HNPS LC 21 HN-12Z


CBurton90 (approved):

salichon (approved):
awesome :)

PR #1956, opened 2024-01-18
FOXKP closure

rumachan:
FOXKP is closed as it was destroyed by a rockfall on December 25.


elidana (approved):
sad to see this one destroyed, but the pr looks good, thanks @rumachan !

PR #1959, opened 2024-01-24
Individual GNSS antenna calibration files - Geoscience Australia

aleks-spes:
Add individual GNSS antenna calibration files
Change files permission. Remove execute option.
Add notes into the assets/antennas.csv file.


elidana (approved):
this looks pretty good to me, thanks @aleks-spes !
I did not pay big attention to individual serial numbers (assets file vs. calibration files added) so if @EvanCGNS can have a more careful double check with that, that'd be great.
Thanks!

PR #1966, opened 2024-02-08
RUD04 open

rumachan:
Files changes to open RUD04, scanDOAS at Wahianoa on SE flank of Ruapehu


elidana (approved):
looks good, thanks @rumachan !

PR #1970, opened 2024-02-17
feat: add chart config application

ozym:
This application will be replacing the existing amplitude and spectral chart configuration tools as part of a reworking of the chart plotting mechanisms


jajera (approved):
looks good on platform's perspective

PR #1971, opened 2024-02-18
MBAZ SP borehole sensor swap

Thomas-Benson:
- Removing 2 Hz Duke Malin Seismometer (600) from MBAZ LC 10 borehole
- Replacing with LE-3D-BH-MkIII (J-227) at same location
- Using same orientation as previous sensor 0(N) - may need future updates
- Streams affected MBAZ LC 10 EH-~ENZ~12Z


mnaguit (approved):
LGTM, noting LC 10 is streaming `EH-12Z`  - sensor orientation may need to be updated later

CBurton90 (approved):

PR #1974, opened 2024-02-20
feat: add a sensor catalogue app

ozym:
This makes a single output for installed sensors and related equipment

an example of the output can be found in `cmd/sensor-config/testdata/sensor.xml`

### DIFF
--- a/cmd/sensor-config/installed.go
+++ b/cmd/sensor-config/installed.go
@@ -1,0 +1,117 @@
+package main
+
+import (
+	"regexp"
+	"sort"
+	"strings"
+
+	"github.com/GeoNet/delta/meta"
+)
+
+func (n *Network) InstalledSensors(set *meta.Set, match *regexp.Regexp, network, prefix string) error {
+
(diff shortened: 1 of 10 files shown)


cmd/sensor-config/installed.go
```diff
@@ -0,0 +1,117 @@
...
+
+					Vertical: c.InstalledSensor.Vertical,
+					North:    c.InstalledSensor.North,
+					East:     c.InstalledSensor.East,
+
+					StartDate: c.InstalledSensor.Start,
+					EndDate:   c.InstalledSensor.End,
+				}
+
+				sensors[sensor] = append(sensors[sensor], c.Code())
```
bpeng, line 54: don't fully understand logic, could multiple sensors with same properties happen here?

bpeng (approved):
Code looks good, some double check before merge

ozym:
Let's call this pre-release. I want to expand the layout to cover more details which should help other applications and also think about a json output option.

PR #1978, opened 2024-02-28
NEZ TC-120PH-2 BB borehole sensor swap

Thomas-Benson:
- Removing TC 120PH-2 (9312) from LC 12
- Replacing with TC 12-PH-2 at same location code
- Sensor depth changed from 20m to 15 m due to borehole degredation, leaving location code unchanged however since very little usable data was streamed at the previous depth
- Orientation left as 0/N like previous sensor
- Streams affected KEZ - LC 12 - HH-12Z


mnaguit:
Thanks @Thomas-Benson , will send my review once the datalogger details are added. Is that okey? Cheers!

Thomas-Benson:
Added entries for missed datalogger deployment:
 - Removed Q330S/6 (6482) from `North Egmont Visitor Centre`
 - Replaced with Centaur (9902) at same place
 - Using same opening & closing times as sensor since this matches the last and newest data from the old and new dataloggers
 - Uing latest centaur firmware version 4.8.5
FYI @mnaguit

mnaguit (approved):
Thanks @Thomas-Benson 👍

gva6500 (approved):

PR #1980, opened 2024-02-29
WBHS BASALT to TitanSMA SM recorder swap

Thomas-Benson:
- Removing BASALT (2156) from LC 21
- Replacing with TitanSMA (002208) at same location
- Recorder is at new orientation 0/N
- Change in channel codes from axial to non-axial (12Z -> ENZ)
- Using latest TitanSMA firmware 4.5.22
- Streams affected: WBHS LC 21 HN-12Z+ ENZ


mnaguit (approved):
👍

gva6500 (approved):

PR #1982, opened 2024-02-29
MSZ CMG-3ESP to TH120-2 BB sensor swap

Thomas-Benson:
- Removing CMG-3ESp (T3E65) from MSZ LC 10
 - Replacing with TH120-2 (2837) at same location
 - Streams affected MSZ LC 10 HH-ENZ


gva6500 (approved):

mnaguit:
Just a thought: should we update the LH* streams as well?

MSZ_LH

gva6500:
> Just a thought: should we update the LH* streams as well?
> 
@mnaguit  this was just a Seismometer swap, so as far as I know there should be no changes to streams....  but I could very well be wrong 😬

mnaguit:
> > Just a thought: should we update the LH* streams as well?
> 
> @mnaguit this was just a Seismometer swap, so as far as I know there should be no changes to streams.... but I could very well be wrong 😬

Most of the time you are very right @gva6500 😄 
Not sure where I got the idea, probably just Friday blues. I thought there was a discussion in the recent past, to close all LH*. Looks like I am very wrong. 😆 

... Approving this now, in a few seconds

mnaguit (approved):
Thanks @Thomas-Benson 😄

CBurton90:
@mnaguit Were you thinking of LN* streams for SM? I think those were the streams we were closing off in delta (doubt there are any actually running in the network). LH* broadband streams are here to stay though : ).

mnaguit:
Hahaha, thanks for that @CBurton90 ! I knew there's something about the 1hz streams! 🤣

PR #1997, opened 2024-03-12
close white_mc

rumachan:
Close all White Island MC locations with the date of the last observation (as this equals last sample and last time location visited).

This does not include WI000 which remains open as it is the site used for "gas flights".


elidana (approved):
for the little I know looks good, i didn't do any data check, but happy to do some if valuable

rumachan:
@elidana HELP I don't know how to deal with this. There were no conflicts, but I think Mark's points.csv/sites.csv changes yesterday have introduced them. I made the same changes to points.csv and sites.csv. The points.csv file is correct.

elidana:
@rumachan I compared the differences that were causing conflicts and it was only on the end date, that you are modifying in this pr. I've resolved the conflicts, hopefully now that is fixing this?

elidana (approved):

PR #1908, opened 2023-11-10
Update firmware.csv

ricardojohansson:
updated firmware on stock CR6

PR #1998, opened 2024-03-13
chore: update firmware of two envirosensors dataloggers

elidana:
replace https://github.com/GeoNet/delta/pull/1908


mnaguit (approved):
👍

PR #2005, opened 2024-03-24
Update coordinates of MAKO

aleks-spes:
LINZ asked to update coordinates of MAKO to have them aligned with the [Geodetic database](https://www.geodesy.linz.govt.nz/gdb/index.cgi?code=MAKO&sessionid=101605915271221709666065&mode=)


EvanCGNS (approved):
Double checked calculations, looks good

PR #2007, opened 2024-03-25
Historical_Overseas_Stations.md

salichon:
For information only. 
Document Delta past content of Overseas stations references. 
Overseas Station localization content only.
Removed from repository since.

### DIFF
--- a/docs/stations/Historical_Overseas_Stations.md
+++ b/docs/stations/Historical_Overseas_Stations.md
@@ -1,0 +1,30 @@
+## About overseas stations in Delta repository - Reference ##
+Historical long standing seismological international stations were refered to 
+in Delta/network/stations.csv and Delta/network/networks.csv as the "OE" network
+these have been removed since. 
+
+#### Station Recordings and Instrumentation ####
+Most likely used in the paper seismograms and pre-miniseed recordings. 
+	
+#### Countries Australia, U.S.A., Japan  ####
+Delta reference as it was for "historical stations": 
+> BRS,OE,Brisbane,-27.389720485,152.774988403,525,,WGS84,1937-09-01T00:00:00Z,9999-01-01
+> CAN,OE,Canberra,-35.3208,148.9987,700,,NZGD49,1958-01-01T00:00:00Z,9999-01-01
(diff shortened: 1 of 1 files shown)


docs/stations/Historical_Overseas_Stations.md
```diff
@@ -0,0 +1,30 @@
+## About overseas stations in Delta repository - Reference ##
+Historical long standing seismological international stations were refered to 
+in Delta/network/stations.csv and Delta/network/networks.csv as the "OE" network
+these have been removed since. 
+
+#### Station Recordings and Instrumentation ####
+Most likely used in the paper seismograms and pre-miniseed recordings. 
+	
+#### Countries Australia, U.S.A., Japan  ####
+Delta reference as it was for "historical stations":
```
Thomas-Benson, line 10: The content here is ok, but if you preview this file it looks like github markdown isn't handling the line breaks/paragraphs here how you probably intended it to, and this csv block is all just running together instead of having each entry on its own line. Easiest fix would either be to surround this block in code tags (triple backtick \`\`\`) or add double spaces at the end of each line.

ozym:
Also noting that these locations (stations) may have been used in the locations of events prior to seiscomp, so need to be retained as the reference locations used at the time.

salichon:
- Will sort the page markdown
- Shall pass it on into a delta meeting for information/legacy purposes.

salichon:
@Thomas-Benson  would you give it another go ? 
thanks a lot!

docs/stations/Historical_Overseas_Stations.md
```diff
@@ -0,0 +1,37 @@
+## About overseas stations in Delta repository - Information/Reference ##
+Historical long standing seismological international stations were refered to 
+in Delta as in the "OE" network.
+
+Reference for "OE" stations in delta: https://github.com/GeoNet/delta/blob/main/network/stations.csv.
+
+N.B.: The network "OE" was removed from network/networks.csv since
```
Thomas-Benson, line 7: Since when? 2019 I think based on https://github.com/GeoNet/delta/pull/598

PR #2008, opened 2024-03-26
WEL LC 21 clean up

mnaguit:
remove entries with EpiSensor 2


salichon (approved):
looks good to me WEL short lived episensor 2

CBurton90 (approved):

PR #2009, opened 2024-03-26
NA002_multigas

rumachan:
Additional of sensors, etc for multi-gas at NA002


assets/sensors.csv
```diff
@@ -40,8 +41,10 @@ Campbell Scientific,Titanium pressure transducer CS456,14010302,,
 Campbell Scientific,Titanium pressure transducer CS456,70010135,,
 Campbell Scientific,Titanium pressure transducer CS456,70010240,,
 Campbell Scientific,Titanium pressure transducer CS456,70010248,,
-City Technology,T3STF SO2 sensor,01.45929880 013,,
 City Technology,T3H H2S sensor,01.44557167 112,,
+City Technology,T3H H2S sensor,01.44865262 112,,
+City Technology,T3STF SO2 sensor,01.45929880 013,,
```
elidana, line 46: hei @rumachan I cannot find a reference to this sensor (`T3STF SO2 sensor,01.45929880 013`) in the ticket, but only the `T3STF SO2 sensor, 01.45929894 013` sn. Is that extra entry (`01.45929880`) a typo?

elidana (approved):
Hi @rumachan , this looks good. I think there is an extra entry in the assets file, but all "installed" metadata looks good. 
I did a check against the raw data to verify sensor code against measured observable

assets/sensors.csv
rumachan: I swapped the order so that it was alphabetical. The entry was already there, just moved. It doesn't seem to matter wit the checks, but wanted it to be in order. Swapped one was from TO003. Sorry, thought it woud be clear it was just moved elsewhere in file. Thank you

elidana: ah thanks ! sorry I missed that. thanks for the explanation!

PR #1967, opened 2024-02-13
Strong Motion Azimuth Deployment Guide/Doc

CBurton90:

PR #2010, opened 2024-03-26
Smtc sm recorder swap

Thomas-Benson:
- Removing CSI EQR120 (46531) from SMTC LC 20
 - Replacing with EQR120 (46628) at same location
 - New EQR120 running latest firmware version 2.3.2
 - Using same orientation 0/N as previous recorder


mnaguit (commented):
special case when axial flag is true but azimuth is 0?

Thomas-Benson:
This site was originally installed at `182`. When the sensor was replaced in march 2021 the orientation was change to true north, but it looks like a decision was made then to keep the stream names as axial since this isn't a free field site.

CBurton90:
@mnaguit Again we are trying to standardize what to do in this scenario. I'm pretty sure we are all in agreement with this guide (after many discussions + delta meetings with this problem):

https://github.com/GeoNet/delta/pull/1967

This PR matches that guide.

CBurton90 (approved):

mnaguit:
Okidoki...a special case then. 😄 Thank you Gentlemen!

mnaguit (approved):

PR #2012, opened 2024-04-02
fix: update wqab to use xyz channel labelling

ozym:
This is to account for the change in subsource labelling whereby the response expects Z12 channel names, but the generated miniseed data uses XYZ labelling.


mnaguit:
thank you @ozym 👍 👍  will do the check after the merge.

mnaguit (approved):

jajera (approved):

PR #2013, opened 2024-04-03
Fix the antenna start date - WGTT

aleks-spes:
Fix the antenna start date timestamp to align the antennas.csv and radomes.csv tables


elidana (approved):
this looks good.

Have generated sitelogs and station.info locally and they both look fine.

PR #2024, opened 2024-04-11
BSWZ L4C-3D to TC120PH-2 Sp to BB sensor swap

Thomas-Benson:
- Removing Sercel L4C-3D (2996) from BSWZ LC 10
 - Replacing with Trillium Compact 120PH-2 (8783) at same location
 - Switch stream codes from E,H to H,H to reflect change in deployed sensor type from short period to broadband


CBurton90 (approved):

mnaguit (approved):
all good, probably (or probably not) a little archive correction needed for the stream components. thanks @Thomas-Benson

PR #2026, opened 2024-04-18
OPSS CUSP3D to TitanSMA SM recorder swap

Thomas-Benson:
- Removing CUSP3D (42038) from OPSS LC 20
 - Replacing with TitanSMA (002212) at same location
 - Using latest firmware version 4.5.22
 - Merged entries in `recorders.csv` to remove uneccessary 1 minute gap.


CBurton90 (dismissed):

mnaguit (commented):
Should the orientation remain as is? like 240? Thanks @Thomas-Benson

Thomas-Benson:
My bad @mnaguit, the orientation should be correct now!

gva6500:
@Thomas-Benson  I think my head was also all over the place yesterday....   Regarding OPSS, Paul confirmed orientation to be alligned with Magnetic North. so the aximuth should be ~21deg. 

Thought I mentioend this in the ticket but seems I did not 😞  I had a chat with Paul yesterday I a noticed that last instrument was recorded as 240... but the first install as 24...   Paul says he the titan was installed at the same orientation as the cusp that was replaced.   So likely that the previous entry was wrong.  The 24deg of the initial would probably be it, accouting for use of a compass.  Likely that the 24 turned to 240 by acident...  As  it wont allign with the building at any  orientation.

Thomas-Benson:
Updated the orientation of this and the precious recorer to 240, hopefully that's all correct now @gva6500?

gva6500 (approved):

gva6500:
> Updated the orientation of this and the precious recorer to 240, hopefully that's all correct now @gva6500?

I hope so too!  at the very least the current instrument is captured correctly ;-)

mnaguit (approved):

PR #2030, opened 2024-04-22
HWHS EQR120 SM recorder install

Thomas-Benson:
- Installing EQR120 46663 at HWHS LC 21
 - Reopening HWHS in `network/stations.csv`
 - Adding new site to `network.sites.csv`
 - Using most recent EQR120 firmware version 2.3.2


mnaguit (commented):
Thanks @Thomas-Benson . Should the streams and recorders be for LC `21` instead of `20`?

Thomas-Benson:
You're right, my bad. I got a little mix up with what was happening with the site/stream codes here. Everything should be using the correct LC `21` now.

mnaguit (approved):
Thanks @Thomas-Benson

gva6500 (approved):

PR #2032, opened 2024-04-29
fix: update geomag calibration info

ozym:
This adjusts the calibration end times, the sensors were moved rather than being recalibrated or retired


elidana (approved):
oh apologies, I missed that in the previous pr!

tanjapetersen:
@ozym is to capture that the fluxgate that was located at Scott Base (2009 - 2023) is now located at EY2M?

ozym:
> @ozym is to capture that the fluxgate that was located at Scott Base (2009 - 2023) is now located at EY2M?

@tanjapetersen no, it's to reflect the calibration of the sensor (i.e. nT -> volts) hasn't changed. Otherwise there would need to be a fall back option for sensors that aren't calibrated (e.g. a standard model) which we haven't done.

PR #2036, opened 2024-05-01
add RU & TO scandoas to features

rumachan:
Additions to environment/features.csv for RU and TO scanDOAS sites. Required for those data to go into Tilde. Other Delta files required no changes.


ozym (approved):

elidana (approved):
looks good and agree with names discussed with scandoas subject matter experts

rumachan:
@ozym this failed `govulncheck` after I updated the branch. Do you understand this? Thanks

ozym:
Yes, there's a program called govulncheck which has a known list of security vulnerabilities (usually in libraries etc.) it gets updated fairly quickly after a vulnerability is announced. But, github actions is a little slow in updating it's releases, so it will be fixed the next time the actions is updated for a new go release version.  We can either skip this test, or wait.

rumachan:
Thanks @ozym Are the actions likely to update before next week? If not, is it possible to skip in this case? It'd allow me progress things.

Though I'm happy with whatever you think is the most sensible approach.

PR #2041, opened 2024-05-03
scandoas into Tilde

CallumNZ:
Reference: https://github.com/GeoNet/tickets/issues/15607


ozym (approved):
looks fine, nice use of the new slices package.

PR #2025, opened 2024-04-15
Westmelton2 GeoMag equipment update

TurnerDave:
Updated metadata for Fluxgate sensor and Centaur datalogger at the new West Melton 2 GeoMag site

- [x] network/stations.csv (geomag has network code = GM)
- [x] network/sites.csv (geomag has channel code = 50 for the fluxgate and 51 for the proton)
- [x] calibrations/fluxgate : a text file with Danish Met Insitute calibration results for individual fluxgate sensor and driver 
- [x] assets/dataloggers.csv
- [x] assets/recorders.csv  _comment: N/A_
- [x] assets/sensors.csv  _comment: Existing sensor redeployed_
- [x] install/calibrations.csv   _comment: Existing sensor redeployed_
- [x] install/channels.csv (check that model is there)
- [x] install/components.csv (check that model and response file is there)
- [x] install/connections.csv
- [x] install/dataloggers.csv
- [x] install/gains.csv
- [x] install/recorders.csv   _comment: N/A_
- [x] install/sensors.csv
- [x] install/streams.csv
- [x] /files (check model is there)  _comment: resp/files/sensor_Danish_Meteorological_Institute_FGE.xml exists_

PR #2042, opened 2024-05-04
fix: update fge bias for EY2M

ozym:
Appears the offsets were in the wrong order, i.e. xyz -> zyx

Tested and the values come much closer, and the two sets at West Melton look closer in offsets as well.


tanjapetersen:
Hi @ozym - the title says EYRM, but from what you mentioned before issue with the offsets is in the EY2M data??

ozym:
> Hi @ozym - the title says EYRM, but from what you mentioned before issue with the offsets is in the EY2M data??

good catch, have updated title.

elidana (approved):
oh good catch @ozym !!! 
did check against photos taken at time of installation and happy with that fix. Apologies if I missed that while reviewing https://github.com/GeoNet/delta/pull/2025

tanjapetersen:
nT values should be the same for EYWM and EY2M after the change
EYMvsEY2M

elidana:
once metadata have been updated, the change should be "deployed to tilde" and data potentially reprocessed before that to be visible on the plots?

ozym:
this doesn't get "deployed" to tilde.

tanjapetersen:
I just wanted to point out a quick way to check if calibration is correct. Both instruments should record the same magnetic field

elidana:
ha ha cool, all good and thanks for the tip!

PR #2044, opened 2024-05-07
chore: add Bartington response

ozym:
This adds the Bartington FGE sensor response.

Assumes 10,000 nT == 1 Volt (e.g 1 nT == 1/10000 Volts)
Frequency chosen as 1 Hz, not important unless full response is given.

Also response only applies to Z component.


elidana (approved):
thanks @ozym ! And sorry you had to do that yourself...

PR #2045, opened 2024-05-08
chore: update tilde config for geomag

ozym:
This adds the geomag meta data config for tilde, it also handles the edge case with SMHS


cmd/tilde-config/main.go
```diff
@@ -41,10 +44,23 @@ func main() {
...
 	flag.StringVar(&settings.output, "output", "", "output dart configuration file")
 
 	flag.Parse()
 
+	extra := make(map[string][]string)
+	for _, s := range strings.Split(settings.extra, ",") {
+		if s = strings.TrimSpace(s); s != "" {
+			keys := strings.Split(s, "=")
+			if len(keys) != 2 {
+				continue
```
wilsonjord, line 58: question: we don't want to throw error here? this is effectively passing incorrect args to the CLI?

wilsonjord (commented):
just a small question from me.
another question is the `chore:` type...it would seem `feat` would have been more appropriate?

cmd/tilde-config/main.go
ozym: good point, yeah it's a bit transparent. will update.

ozym:
> just a small question from me. another question is the `chore:` type...it would seem `feat` would have been more appropriate?

feels like a chore, but yup.

CallumNZ (approved):
LGTM

PR #2047, opened 2024-05-14
fix TOD01 end mistake, features.csv

rumachan:
TOD01 end time had not been set in environment/features.csv. Showed still open but was closed when instrument was removed to install elsewhere.


elidana (approved):
looks good and good catch!

PR #2051, opened 2024-05-21
feat: add grouped XML formatted sensor list

ozym:
This XML format is a little easier to handle sensors that are installed at stations with unrelated network details, such as an acoustic sensor at a coastal gauge site, or a geomagnetic sensor at a strong motion site.

It removes the concept of the network (other than the total as a whole) and adds groups of sensors with a given characteristic. The type of installation is maintained, such as a Station/Sensor, or a Mark/Monument, or a Mount/View.

### DIFF
--- a/cmd/sensor-config/envirosensor.go
+++ b/cmd/sensor-config/envirosensor.go
@@ -104,3 +104,107 @@ func (n *Network) EnviroSensor(set *meta.Set, enviro string, label string) error
 
 	return nil
 }
+
+func (s Settings) EnviroSensor(set *meta.Set, name, enviro string) (Group, bool) {
+
+	if _, ok := set.Network(enviro); !ok {
+		return Group{}, false
+	}
+
+	valid := make(map[string]interface{})
+	for _, f := range set.Features() {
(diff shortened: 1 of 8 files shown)


cmd/sensor-config/envirosensor.go
```diff
@@ -104,3 +104,107 @@ func (n *Network) EnviroSensor(set *meta.Set, enviro string, label string) error
 
 	return nil
 }
+
+func (s Settings) EnviroSensor(set *meta.Set, name, enviro string) (Group, bool) {
+
```
bpeng, line 109: seems this is almost the same as the above `func (n *Network) EnviroSensor(set *meta.Set, enviro string, label string)`, is it possible to share code? there are other similar cases...

ozym: The idea is we only use one output, this is aimed at keeping the existing method and adding a slightly different new method. Once the existing code is no longer needed, or the new one is rejected, then the code can be cleaned up a lot.

bpeng (approved):
Looks good

PR #2053, opened 2024-05-22
Update the PositioNZ-RT network coordinates

aleks-spes:
Update marks.csv to include updated coordinates for the PositioNZ-RT stations.


elidana (approved):
looks good but I did not check all coordinates in detail. If @EvanCGNS can have another look w.r.t. what LINZ provided that'd be great!

EvanCGNS (approved):
I can confirm all updated coordinates are correct!

PR #2057, opened 2024-05-27
chore: add SEPCHOKE antenna model for GNSS sitelogs production

elidana:
Hi @ozym ,

we have a new antenna model now installed and need to update the GNSS sitelogs generation accordingly. 

The antenna graph has been copied from the official IGS file (link is also updated here, as the old IGS website was deprecated a while ago).

@aleks-spes can you also please double check the graph and that I picked the correct one for TGHM? 

@ozym , hope I added this in all the right places. There's a parallel pr to actually install the new antenna at the site

PR #2058, opened 2024-05-28
feat: add check for radome graph for sitelogs

ozym:
This update allows the sitelog code to check if there is a graph which includes a radome, if present, in which case that graph will be used. Otherwise, the antenna only graph will be used. In the configuration the antenna and radome should have a space between them.

### DIFF
--- a/tools/sitelogs/main.go
+++ b/tools/sitelogs/main.go
@@ -591,10 +591,19 @@ func main() {
 					var graphs []string
 					models := make(map[string]interface{})
 					for _, a := range antennas {
-						if _, ok := models[a.AntennaType]; ok {
+						antennaType := a.AntennaType
+						// check for any graph which may include radome
+						if a.AntennaRadomeType != "NONE" {
+							withRadome := a.AntennaType + " " + a.AntennaRadomeType
+							if _, ok := antennaGraphs[withRadome]; ok {
+								antennaType = withRadome
+							}
+						}
+
+						if _, ok := models[antennaType]; ok {
 							continue
 						}
-						switch g, ok := antennaGraphs[a.AntennaType]; {
+						switch g, ok := antennaGraphs[antennaType]; {
 						case ok:
 							b, err := hex.DecodeString(g)
 							if err != nil {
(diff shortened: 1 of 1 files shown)


ozym:
To use this feature the config/config.yaml file needs to have extra graphs added.

elidana:
Hi @ozym , 

this is great thank you!!! 
I tested this with the new model and wasn't fully working, so I added a couple of minor things added in my commit:
- IGS standard `antenna.gra` file add some extra spaces to separate antenna and radome
- the name to be shown on the sitelog is the antenna+radome model


as a side note: 
@aleks-spes , I cannot find the `antenna.gra` source with the antenna+radome model.
This one https://files.igs.org/pub/station/general/antenna.gra only has antennas. Where did you find the one you posted in https://github.com/GeoNet/delta/pull/2057#issuecomment-2133999272? 
I want to check that the other antenna+radome models we have installed have the same spacing style

aleks-spes:
Hey @elidana ! Please see the comment: https://github.com/GeoNet/tickets/issues/14226#issuecomment-1718674301

Regarding the antenna+ radome code convention, I believe they use of the RINEX standard:
`ANT # / TYPE        | Antenna number and type                  |    2A20 `
The sting length is 20 characters and the last 4 char are booked for radome code.

I still think we can go back and use only the antenna diagram, because the radome part of the diagram doesn't have useful information.

elidana:
@aleks-spes thanks. I saw your comment on the radome and have included that in my pr.

Can you point me to source file please? I wanted to check what is the spacing used for the other radomes we have installed in the geonet network

And now that we can, I think is a good addition to make our sitelogs more compliant! :)

aleks-spes:
Hey @elidana ! In that comment, there is a link to the log file from one IGS site: https://network.igs.org/api/public/download/GOP700CZE.log from which I took this plot, but it seems the file was updated a couple months ago and currently doesn't contain the diagram.

elidana:
Ah bummer. I was hoping we could get the information from an official IGS file, not someone else's site log

I'll write to the IGS antenna WG to ask.
Then I think we can revert back to use one space only, and wait and then update the graph and spacing in the code when we get the info about the antenna.gra

aleks-spes:
Hey @elidana !
We can find information from "official" sources (see below), but I think here it is more important to decide how we want to manage this kind of thing in the future.
image

elidana:
ok. for this let's keep what we have then. And we can check with IGS working group if there is an official IGS place where we can grab diagrams of other geonet antenna+radome combo.

will keep this pr, and add a ticket to remember to do the rest.

elidana (approved):
thanks @ozym !

tested and works fine

elidana:
can someone please in the @GeoNet/platform-team  or @GeoNet/ssd-team team review this?
Otherwise it can't be merged.

There's some parallel "metadata convo" but I tested the code locally (added a small tweak with the second commit) and all works fine

wilsonjord (approved):
No issues here. I had a vague feeling if the construction of "antenna picture" should be part of `GnssAntenna` itself, but didn't know enough to formulate anything concrete.

PR #2068, opened 2024-06-04
chore: update sensor xml to add types at the sensor level

ozym:
This adds the sensor type to the sensor struct, usually this will match the group name.


bpeng (approved):
OK

PR #2069, opened 2024-06-05
Update firmware.csv

ricardojohansson:
Updated firmware version on TO006 CR6 (SN 4110) yesterday onsite.


rumachan (approved):

elidana:
the firmware record has been already captured in delta, and with timestamp that agree with data records
(i.e. new firmware in data starting from `TO006_2024-06-04T04:00:00Z_4110.toa5.csv`)

this one can be closed without merging

PR #1990, opened 2024-03-04
cmd: fix a few minor typos in comments

robvdl:

PR #2070, opened 2024-06-05
bug: update spelling

ozym:
Fix a couple of minor typos

This is based on #1990 which got left in merge purgatory


raki-git (approved):
good catch Mark

PR #2072, opened 2024-06-06
chore: remove tides module

ozym:
This module is no longer needed as the current Set mechanism can handle this requirement, it also had not been updated to account for the new stations


salichon:
Ok thanks @ozym! 
so  to be clear to me ... :) 
we are deprecating/cleaming "github.com/GeoNet/delta/tides" 
which is accounted for by "lasso/internal/tidal/",  aren't we ?

ozym:
> Ok thanks @ozym! so to be clear to me ... :) we are deprecating/cleaming "github.com/GeoNet/delta/tides" which is accounted for by "lasso/internal/tidal/", aren't we ?

This was a hold over from when we used the LINZ tidal files directly, once we shifted to use the constituent tables etc, this module was no longer needed as it simply did the same thing.

PR #2073, opened 2024-06-06
fix: update auto set code generation to improve application speed

ozym:
This update removes the memory copies to improve the overall speed of operations, the original idea was to protect the underlying slice backends. But this approach is very slow with the memory allocation and handling, this update now returns the slices directly and is up to the module user to ensure integrity. The best approach is via using a range over the returned slice

On testing this change gives some programs a speed up of about x20.

### DIFF
--- a/meta/generate/generate.go
+++ b/meta/generate/generate.go
@@ -23,7 +23,7 @@ var generateTemplate = `
 package meta
 
 {{range $k, $v := .Fields -}}
-// {{title $k}} is a helper function to return a slice copy of {{$v.Key}} values.
+// {{title $k}} is a helper function to return a slice of {{$v.Key}} values.
 func (s Set) {{title $k}}() []{{$v.Key}} {
         return s.{{$k}}
 }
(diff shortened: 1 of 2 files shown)


meta/generate/generate.go
```diff
@@ -25,9 +25,7 @@ package meta
 {{range $k, $v := .Fields -}}
 // {{title $k}} is a helper function to return a slice copy of {{$v.Key}} values.
```
wilsonjord, line 26: question: this is not longer accurate right? `return a slice copy`

ozym: good catch, missed that update.

wilsonjord:
I think use of `chore` is not appropriate, I think replacing a copy of a data structure with a pointer to the underlying data structure itself not trivial...code that does pointer operations with the copy will potential be an issue?
I suppose it doesn't matter too much in the scheme of things, but I do think `refactor` or `feat` or something is more appropriate, but if the change is actually more trivial than I think it is, then sure stick with `chore`.

ozym:
Maybe it's a fix, in the sense of speeding things up.

wilsonjord (approved):

PR #2074, opened 2024-06-06
fix: remove the blessed impact mechanism

ozym:
This was a holdover from when certain instruments were not trusted, however now that the instruments have been consolidated the current response mechanism should be able to cope.  It is also the first step to replacing this code to allow for testing of the new version which also wouldn't have this check


wilsonjord (commented):
Seems good, looks like no public exported symbols to worry about.
Need README update though?

ozym:
This app has an end of life so I'm really just aiming to make the testing of its replacement easier. But more than likely the README will be copied as is will need updating then too.

wilsonjord (approved):
Ok, based on my understanding of previous comment, it seems the "Impact" app is EOL, so updating https://github.com/GeoNet/delta/blob/fix-impact/tools/impact/README.md is not really needed, so this PR looks fine.

PR #2076, opened 2024-06-07
chore: remove deprecated applications

ozym:
The tools chart applications are no longer used, these have been replaced by the cmd/chart-config application


wilsonjord (approved):
Appears to be a simple removal of 3 apps (not libraries): chart, amplitude, and spectra.

PR #2077, opened 2024-06-07
chore: remove unused sc3 applications

ozym:
These have been replaced by the version equivalent versions under cmd


CallumNZ (approved):
🧹

wilsonjord (approved):
lgtm, seem equivalent to `cmd/sc3-config` and `cmd/caps-config`.
Perhaps there was a good reason, but for future consider `git mv` to keep commit history.

ozym:
> lgtm, seem equivalent to `cmd/sc3-config` and `cmd/caps-config`. Perhaps there was a good reason, but for future consider `git mv` to keep commit history.

bit tricky as both had to run for together for a while until downstream could switch over.

PR #2078, opened 2024-06-09
feat: update polynomial coefficents to no longer be numbered

ozym:
This is to avoid having to decide whether to start from one or zero, each has its problems


raki-git (approved):
LGTM

PR #2079, opened 2024-06-09
feat: add a NRL directory to add externally derived response XML snippets

ozym:
These can be used in the same way as the existing files in components or channels


salichon:
Okay super 
 
As i undertsand for the context 
- `resp/files` locally created XML  files e.g. DART 15s. 
- `resp/auto` derived from `auto.go`  and the yaml files
-  now `resp/nrl` xml nrl snippet xml source from nrl service or archice (reference-source of truth)

salichon (approved):
file to "activate" the `nrl` directory path

PR #2081, opened 2024-06-10
chore: remove the cusp config app

ozym:
This cusp config app is no longer used and the configuration in this format is no longer needed.


salichon (approved):
moght need to precise the time uit to add or remove

junghao (approved):
Looks good

PR #2083, opened 2024-06-10
scandoas features change

rumachan:
Changes to environment/features.csv to account for latest changes to stream naming.
Applying to RUD and TOD stations. Resulting file a lot smaller as removed wind and plume completeness and changing how Aspect is used.


elidana (approved):
thanks @rumachan ! this looks pretty good to me

PR #2084, opened 2024-06-10
chore: replace response type in stationxml with resp module

ozym:
The resp module should now handle response file generation and encoding. This should be a no-op change to how things run.

The Frequency() part is needed to expose the private frequency value (which is private to avoid it being encoded into the XML), this is only needed for version 1.0 and will be removed once we no longer need to maintain 1.0 (and 1.1).

### DIFF
--- a/cmd/stationxml-build/build.go
+++ b/cmd/stationxml-build/build.go
@@ -5,8 +5,6 @@ import (
 
 	"github.com/GeoNet/delta/meta"
 	"github.com/GeoNet/delta/resp"
-
-	"github.com/GeoNet/delta/internal/stationxml"
 )
 
 // Builder is a cache of response files.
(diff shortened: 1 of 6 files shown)


junghao (approved):
Looks good

jajera (approved):
looks good on platform perspective

quiffman (commented):
LGTM

quiffman (approved):

salichon:
files changed 
> cmd/stationxml-build/build.go
> internal/stationxml/encoder10.go
> internal/stationxml/encoder11.go
> internal/stationxml/encoder12.go
> internal/stationxml/root.go
> resp/response_type.go

Built stationxml-build
Created a response file 
looks good thanks!

salichon (approved):

PR #2085, opened 2024-06-10
chore: update geomag config to use resp

ozym:
The resp module should now handle response file generation and encoding. This should be a noop change.


jajera (approved):
looks good on platform perspective

PR #2087, opened 2024-06-11
chore: update the impact tool

ozym:
Update the impact too to use Set and Resp XML snippets


jajera (approved):
for the better. lets give this a go

PR #2094, opened 2024-06-14
OPRZ add infrasound and replace seismic sensors

Thomas-Benson:
Adding ultrasound sensor at OPRZ LC 30
 - Setra 270-600/24V (2625799)
 - Adding neccessary streams, connection, site
 - Using same coordinates as existing seismic site
Replacing seismic sensor at OPRZ LC 10
 - Removing Guralp CMG-3ESP (T3E84)
 - Replacing with Nanometrics TC120PH-2 (10291)


mnaguit (approved):
Thanks @Thomas-Benson , LGTM. 
Just a thought: Is there a need to update the role in `connections` to acoustic?

Thomas-Benson:
@mnaguit the acoustic sensor is connected to the same datalogger as the seismic sensor, so we do not need to add roles. We'd only have to use roles if there were mulitiple dataloggers at the same place we were trying to differentiate between.

CBurton90 (approved):

PR #2102, opened 2024-06-18
fix: switch to use camera group

ozym:
This makes the group name for cameras "Camera" rather than "Volcano Camera"

This option simply lists the camera networks,  an alternative would be to have a master "camera" network code and use this as the external camera code.  This could be addresses if we add more camera networks.

### DIFF
--- a/cmd/sensor-config/settings.go
+++ b/cmd/sensor-config/settings.go
@@ -109,7 +109,7 @@ func (s Settings) Groups(set *meta.Set) Network {
 		network.Groups = append(network.Groups, group)
 	}
 
-	if group, ok := s.Cameras(set, "Volcano camera", s.volcano); ok {
+	if group, ok := s.Cameras(set, "Camera", s.camera); ok {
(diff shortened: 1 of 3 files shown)


cmd/sensor-config/settings.go
```diff
@@ -109,7 +109,7 @@ func (s Settings) Groups(set *meta.Set) Network {
 		network.Groups = append(network.Groups, group)
 	}
 
-	if group, ok := s.Cameras(set, "Volcano camera", s.volcano); ok {
+	if group, ok := s.Cameras(set, "Camera", s.camera); ok {
 		network.Groups = append(network.Groups, group)
```
bpeng, line 113: are `s.volcano` and `s.camera` the same?

ozym: no, at the moment the volcano one is used for the existing system (non-group) whereas camera is the new group version and is a list of networks (rather than a single network code).

bpeng (approved):
OK

PR #2110, opened 2024-06-21
fix: update tilde config to update site times instead of station times

ozym:
This should handle the case of sites closing meaning the sensor is no longer installed rather than relying on the station times.


jajera (approved):
+ for the better handling

PR #2113, opened 2024-06-21
MAVZ Install surface sensor + datalogger

Thomas-Benson:
Adding new surface sensor and digitiser to MAVZ
 - Installing Trillium Compact 120PH-2 (10215) at LC 11
 - Installing Centaur (10970) at place `Matarangi`
 - Using role `Surface` to distinguish new digitiser from the existing one connected to the borehole sensor
   - (noting that it is not necessary to add roles to the old datalogger(s) and connections)
 - Reopening site LC 11 since this is a reoccupation


mnaguit (approved):
Looks like a complex setup in this station, but these delta changes LGTM, for LC11
Thanks @Thomas-Benson

CBurton90 (approved):

PR #2116, opened 2024-06-24
WTVZ L4C-3D to TC120PH-2 SP to BB sensor upgrade

Thomas-Benson:
- Removing short period L4C-3D (2823) from WTVZ LC 10
 - Replacing with broadband TC120PH-2 (10094) at same location
 - Updating stream codes to broadband


mnaguit (commented):
Thanks @Thomas-Benson , do we need LH* added in streams?

CBurton90:
@mnaguit Yes, I'll do that today (at present there are no LH* streams). I had a few issues yesterday re-uploading the q330 xml config to the sites.

cc @mnaguit

Thomas-Benson:
Added LH-ENZ codes now that they are streaming, wasn't sure if they were going to be enabled or not.

CBurton90 (approved):

mnaguit:
Thanks @Thomas-Benson @CBurton90 👍

mnaguit (approved):

PR #2119, opened 2024-06-24
ALS2R close

rumachan:
Closing site. Happened in March, but only found out last week.
This is landslide response rain gauge on CR6 data logger. 

ALS1R was closed some months ago.


salichon (approved):
PR Fell into a (landslide ) crack 
All good compliant with geonet-archive

PR #2120, opened 2024-06-25
Update firmware.csv

ricardojohansson:
updated csv to reflect change of firmware on a logger currently in storage


elidana (approved):
thanks @ricardojohansson !!!

PR #2125, opened 2024-07-03
SMHS swap BASALT SM recorder to sensor + centaur

Thomas-Benson:
- Removing Basalt (1275) SM recorder from SMHS LC 20
 - Replacing with Centaur (10891) and Titan TACCL-N1 (3399) at same location
 - Added connection for new sensor, and changed existing connection for GeoMag sensor to point to the new datalogger


ozym:
The current blocks will likely go away when github updates it's golang version for github actions, likely within a day or so.

ozym (approved):
looks good to me.

CBurton90 (dismissed):

mnaguit (commented):
A few questions/comments:

- there is an existing sensor entry for SMHS at LC 50 (`Bartington | Mag670 L-100`), how does this relate with the new sensor (`Titan TACCL-N1`)? Should this sensor be left open?
   - and for Titan TACCL, do we need to add a new entry in the sensor? and add the azimuth of 230 as noted in the main ticket?  
- please update the streams for LC 20 from triggered to continuous (HN*)
   - for geomag: is there a need to update the streams for LC 50?

Thomas-Benson:
Fixed sensor entry and updated SM streams to continuous from `2020-02-09`. As far as I can tell the geomag streams are unchanged. 
The current setup is a new `Titan TACCl-N1` Strong motion sensor and the exisiting `Bartington Mag670 L-100` GeoMag sensor both connected to the new Centaur

CBurton90 (dismissed):

mnaguit (approved):

Thomas-Benson:
Added the GeoMag gains on the centaur here, so we can merge this PR without leaving the LC 50 response hanging

mnaguit (commented):
For the connections, would it look cleaner to use `Summer Hill `for all place entries? And for the new LC 20 entry, still in connections, is there a need to enter `0` in the Number row to indicate it is connected to Centaur? 

Otherwise, LGTM.

Thomas-Benson:
@mnaguit The pin number isn't strictly neccessary, since it is assumed by default that no pin number means pin 0, but I can add it to be tidy anyway.
As for the place names, the old connection on LC 50 has to remain as SMHS, since it is using the extra channels on a Basalt rather than a proper datalogger.

mnaguit (approved):
Thanks @Thomas-Benson 👍

CBurton90 (approved):

PR #2131, opened 2024-07-08
chore: remove legacy parts to sensor-config

ozym:
This updates the settings used for building the current sensor.xml file and removes the initial trial mechanisms including json output.


jajera (approved):
clean is good

PR #2140, opened 2024-07-18
feat: update visibility table to use Mark label

ozym:
This is part of making delta easier to reference via variables.

It is only used for GNSS sites and switching from Code to Mark makes it easier to track, other than the initial header swap there is no change to data or other applications.

### DIFF
--- a/environment/README.md
+++ b/environment/README.md
@@ -111,7 +111,7 @@ The level should be from 0 to 3, with 3 used for the smallest places.
 
 | Field | Description |
 | --- | --- |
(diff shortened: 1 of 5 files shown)


wilsonjord (approved):
lgtm. Not sure how exactly it improves as the PR comments said, but I'll just accept it at face value (I guess `Mark` is unambiguously one thing, whereas `Code` could be two things?)

jajera (approved):

elidana (approved):
looks good, thanks Mark!

PR #2146, opened 2024-07-24
CNCL - Fix start timestamp

aleks-spes:
Fix start timestamp in radomes.csv to align with antennas.csv


aleks-spes:
Hey @elidana ! Could you please have a look. This fix reslves an issue with CNCL.

elidana (approved):
good catch @aleks-spes !!!

PR #2151, opened 2024-07-25
feat: add extra equipment installation checks

ozym:
These checks help spot when something has been installed after it has been removed, or removed before it has been installed


CallumNZ (approved):
Nice

elidana:
@ozym this pr is now passing all tests

PR #2158, opened 2024-07-29
KHEZ - kahuihut - document infrasound (InfraBSU) gain configuration across datalogger history (q330/Centaur)

CBurton90:


salichon (approved):
Thanks @Thomas-Benson 
Smart "subsource" F ...  for 
_Subsource | The sensor channel(s), as defined in the response configuration,  which requires a gain adjustment, multiple subsource channels can be  joined (e.g "Z")_

IF EVER ..  out of  (IDK what .. maybe  clarity?) 
shall  we  also populate the _preamps subsource_  too ? 
..say for the the 3?'s  channels ? 

either ways is good

salichon:
Hello @Thomas-Benson @rumachan  
would you go  through this PR and check if still valid ? 
(fell in a crack instruments ...  infrasounds...) 
so that we can merge or dispose.! sorry 
FYI @CBurton90 
Cheers
jerome

Thomas-Benson:
@salichon I think this is still something we should go forward with. It lost traction since there was some confusion in the other ticket around if the gains were being properly applied to polynomial response stages. Mark suggested we might need to look changing the response of the BSU infrasound to use a normal (CoefficientsTypeResponseStage) instead.

salichon:
Will  be merged

PR #2162, opened 2024-08-01
NEZ replace borehold BB sensor + datalogger

Thomas-Benson:
Removing TC120PH-2 (6637) from `NEZ LC 12`
Replacing with TC120PH-2 (10258) at `NEZ LC 11`
 - Orientation unknown for now, will be updated once a surface sensor can be deployed
 

Removing Centaur (9902) from place `North Egmont Visitor Centre`
Replacing with Centaur (11443) at same place
 - Using latest firmware version 4.8.5
 

Closing LC 12, reopening LC 11 with broadband streams


CBurton90 (approved):

gva6500 (dismissed):
lgtm

mnaguit (commented):
Thanks @Thomas-Benson , is there  a need to open the connections for LC 11? and close LC 12? 🤔

Thomas-Benson:
Yes there is 🤦 Good catch @mnaguit

mnaguit (approved):
Thanks for the add-on for the connections @Thomas-Benson 👍 . Looks good, for now. 
A few side comments:

- sensor orientation for `LC 11` to be updated later? (although the now closed `LC 12` is axial, with 0 azimuth)
- there is now an overlap for `LC 11` and `LC 12` under sites. I'm not sure if it matters, but does not seem to be a clean metadata (such that `LC 11` was open since 2010 until 2020, without data from 2020 until `LC 12` is closed July 2024 ---meaning there is data gap for `LC 11` from 2020 to 2024).

Thomas-Benson:
@mnaguit yes we will update the orientation at a later date, but apparently there was no time to install a surface sensor at while making this swap, so there'll be a bit of a delay before we can get the new orientation

And yes there is an annoying overlap in the sites times. With the current implementation of Delta we can only have one entry per site in `network/sites.csv`, so I had to include times when we weren't actively occupying the site. There aren't any overlaps in the stream and sensor install files so this won't be evident in any downstream metadata products though.

CBurton90:
@Thomas-Benson @mnaguit We can estimate the orientation without a surface sensor (with ambient noise or an large teleseism) its just a bit more accurate to do it with one. I'll put it on the list to do!

CBurton90 (approved):

PR #2165, opened 2024-08-01
feat: Skeleton file generating for bnc-config

junghao:
- Added an optional switches for `bnc-config`
  -  `-skl {path}` to generate Skeleton files, based on the configured sites. 
  - `-summary {path-name}` to generate skeleton files summary so operators know how many files success/failed

- Also remove functionality to import aliases for bnc-config. ref https://github.com/GeoNet/tickets/issues/6495

Plan to run this from ac-tool, and generate SKL files into `puppet_gps_files/files/data/skl-{environment}` (or somewhere else, will decide in ac-tool). PR for ac-tool : https://github.com/GeoNet/ac-tools/pull/408

This is part of the steps towards https://github.com/GeoNet/tickets/issues/15788.

### DIFF
--- a/cmd/bnc-config/main.go
+++ b/cmd/bnc-config/main.go
@@ -6,6 +6,8 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"slices"
+	"strings"
 	"time"
 
 	"github.com/GeoNet/delta"
@@ -13,12 +15,13 @@ import (
 )
 
 type Settings struct {
-	base    string // delta base directory
-	common  string // ntrip common files directory
-	input   string // ntrip input files directory
-	extra   bool   // add aliases to mounts list
-	output  string // optional output file
-	sklPath string // optional path to write skeleton files
+	base        string   // delta base directory
+	common      string   // ntrip common files directory
+	input       string   // ntrip input files directory
+	extra       bool     // add aliases to mounts list
+	output      string   // optional output file
+	sklPath     string   // optional path to write skeleton files
+	skippingSkl []string // optional list of marks to skip skeleton file generation
 }
 
 func main() {
@@ -44,8 +47,14 @@ func main() {
 	flag.BoolVar(&settings.extra, "extra", false, "add aliases to mounts list")
 	flag.StringVar(&settings.output, "output", "", "optional output file")
 	flag.StringVar(&settings.sklPath, "skl", "", "optional path to write skeleton files")
+	var skippingSkl string
+	flag.StringVar(&skippingSkl, "skip-skl", "", "optional comma separated list of marks to skip skeleton file generation")
 
 	flag.Parse()
+
+	if skippingSkl != "" {
+		settings.skippingSkl = strings.Split(skippingSkl, ",")
+	}
 
 	// set recovers the delta tables
 	set, err := delta.NewBase(settings.base)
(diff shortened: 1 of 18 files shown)


cmd/bnc-config/skl.go
```diff
@@ -0,0 +1,154 @@
...
+G   15 C1C C1X C2W C2X C5X L1C L1X L2W L2X L5X S1C S1X S2W  SYS / # / OBS TYPES 
+       S2X S5X                                              SYS / # / OBS TYPES 
+E   15 C1X C5X C6X C7X C8I L1X L5X L6X L7X L8I S1X S5X S6X  SYS / # / OBS TYPES 
+       S7X S8I                                              SYS / # / OBS TYPES
+R   15 C1C C1P C2C C2P C3X L1C L1P L2C L2P L3X S1C S1P S2C  SYS / # / OBS TYPES 
+       S2P S3X                                              SYS / # / OBS TYPES
+C   15 C1X C2I C5X C6I C7I L1X L2I L5X L6I L7I S1X S2I S5X  SYS / # / OBS TYPES 
+       S6I S7I                                              SYS / # / OBS TYPES
+J   18 C1C C1X C1Z C2X C5X C6L L1C L1X L1Z L2X L5X L6L S1C  SYS / # / OBS TYPES 
+       S1X S1Z S2X S5X S6L                                  SYS / # / OBS TYPES
```
ozym, line 130: There seems to be some differences with the trailing spaces on some of these template lines. I assume they either should all line up, or all be trimmed.

ozym (dismissed):
It's all there and it looks like it will work.

Approving to get things moving, the comments I've made are probably more related to how the code fits into its running environment rather than the code itself.

The comments may be used to refine the code later.

cmd/bnc-config/skl.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/GeoNet/delta/meta"
+)
+
+var geo = ellipsoid.Init("WGS84", ellipsoid.Degrees, ellipsoid.Meter, ellipsoid.LongitudeIsSymmetric, ellipsoid.BearingIsSymmetric)
+
+// genetrate skeleton file for a site,
+// based on the receiver, firmware, and antenna metadata at a given time (e.g. now)
+func skeleton(code string, set *meta.Set, ts int64) (string, error) {
+	var mark meta.Mark
+	for _, m := range set.Marks() {
```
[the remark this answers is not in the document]
ozym: Or this loop could be lifted to the caller function, i.e.
for _, m := range set.Marks() {
  if !inWindow(ts,  mark.Start) {
    continue  
  }
  s, err := skeleton(set, m, ts) 
  ....
}

elidana (commented):
Thanks @junghao and @ozym !

I saw that and could not resist (and also added @aleks-spes as reviewer, as he can comment on some of the specifics of the skeleton format and content)

Also, I wonder if some of the modules use here are replicated? I know that for things like the header comment (GeoNet or LINZ), there is a similar one in the code that generates gloria protobuf for 30s RINEX headers?

more a curiosity than anything else!

junghao:
> Thanks @junghao and @ozym !
> 
> I saw that and could not resist (and also added @aleks-spes as reviewer, as he can comment on some of the specifics of the skeleton format and content)
> 
> Also, I wonder if some of the modules use here are replicated? I know that for things like the header comment (GeoNet or LINZ), there is a similar one in the code that generates gloria protobuf for 30s RINEX headers?
> 
> more a curiosity than anything else!

Yes I know the iterations for receivers/antennas also happens in `delta/tools/gloria` and `gloria/internal/rinex`, however, they have something slightly different and designing a unified interface doesn't seem to be necessary.

elidana:
ok sounds good! we'll just need to remember to keep things in sync where needed (i.e., that comment, that never changes :laughing: )

ozym (dismissed):

junghao:
@ozym I've pushed another commit due to the user's request, and here's the conclusion comment: https://github.com/GeoNet/tickets/issues/15788#issuecomment-2274455579

The user prefer to use generic header (no receiver's detail) instead of out-dated metadata for rinex 3 files. So the latest commit output a summary file of which sites using generic headers. Then from ac-tool's puppet-gps-file.sh, we include the content of summary file into the commit body pushed to puppet-gps-file. Hope this would work.

cmd/bnc-config/main.go
```diff
@@ -74,4 +78,34 @@ func main() {
...
+			err = os.WriteFile(filepath.Join(settings.sklPath, fmt.Sprintf("%s00NZL.SKL", m.Mark)), []byte(s), 0600)
+			if err != nil {
+				log.Fatalf("couldn't write skeleton file: %s", err)
+			}
+		}
+		if settings.summary != "" {
+			// output a summary info of skeleton generating
+			var output strings.Builder
+			output.WriteString("successes: " + fmt.Sprintln(successes))
+			output.WriteString("generic headers: " + fmt.Sprintln(fallbacks))
```
wilsonjord, line 105: suggestion(non-blocking): `strings.Join(successes, ", ")` might result in slightly more pleasant output.

cmd/bnc-config/skl.go
```diff
@@ -0,0 +1,190 @@
...
+// genetrate skeleton file for a site,
+// based on the receiver, firmware, and antenna metadata at a given time (e.g. now)
+func skeleton(code string, set *meta.Set, ts int64) (content string, err error) {
+	var mark meta.Mark
+
+	defer func() {
+		if err != nil {
+			// if any error occurs, compose generic header + distribution comment as the result
+			// however, still keep the error
+			content = generic_header
```
wilsonjord, line 21: nitpick: camel case?

cmd/bnc-config/skl.go
```diff
@@ -0,0 +1,190 @@
...
+		domesNumber = monument.DomesNumber
+	}
+
+	content = fmt.Sprintf(skeletonFormat,
+		fmt.Sprintf("%s00NZL", code),       // MARKER NAME
+		domesNumber,                        // MARKER NUMBER
+		dr.Serial, dr.Model, ifirm.Version, // REC # / TYPE / VERS
+		ia.Serial, ia.Model, radome, //ANT # / TYPE
+		x, y, z, //APPROX POSITION XYZ
+		ia.Offset.Vertical, ia.Offset.East, ia.Offset.North) // ANTENNA: DELTA H/E/N
```
wilsonjord, line 120: nitpick: I'm surprised the formatter didn't align the comments

wilsonjord (approved):
Looks good. A few minor things I don't really care about. Tests pass. I couldn't review the logic a huge amount, so just tried to focus on style and anything obviously weird.

cmd/bnc-config/skl.go
junghao: Yeah tried to align them manually. And it bounced back like this right after I clicked "save".

wilsonjord: huh weird. The online Go Playground doesn't do this (it aligns comments). Ah well.

wilsonjord (approved):
lgtm

ozym (approved):
Looks fine to me, worth an initial go.

PR #2171, opened 2024-08-06
Fix: XYZ labels for JSCB

mnaguit:
- end/start dates align with the recorder upgrade
  - end XYZ continuous streams on `2022-06-18T00:03:00Z`
  - start 12Z continuous streams on `2022-06-23T01:55:00Z`


salichon (approved):
sweet

PR #2173, opened 2024-08-07
poc: add GeoNet datasets DOI to delta

elidana:
Hi @ozym (and @staylorofford )

this is a very drafted proof of concept to try to capture into delta those DOIs that are associated with GeoNet core or "adopted" data.

It is not comprehensive of all DOIs we have assigned, but is a very first step towards potentially providing the DOI information alongside our data (i.e. via StationXML, GeodesyML, potentially tilde??).

An up to date list of all GeoNet datasets DOIs minted so far is available here: https://www.geonet.org.nz/policy

there is more behind all that, this pr was is just a very first step, and prompted by the SZNO use case (and some others that we discussed in the past).

### DIFF
--- a/network/networks.csv
+++ b/network/networks.csv
@@ -13,6 +13,7 @@ GN,GN,The GNS CGPS network,false
 HA,NZ,Hydrophone Array,false
 HB,NZ,Hawke's Bay regional seismic network,false
 IG,LI,LINZ IGS CGPS Network,false
+IU,NZ,Global Seismograph Network,false
 KI,NZ,Kermadec Islands volcano seismic network,false
 LG,NZ,Lake tsunami gauge network,false
 LI,LI,The LINZ national CGPS network,false
--- a/references/README.md
+++ b/references/README.md
@@ -7,6 +7,7 @@ _Reference citations and external source information._
 Reference information for the GeoNet sensor networks and manual data collection systems.
  
 * `citations.csv` - Reference citations
+* `datasets.csv` - Datasets Digital Object Identifiers
 * `tilde_methods.csv` - Tilde  method descriptions
 
 #### _CITATIONS_ ####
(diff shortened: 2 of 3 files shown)


ozym:
Looks okay I'll see what code needs to be updated, you need to change IU to have an external network of IU as well. That's what the data is recorded as.

(it would be easier if this was a PR from a clone of the repo rather than your fork)

e.g.

```
IU SNZO  00 BH1 D 2024-08-07 17:34:38  -  2024-08-07 20:55:46
IU SNZO  00 BH2 D 2024-08-07 17:34:37  -  2024-08-07 20:55:44
IU SNZO  00 BHZ D 2024-08-07 17:34:34  -  2024-08-07 20:55:42
IU SNZO  00 HH1 D 2024-08-07 17:34:37  -  2024-08-07 20:55:48
IU SNZO  00 HH2 D 2024-08-07 17:34:39  -  2024-08-07 20:55:49
IU SNZO  00 HHZ D 2024-08-07 17:34:39  -  2024-08-07 20:55:51
```

elidana:
thanks @ozym !! I was also wondering if we shall just use the "external code" (we will have same DOIs for all seismic waveforms data, for example...and I've been trying different options such as including domain or not (I'm tempted to include it, have done a couple of trials on that).

If you agree that external network code only is ok and enough, I can add that to the pr (and will redo that from a branch rather than my fork).

thanks for the super quick feedback!

elidana:
for the code space, would be great if we could add the DOI in the metadata or data headers (depending on domains).

- For waveform data, we could use the StationXML _identifier_ attribute DOI type (https://docs.fdsn.org/projects/stationxml/en/latest/reference.html#network-description)
- for GNSS data, we could use the DOI attribute in RINEX headers (and in future use the GeodesyML attribute) 

and so on...

ozym:
My comment is mostly that for external data there only needs to be an external code, whereas internal data has both. So the fact that internal is the same as external shouldn't matter too much.  Having internal codes means you can be a little more fine grained with your entries.  (e.g. NZ is used for weak motion / strong motion/ tsunami / geomag etc.).

elidana:
replaced by https://github.com/GeoNet/delta/pull/2175

PR #2174, opened 2024-08-07
WHTZ L4C-3D to TC120PH-2 SP to BB sensor upgrade

Thomas-Benson:
- Removing short-period L4C-3D (2826) from WHTZ LC 10
- Replacing with TC120PH-2 (10636) at WHTZ LC 13
  - New location coordinates -38.665838, 175.957742, 778m
- Updating stream codes from LC10 EH-ENZ to LC13 HH-ENZ + LC13 LH-ENZ


mnaguit (commented):
Thanks @Thomas-Benson , considering there are other LCs in this station prior to `LC 13`, is there a need to match the elevation in `sites.csv` and `stations.csv`?

CBurton90 (approved):

Thomas-Benson:
@mnaguit yes we probably should, there is a fair bit of discrepancy between the new and old measurements. Would it be preferable to go wth the new elevation, or the historical measreuemnt?

mnaguit:
Tricky question, but I think the most recent measurement is a better option.

Thomas-Benson:
Updated the stations entry, do we want to change the previous entries in sites.csv as well?

mnaguit:
any thoughts @salichon @CBurton90 @gva6500 @ozym ?

CBurton90 (approved):

CBurton90:
@Thomas-Benson @mnaguit There will be a fair bit of error in the vertical position from the internal gps but I think it makes sense to go with the latest value we have `778`.

Thomas-Benson:
For comparison Google earth puts the elevation of the site around 780, so the new value definitely makes more sense

mnaguit (approved):
Thanks @CBurton90 @Thomas-Benson , sounds good for the elevation (`778`) plus the update on sensor depth (`1.7`).

PR #2175, opened 2024-08-08
feat: add GeoNet datasets DOI and Tilde domains to delta

elidana:
This pr is an inital step to add DOIs that are associated with GeoNet core or "adopted" data, and introduce a link with "domains" as represented in the GeoNet's Tilde time series application.

It is hopefully a very first step towards providing the DOI information alongside our data when the domain standard allow for it (e.g. via StationXML, RINEX data or GeodesyML) or when the application and data dissemination can be expanded (e.g. Tilde), and to provide the correct DOI for stations distributed via GeoNet by operated by others (i.e. `SZNO`).

An up to date list of all GeoNet datasets DOIs minted so far is available here: https://www.geonet.org.nz/policy

This replace https://github.com/GeoNet/delta/pull/2173

-------
Some other details that need reviewing/improvements (mostly for datasets DOIs need to be reviewed, and still some open questions)
- [x] shall we add a reference to weather a derived time series is available via Tilde  or not?
- [x] GeoNet DOIs has some granularity for GNSS datasets, shall we rectify and how should we represent this here? 
- [x] does the geomagnetic dataset have a DOI? GNS Dataset catalogue entry is: https://data.gns.cri.nz/metadata/srv/eng/catalog.search#/metadata/56c42b24-b93d-46d1-8e7a-4e0a1051881f
- [x] is there a doi for RCET array?
- [ ] does Scandoas have a network code?? (doi is https://doi.org/10.21420/PSP7-KW60?x=y)


ozym:
One possibility is to investigate how we can better construct the citations table as that already has a DOI field.

One failure of the current system is that the link between "data entry" and "citation" is not explicit. An example in the
classes file:

```
Station,Site Class,Vs30,Vs30 Quality,Tsite,Tsite Method,Tsite Quality,Basement Depth,Depth Quality,Link,Citations,Notes
...
046A,D,210,Q3,0.7,Mw,Q2,50,Q3,,McVerry2011 Perrin2015 VanHoutte2017,some conflicts
...
```

where in fact there should be some form of linking the station to the references. This gets a little awkward if there is no "unique" field to latch onto (as otherwise you would need a table for each station, one for each mark, one for each instrument ...., or you would need to have a column for each possibility).  Needs more thinking.

elidana:
thanks for the comment Mark! have tried to reshape this a bit to add a possible link between citation, network codes, and domain

I have also tried to add a link to weather the dataset is available from tilde or not , not super happy with what I got here but perhaps we can progress this idea? Happy to rework anything in the best way for future programmatic access.

There are a bunch of details that will need to be reviewed, and some datasets have different level of granularity in how their DOIs were minted (sic....see for example GNSS datasets, for which we have two different DOIs for raw/rinex and timeseries data, my bad!)

elidana:
your wish is my command! 
have removed the external network code, and added a `domains.csv` that list all domains used in dataset but have a description only if the timeseries is in tiled. 

for the edge case....good point, we may add a mechanism for exceptions based on site code perhaps? (a clumsy one might be that in the dataset table we have an additional field with a space separated list of site codes...? need more thinking from my side)

salichon:
yo  @elidana why don t we simply also populate network DOIs in  the `install/network.csv`  Network description field  
with also the  short DOI  link as in `https://doi.org/10.21420/XQS0-0Z48`    ?
- This description field  is parsed  by  FDNS webservice stationXML for instance into the tags

```
> > <Network code="NZ" startDate="1916-01-01T00:00:00" restrictedStatus="open">
> > <Description>New Zealand National Seismograph Network+Add DOI  here</Description>
```

nimost likely  simple to be good enough.

elidana:
thanks for looking at this POC @salichon . The reason why I didn't go down the path of adding the DOI to the `network/networks.csv` is that I was hoping to start building a mechanism that could be used by different systems (not just FDSN, but also GNSS, multi-domain applications such as tilde, etc). I'm also thinking about derived data (we can potentially still be able to map those against network codes but it might be a bit more complex than 1-1 relations? )

There is also some growing need to be a bit more explicit on how we link sensors (and their network ) to tilde domain, so this is a "due piccioni con una fava" possible option. I'm planning to show this to the SOD team, so please keep your comments/thinking coming!

JonoHanson (dismissed):
LGTM

JonoHanson:
Not to stall this going through, but it would be great to include the GeoMag DOI.

Also, as I mentioned when Elisabetta presented it. This treats our current DOIs which are for the datasets as DOIs for the networks, merging those concepts a bit. Totally in favour of that.

elidana:
thanks @JonoHanson for the review!! really appreciate, and totally agree with the simplified approach we're taking of merging network and datasets DOIs (leveraging the parent-children concept in the dataset catalogue side of things).

For the geomag DOI, I believe there isn't one yet? At list that's what I get when looking at the related (if I'm picking the correct one) [GNS Dataset Catalogue entry for Geomag data](https://data.gns.cri.nz/metadata/srv/eng/catalog.search#/metadata/56c42b24-b93d-46d1-8e7a-4e0a1051881f)

@tanjapetersen would you be able to comment on that please?

tanjapetersen:
@elidana there are currently only DOIs for the individual data products (dataset children). 

I will organize you one for 
[GNS Dataset Catalogue entry for Geomag data](https://data.gns.cri.nz/metadata/srv/eng/catalog.search#/metadata/56c42b24-b93d-46d1-8e7a-4e0a1051881f)
as it does make sense to have one as the access point.

tanjapetersen:
@elidana Here it is :-)

https://doi.org/10.21420/5q60-k348

elidana:
"Citations" field in `datasets.csv` now changed to "Key" to be consistent with `citations.csv`
added geomag DOI as a placeholder, to be refined once correct DOI/GNS catalogue entry is ready

elidana:
I have now updated this with some recent comments and squashed all commits.

Can someone please do a final review so it can be merged? (@ozym or @JonoHanson or @rumachan ?) thanks!

ozym (approved):
This looks like a good start, we can refine it as needs or requirements evolve.

elidana:
thanks a lot @ozym !

PR #2178, opened 2024-08-08
labels for old cameras

rumachan:
@elidana and I agreed we'd retrospectively change the labels for cameras that are no longer used to be consistent with what we have done for current ones.

This affects one current camera "Tongariro from North" as earlier I mistakenly made the label change to an old camera.

Two cameras do not have "from" or "looking" in the label as the description is already clear and we shouldn't add text if it doesn't improve clarity:
- Raoul Island
- Tongariro Te Maari Crater

For old Christchurch cameras, as they were installed at a single site with views in different directions, "from" was not appropriate so I used "looking".


elidana (approved):
thanks @rumachan , looks good!

PR #2179, opened 2024-08-09
resurrect OT013

rumachan:
Resurrecting OT013 after earlier vandalism.


elidana (commented):
this one looks good.
I just wonder if we should add the gap in the metadata since majority of sensors/dataloggers that have been changed have just been reinstalled as is?

rumachan:
@elidana There no agreed procedure around data gaps such as this. We have never been in this situation before so I was winging it.
If the outage is short, a few days to weeks, then no, a gap doesn't add much value to metadata.
If a gap is long, many months to years???, then I think its obvious there should be a gap.

In this case the gap was about 6 weeks, a grey area.
If I had know @ricardojohansson was going to reinstall so soon I wouldn't have bothered closing the site in the first place.
@elidana Please decide if you want a gap or not. I'll make the metadata match that.

elidana:
thanks @rumachan , yes I understand and sorry to be a pain.
I know of examples from other domains with data gaps that are longer than the period you have here, but those "outages" are not represented in delta.
So, acknowledging that we don't have a good way to represent station outages (whatever their length), I would suggest to keep the record as continuous in the metadata

rumachan:
Got it, will do

elidana:
hi @rumachan , this now has some conflicts that needs resolving. Let me know if you'd like me to contribute and address those, or ping me when ready for review! thank you!!

rumachan:
@elidana I don't understand how that happened but I removed the incorrect part of the file it said had conflicts. Says it has now passed checks. Thanks

network/stations.csv
```diff
@@ -1583,8 +1583,8 @@ ORCS,SM,Ohakune Ruapehu College,-39.417438387,175.412599486,606,,WGS84,2001-11-2
 ORE,XX,Oreore,-40.05956403,176.373189007,360,,WGS84,1990-05-15T03:55:00Z,1990-05-17T05:08:00Z
 ORER,XX,O'Regan Road RATS,-37.736921601,178.249720016,180,,WGS84,1994-08-04T05:56:00Z,1994-12-14T23:17:00Z
 ORPS,SM,Orepuki,-46.286797,167.76512,71,,WGS84,2016-04-18T00:00:00Z,9999-01-01T00:00:00Z
-OT001,EN,Waimangu Inferno Crater,-38.2819,176.3997,380,,WGS84,2017-11-07T00:00:00Z,9999-01-01T00:00:00Z
-OT013,EN,Rotoma Soda Spring,-38.03852307,176.54950786,300,,WGS84,2021-03-17T00:00:00Z,2024-07-05T14:50:01Z
+OT001,EN,Waimangu Inferno Crater,-38.2819,176.3997,0380,,WGS84,2017-11-07T00:00:00Z,9999-01-01T00:00:00Z
```
elidana, line 1586: is the leading 0 a typo here?

elidana (approved):
Hi @rumachan ! This looks good

just one minor comment in the station height (`0380` instead of `380`, suspect is a typo?) in `network/stations.csv` but looks good!!
Thanks for all the changes in this one!

PR #2184, opened 2024-08-12
DUNT - fix radome timestamp

aleks-spes:
Fix the start timestamp  in `radomes.csv` to align it with the start timestamp in the `antennas.csv` file


elidana (approved):
aaah good catch! gosh we shall add some tests for that stuff...

aleks-spes:
Thank you @elidana !

PR #2185, opened 2024-08-13
SNZO adoption add site metadata

Thomas-Benson:
Adding current metadata for newly adopted GSN site `SNZO - South Karori`
 - LC 00:
   -  Streckeisen STS-6A VBB (SN: `SNZO_BB_202408` - placeholder)
   - Azimuth 107, depth 91m
   - Streams HH-12Z + BH-12Z
 - LC 20:
   -  Kinemetrics,FBA-ES-T (SN: `SNZO_SM_202408` - placeholder)
   - Azimuth 0, depth 0m
   - Streams HH-12Z + BH-12Z
 - Datalogger:
   - Kinemetrics Q330HR/6 (SN `4226`)

USGS metadata is mostly available from: http://ds.iris.edu/mda/IU/SNZO/
There is no NRL response for the new STS-6A that I could find, currently using a reformatted version of the response supplied by IRIS FDSN:
https://service.iris.edu/fdsnws/station/1/query?net=IU&sta=SNZO&loc=00&cha=HH?&starttime=2024-08-01T00:00:00&level=response&format=xml

TODO:
 - ~Add sensor serial numbers~
 - check response is correct for new sensor
 - check stationxml-build handling of external networks
   - I can build the stationxml if I add `IU` to the external network flag manually, but it's not currently built by default


ozym:
```
{106} Filters & delay 20-HN2:6@200,Linear below 100sps=0.015462
{123} Station: IU-SNZO, quant2dali
{600} Q330 Serial Number: 0100001xxxxxxxxxx
{600} AMB Serial Number: 0100001xxxxxxxxxx
{600} Seismo 1 Serial Number: 0000000000000000
{600} Seismo 2 Serial Number: 0000000000000000
{600} QAPCHP 1 Serial Number: 4881496
{600} QAPCHP 2 Serial Number: 401631
{600} KMI Property Tag Number: 4226
{600} System Software Version: 1.150

```

Thomas-Benson:
Just checking @ozym, it looks like we use the `KMI Property Tag Number` as the serial number for Q330s?

ozym:
@Thomas-Benson yes, I'm pretty sure we do as this is something on the outside of the box which looks like a serial number (as opposed to the internal serial number (what we call the long number) which is used for connections and is quite cumbersome.

assets/dataloggers.csv
```diff
@@ -374,6 +374,7 @@ Kinemetrics,Q330HR/6,4004,13957,
 Kinemetrics,Q330HR/6,4008,13958,
 Kinemetrics,Q330HR/6,4009,13959,
 Kinemetrics,Q330HR/6,4010,13960,
+Kinemetrics,Q330HR/6,4226,,
```
salichon, line 377: We shall clearly indiacte that this is not a GeoNet asset 
like
  "IU: Global Seismograph Network (GSN - IRIS/USGS) and https://doi.org/10.7914/SN/IU"

assets/sensors.csv
```diff
@@ -471,6 +471,7 @@ Intech,Platinum resistance thermometer,004,,
 Kinemetrics,FBA-23,25073,,
 Kinemetrics,FBA-23,45577,,
 Kinemetrics,FBA-23,WEL_WLN,,
+Kinemetrics,FBA-ES-T,SNZO_SM,,
```
salichon, line 474: Same as above comment for the datalogger

assets/sensors.csv
```diff
@@ -1539,6 +1540,7 @@ Streckeisen,STS-2,30203,9207,
 Streckeisen,STS-2,30204,9209,
 Streckeisen,STS-2,60732,12994,
 Streckeisen,STS-2,TMP_384,,
+Streckeisen,STS-6A VBB,SNZO_BB,,
```
salichon, line 1543: same same

network/networks.csv
```diff
@@ -13,6 +13,7 @@ GN,GN,The GNS CGPS network,false
 HA,NZ,Hydrophone Array,false
 HB,NZ,Hawke's Bay regional seismic network,false
 IG,LI,LINZ IGS CGPS Network,false
+IU,IU,Global Seismograph Network,false
```
salichon, line 16: would thsat be relavant to add  the IU/GSN DOIt in the description ? 
https://doi.org/10.7914/SN/IU

salichon (commented):
Perhaps also the reference on the assets shall contain a form of date
like 202408  ?

network/networks.csv
Thomas-Benson: @salichon adding DOIs to delta is being worked on here: https://github.com/GeoNet/delta/pull/2175

install/streams.csv
```diff
@@ -2122,6 +2122,9 @@ SNGZ,10,E,H,100,false,false,false,2009-06-08T03:00:00Z,9999-01-01T00:00:00Z
 SNVZ,10,E,H,100,false,false,false,2015-03-27T02:00:00Z,9999-01-01T00:00:00Z
 SNVZ,30,L,D,1,false,false,false,2020-02-20T04:50:00Z,9999-01-01T00:00:00Z
 SNVZ,30,H,D,100,false,false,false,2020-02-20T04:50:00Z,9999-01-01T00:00:00Z
+SNZO,00,B,H,40,true,false,false,2024-08-05T08:54:00Z,9999-01-01T00:00:00Z
```
CBurton90, line 2125: I know BH streams are available but I thought the intention was to ignore them and only collect HH. One to double check, I'll re-read the ticket in case the scope has changed.

salichon: BH streams are strored presently
we may keep it like this  as we re using - at the moment - BH for the Non-NZ network stations ( practicla resons)

network/networks.csv
salichon: Okay so i think this would be ann easy way to set ibn 
since the descritpion of the network  is provided with the stationxml

Thomas-Benson:
Changed the placeholder serial numbers for the sensor to `SNZO_BB_202408` and `SNZO_SM_202408`
Also added references to GSN to the assets, although the wording/formatting might need some tweaks

install/streams.csv
CBurton90: I mean the purpose of this process was to adopt SNZO as a national network site and treat it as such. That would mean it would be the only national network site supplying BH streams but it seems we are ok with that? Were we not only collecting the BH streams historically because HH was not available (and now it has been turned on for us)? I cant find any reference to the sampling rate requirements in existing tickets/programme decision making template but found this comment from Sam:

https://github.com/GeoNet/tickets/issues/16239#issuecomment-2297946122
"data are available at the equivalent sampling rate"

any thoughts? @ozym @staylorofford

ozym: The current processing system uses BH?.  Until that switches over we will continue to recover BH? and it's easier simply to store what comes in. We could clearly backfill this if needed or wanted.  Once we're switched over to HH? completely then the question could perhaps be asked again (i.e. there will be no picks being made on the SNZO BH? streams).

salichon (approved):
Hi @Thomas-Benson 
this is looking good 
Is the station.csv entry for later ? 

Also i would be in favor of adding a documentation into 
https://github.com/GeoNet/delta/tree/main/docs/stations
With regard to the specifity of this station into Delta  and how it is different from the IRIS metadata.

What we do now shall be forgotten or almost in 3 years of time ...and it ll be easier to answer questions from users then.

Thomas-Benson:
[UTU](https://trello.com/c/7nnpJMVR/3-utu)

CBurton90 (approved):

Thomas-Benson:
I beleive SOD is happy with this now, are we good from a platform perspective to merge this @ozym? I know there's a bunch of downstream stuff that'll need to be sorted afterwards.

ozym (approved):
Great, lets get this merged in and then work on the downstream updates needed.

PR #2202, opened 2024-08-28
Fix XYZ streams for AVAB

mnaguit:
**Background**
- this structural array has mixed channel streams (`XYZ` and `12Z`)
- due to this, the instrument response is not available for all streams with XYZ in the data filename

**How to fix** 
- identify the time to switch on/switch off the XYZ streams (e.g. check historical data in the archive) 
- correct the metadata 
- check if data can be corrected to acceleration


salichon (approved):
sweet as ! 
great :) 
ta mumu

PR #2210, opened 2024-09-04
EAZ Q330 to Centaur datalogger swap

Thomas-Benson:
- Removing Q330S/6 (5442) from Earnscleugh
 - Replacing with Centuar (9543) at same place
 - Using Centaur firmware version 4.8.5


CBurton90 (approved):

salichon (approved):
sweet x3 thanks !

PR #2211, opened 2024-09-04
LIRZ L4C-3D to TC120PH-2 SP to BB sensor upgrade

Thomas-Benson:
- Remove SP L4C -3D (2987) from LIRZ LC 10
 - Replace with BB TC120PH-2 (10690) at same location
 - Stream codes change from 10.EH? to 10.HH? + 10.LH?


CBurton90 (approved):

salichon (approved):
sweet thx!

PR #2212, opened 2024-09-05
MKBS Basalt to TitanSMA SM recorder swap

Thomas-Benson:
- Removing Basalt(1682) from MKBS LC 21
 - Replacing with TitanSMA (002349) at same location
 - Using TitanSMA firmware version 4.5.22


CBurton90 (approved):

salichon (approved):
sweet as

PR #2213, opened 2024-09-05
feat: add SNZO to sensor config

ozym:
This adds the SNZO non-standard settings to the sensor config code


jajera (approved):
looks fine to me

PR #2218, opened 2024-09-10
MCHZ Q330 to centaur datalogger swap

Thomas-Benson:
- Removing Q33o/3 (5166) from `McNeil Hill - MCHZ`
 - Replacing with Centaur (10972) at same place
 - Using centaur firmware version 4.8.5


mnaguit (approved):

gva6500 (approved):
lgtm

PR #2224, opened 2024-09-16
fix: BNC-Client: SCTB skeleton file has incorrect country code

junghao:
As title. Existing process is using NZL as SCTB's country code. Should be ATA instead, and the country code information is already in metadata (https://github.com/GeoNet/dmc-config/blob/main/config/gps-files/staging/mounts.csv).

This PR uses the country field in the metadata, instead of hard-coding NZL.


wilsonjord (approved):
looks good. No changes to end user API. Should be no breaking code, unless user relied on previous incorrect behavior from the app.

jajera:
@junghao, any live system that depends on this code?

junghao:
@jajera No, these are for the staging bnc.

jajera (approved):
looks good and safe. no live services that will be affected yet.

PR #2226, opened 2024-09-16
feat: add dataset table

ozym:
This adds a simple domain and network lookup table, it also allows for a reference citation and whether it is used in tilde


elidana (approved):
looks good, thanks @ozym , that's awesome!! I haven't fully reviewed the code, so would be great if someone from @GeoNet/ssd-team can also review

meta/dataset.go
```diff
@@ -0,0 +1,112 @@
...
+	"Tilde":   datasetTilde,
+}
+
+type Dataset struct {
+	Domain  string
+	Network string
+	Key     string
+	Tilde   bool
+
+	tilde string
```
wilsonjord, line 30: question: assume this is because when decoding, you want to produce the exact same input, i.e. "True" -> "True", "TRUE" -> "TRUE" etc.?

wilsonjord: It's a shame Go doesn't allow for optional parentheses for no argument functions, otherwise you could make `Tilde` be a function that returns bool based on the value of `tilde`. But since it is required, making it into a function is a little ugly, as you'd be calling `d.Tilde()`, and other values `d.Domain`. You do get to remove a mutable variable though.
Still, I'm happy with this, thanks for the context.

wilsonjord (approved):

elidana (approved):

jajera (approved):
lets do it.

PR #2230, opened 2024-09-20
Fix XYZ streams for CPXB

mnaguit:
- mixed streams (`XYZ` and `12Z`) for `CPXB`
- switch-on/off time-date identified to fix this & make the instrument response metadata available for both streams


salichon (approved):
sweet as !

PR #2232, opened 2024-09-20
WAKC Basalt to TitanSMA SM recorder swap

Thomas-Benson:
- Removing Basalt (1762) from WAKC LC 20
 - Replacing with TitanSMA (002234) at same location
 - Recorder azimuth changed from `76` to `245` degrees
 - Using TitanSMA firmware version 4.5.22


mnaguit (approved):
Quite a change in sensor bearing, otherwise LGTM. Thanks @Thomas-Benson !

CBurton90 (approved):

PR #2234, opened 2024-09-23
feat: bnc skeleton file - marker name required in generic header

junghao:
As title. 

As requestd by product owner.


wilsonjord (approved):
the change of `genericHeader` to a format string looks consistent with the skeleton file.

jajera (approved):
looks good to me syntax wise.

PR #2240, opened 2024-09-25
KHLS TitanSMA SM recorder replacement

Thomas-Benson:
- Removing TitanSMA (002184) from KHLS LC 20
 - Replacing with TitanSMA (001914) at same location
 - Reopening firmware for TitanSMA 001914 with version 4.5.22
   (this sensor has been in use before, but was removed for repairs)


mnaguit (approved):
Thanks @Thomas-Benson, LGTM noting no change in location and azimuth. 👍

CBurton90 (approved):

PR #2242, opened 2024-09-26
MLnz20 Original calibration - table and reference

salichon:
Provisional entry for magnitude calibration (MLnz20)


salichon:
WIP - will update with the station  correction

salichon:
Need an entry in README

environment/README.md
```diff
@@ -16,6 +18,19 @@ _Lists the geological and physical environment details of collection points._
 
 * `visibility.csv` - Sky View Descriptions.
 
+#### _CALIBRATIONS_ ####
+
+| Field | Description |
+| --- | --- |
+| _Station_ | Station Code Name  
+| _Location_ | Sensor Location (e.g. Borehole or surface sensor)
```
staylorofford, line 26: MLNZ20 doesn't use location codes, it only has the correction at the station level. It's possible this is an oversight and we may be able to test this later with borehole installations which would make the loc column necessary. When SeisComP pulls data from this table, can it ignore loc column for MLNZ20 reflecting how its corrections are defined now?

staylorofford (approved):
Barring one comment to consider and assuming the corrections are pulled straight from the Table S3.csv, LGTM.

elidana:
just a comment as I bumped into this. In delta, we already have a `calibration` folder, that is used to upload instrument-specific laboratory calibrations (currently for GNSS antennas and fluxgates).
I wonder if we should use a different terminology for site-specific magnitude calibration factors?

staylorofford:
Good call @elidana. They are not calibrations in a true sense but corrections used to improve magnitude scale model fit. Can we just use `corrections` terminology @salichon ?

salichon:
@staylorofford  Okay 
I ll use the same terminology as the paper

staylorofford (approved):

ozym:
This is probably not how I would have laid out this file.

I'd think:

(a) Is the notes field really necessary?  I'd much prefer a "lookup table" if there was something to say, free form fields are hard to validate, and is it needed.
(b) I tend to want to put the dates at the end of the list of columns (the notes tend to end up there due to external forces).
(c) The citations should ideally be done in alphabetical order, which is deterministic (e.g. A2023 B2024 vs B2024 A2023 is the same but different).

In the past there wasn't really any concept of a location code when it came to station corrections, do you therefore need to keep track of all variations, and magnitude differences related to boreholes would be very hard to isolate.  Jury is out on this one, as it really limits when these corrections can be applied.

salichon:
Hello @ozym !
thanks :)
fair enough, will keep working on that one indeed, 
FYI @staylorofford 
and Follow up on the ticket

PR #2243, opened 2024-09-29
fix: switch "Short Period Borehole Seismometer" to "Short Period Seismometer"

ozym:
This should allow Duke sensors on the web site


elidana:
hi @ozym , this one looks good.

I don't think those are in use anymore, but that type is also used in:
```
resp/auto.go:		Type:         "Short Period Borehole Seismometer",
resp/auto.go:		Description:  "Duke University 2 Short Period Borehole Seismometer",
resp/auto.go:		Type:         "Short Period Borehole Seismometer",
resp/auto.go:		Description:  "Duke University 4.5 Short Period Borehole Seismometer",
```

should the auto.go also be updated or can we just ignore?

elidana (approved):

elidana (commented):
this looks good, but have one question about the response yaml that might also need updating?

elidana (commented):
I pushed approve button too early, repeating main comment here to make github happy

Shall the resp/responses/sensors/duke.yaml also be updated in those two entries?
```
resp/responses/sensors/duke.yaml:    type: Short Period Borehole Seismometer
resp/responses/sensors/duke.yaml:    type: Short Period Borehole Seismometer
```

ozym:
That is the old system and is no longer used, I really should put in a PR to remove it completely.

elidana:
ooh ok, so the fact they're obsolete is valid for both(augo.go and yaml files)
...ok, sorry for the confusion. 
And yes, would be good to tidy up anything that is not used!

elidana (approved):
looks good

ozym:
@elidana https://github.com/GeoNet/delta/pull/2248

PR #2248, opened 2024-09-30
feat: remove legacy response code and update README

ozym:
This is aimed at adding clarity by removing the legacy code that is no longer used for building response files


resp/README.md
```diff
@@ -4,203 +4,23 @@ The goal of the `resp` directory is to build a `golang` source module that allow
...
 
-### generate
+The location of the base XML files can be found in:
 
-The code needed to build the `auto.go` file using the raw configuration files can be found in the
-`generate` sub-directory. This code is usually run in the main `resp` directory via a call to
-`go generate`.
+- `auto` -- where the original response files have been converted into an XML equivalent.
+- `files` -- where response files have been handcrafted as needed.
+- `nrl` -- where responses derived from the NRL (notional response library) can be stored.
```
elidana, line 17: minor comment, I think the name is
"Nominal Response Library"

elidana (approved):
For the little I know, this one looks good to me (see just one minor comment for the Readme file).
Would be good to have another thumbsup, at least from @salichon or @Thomas-Benson 

I have locally tested the station xml build with the following steps:
- build local stationXML file: `cd ~/cmd/stationxml-build; go build; ./stationxml-build -station WTAZ -location 13 -version 1 > wtaz.xml`
- retrieve stationXML file from FDSN: `curl -o tmp 'https://service.geonet.org.nz/fdsnws/station/1/query?network=NZ&station=WTAZ&loc=13&level=response'; xmllint --format tmp > wtax-fdsn.xml`
- compare the two outputs

comparison looks good beside some optional(?) fields that might be related to the GeoNet FDSN station service implementation (??)

image

resp/README.md
ozym: good catch, have updated.

The actual XML is generated by other code so you're checking the difference between the two services and how the decode/re-encode the XML (some optional features have been dropped as the were actually removed in later versions).

CBurton90 (approved):

PR #2249, opened 2024-09-30
fix XYZ streams for WHSB

mnaguit:
- modify switch-on/switch-off times for XYZ streams


salichon (approved):
yup ! 
kia ora @mnaguit 
cheers!

PR #2262, opened 2024-10-22
add gnss domain to tilde config

wilsonjord:


ozym (approved):
nice a clean, all good

cmd/tilde-config/gnss.go
```diff
@@ -0,0 +1,34 @@
...
+			Latitude:    toFloat(fmt.Sprintf("%0.4f", m.Latitude)),
+			Longitude:   toFloat(fmt.Sprintf("%0.4f", m.Longitude)),
+			Elevation:   toFloat(fmt.Sprintf("%0.0f", math.Round(m.Elevation))),
+		})
+	}
+
+	// update domains
+	t.Domains = append(t.Domains, Domain{
+		Name:        "gnss",
+		Description: "Global Navigation Satellite System Sensors",
```
aleks-spes, line 29: Hey @wilsonjord ! After a discussion with @elidana  and @rumachan I still would like to propose changes to the domain description and go with `"Global Navigation Satellite System Stations"`. Sorry for the slow review!

elidana: "station" is the only terminology that is not used in tilde domain descriptions (see https://tilde.geonet.org.nz/v4/dataSummary/) and in the domain model it refers to the generalised "location" where time series data are available (can be from multiple `sensors` for some domains). So maybe let's just stick to `Global Navigation Satellite System` as a compromise?

wilsonjord: @elidana @aleks-spes could I get a final confirmation on the last question from @elidana ?

elidana: i personally vote for `Global Navigation Satellite System` as a compromise

aleks-spes: It seems there are a lot things behind the Tilde terminology and models that I don't know so let's go with the compromise option: `Global Navigation Satellite System`

aleks-spes (approved):
Looks good

ozym (approved):

PR #2269, opened 2024-10-25
TRVZ repairs

Thomas-Benson:
Replacing damaged sensor and datalogger at TRVZ LC 10
 - Removing sensor Guralp CMG-3ESPC (T36196)
 - Replacing with TC120PH-2 (10439)
 - Removing datalogger Q330S/6 (6060)
 - Replacing with Centaur CTR4-6S (9905)
   - Firmware version 4.8.5
 - Streams affected TRVZ.10.HH-ENZ, TRVZ.10.LH-ENZ (closed) and infrasound TRVZ.30.HDF + TRVZ.30.LDF from the datalogger

After investigation, the cabling for the old sensor was found to have a reversed E component, so a correction has been applied for the whole time period this sensor was installed.


Thomas-Benson:
Doublechecking,  it looks like the polarity is applied correctly:
```
...
<Channel code="HHE" startDate="2009-04-14T22:00:00" endDate="2012-05-04T00:00:00" restrictedStatus="open" locationCode="10">
        <Comment id="1">
          <Value>Location estimated from internal GPS clock</Value>
        </Comment>
        <Comment id="2">
          <Value>Location is given in WGS84</Value>
        </Comment>
        <Comment id="3">
          <Value>Sensor orientation not known</Value>
        </Comment>
        <Latitude datum="WGS84">-39.298816259</Latitude>
        <Longitude datum="WGS84">175.547821602</Longitude>
        <Elevation>2062</Elevation>
        <Depth>0</Depth>
        <Azimuth>270</Azimuth>
...
```

mnaguit (approved):
Thanks @Thomas-Benson , I had to double check the time period for Guralp for the polarity correction. Looks like the Guralp installed in 2009 + 2012 need a fix. All good.

salichon (commented):
@thomas 
should we consider to populate the fields
"Method, Citation" for reference ?
thanks!
jerome

install/streams.csv
```diff
@@ -2559,7 +2559,7 @@ TROB,28,H,N,200,XYZ,false,true,2013-08-02T00:00:00Z,2019-12-16T00:00:00Z
 TROB,28,H,N,200,XYZ,false,false,2019-12-16T00:00:01Z,9999-01-01T00:00:00Z
 TRTS,20,B,N,50,true,false,false,2012-02-10T00:00:10Z,2020-06-16T00:00:00Z
 TRTS,20,H,N,200,true,false,true,2012-02-10T00:00:10Z,9999-01-01T00:00:00Z
-TRVZ,10,L,H,1,false,false,false,2009-04-14T22:00:00Z,9999-01-01T00:00:00Z
+TRVZ,10,L,H,1,false,false,false,2009-04-14T22:00:00Z,2024-09-23T21:34:00Z
```
CBurton90, line 2562: Are you closing the LH* (1Hz) streams? I think we should keep them open no?

salichon: @CBurton90 the LH is  not configured   on the datalogger (from the data point of view) 
no data since 2024.267 
FYI @Thomas-Benson

salichon (changes requested):
When LH is set as for any BB 
I reckon no data backfil is required.

ozym:
> the LH is not configured on the datalogger (from the data point of view)

That is a configuration error and not a design choice, the streams should still be added though (but you could argue the start date should be closer to the actual time this feed starts again but it isn't needed).

ozym:
What I'm staying is that it is probably easiest to remove the change above and leave the TRVZ L line as is and treat it like an outage.

CBurton90:
Thanks @salichon @ozym I misinterpreted the first plot here https://github.com/GeoNet/tickets/issues/16888#issuecomment-2436142351 (clearly the LH streams have stopped). I've re-enabled them as of:

`Thu Oct 31 22:44:56 UTC 2024`


cc @Thomas-Benson

Thomas-Benson:
I've removed the changes to LH streams, now they are left open like the HH streams. (just with a longer break in the data)

salichon (approved):
tiptop

CBurton90 (approved):

PR #2275, opened 2024-11-03
feat: add sqlite database generation

ozym:
This adds the ability to build a standalone sqlite database which can be used as desired in third party applications. It also adds an example limited REST API server as a technology demonstation.  This has been made possible by using the pure go module `"modernc.org/sqlite"`.  This does some interesting automatic conversion of c-code to go to build a version that doesn't require `cgo`.  

e.g.
```
./deltadb -init -db delta.db -debug
2024/11/03 20:49:20 initialise database
2024/11/03 20:49:22 database initialised in 2.179325139s
```
and then
```
 sqlite3  delta.db
SQLite version 3.46.1 2024-08-13 09:16:08
Enter ".help" for usage hints.
sqlite> .schema Network
CREATE TABLE Network (
        Code TEXT PRIMARY KEY,
        External TEXT,
        Description TEXT,
        Restricted TEXT
);
sqlite>
```

It uses a combination of helper structures (named Table in the code) and the ability to reuse the current CSV files (these are massaged to be the same as SQL INSERT commands). The sqlite database is the only one to be coded, but the wrapper is an interface so coding other flavours should still be possible.

As an example application I have also coded up a REST API which can be run from memory (or off of a database file), e.g.

```
./deltadb -debug -listen
2024/11/03 20:55:11 initialise database
2024/11/03 20:55:12 database initialised in 1.181009504s
2024/11/03 20:55:12 handling requests on :8080
```

and then the service can be accessed via curl, e.g.

```
curl -s localhost:8080/network|jq .|head
[
  {
    "network": "AK",
    "external": "NZ",
    "description": "Auckland volcano seismic network"
  },
  {
    "network": "BC",
    "external": "BC",
    "description": "Building monitoring camera network"
  ...
```
or

```
curl -s localhost:8080/station/WEL/site/20/sensor/|jq . | head
[
  {
    "make": "Kinemetrics",
    "model": "FBA-23",
    "serial": "25073",
    "start": "1990-10-03T03:04:00Z",
    "end": "2003-04-08T00:00:00Z",
    "dip": 0,
    "azimuth": 0,
    "factor": 0,
```


wilsonjord:
Looks interesting. Assume it has various constraints defined? Can it be used for error checking the CSVs? Seems like the DB only takes 2 seconds to be created...

wilsonjord:
> Yes, to some degree. I have encoded a time overlap trigger which has caught at least one table (dart, NZC). There's also the inherent uniqueness checks when applying primary keys.

Nice. I recall a time when I was trying to figure out an issue, and eventually traced it back to a delta data issue (time span was incorrect for a particular resource), and remember thinking at the time "a database wouldn't have even let it through". Anyway, will give the PR a review.

meta/sqlite/sql.go
```diff
@@ -0,0 +1,71 @@
...
+
+func Serial(v any) QueryOpt {
+	return Option("Serial", v)
+}
+
+func Sample(v any) QueryOpt {
+	return Option("Sample", v)
+}
+
+func ParseBool(str string) (bool, bool) {
```
wilsonjord, line 62: suggestion: could this go into an internal package? `meta/mark` uses almost identical parsing when decoding.
I also think that the function name should either change to `IgsParseBool`, or the function name stays the same, and you incorporate `strconv.ParseBool` (or add "1" "T" etc. directly), to justify the name - essentially, provide a better `ParseBool` function than `strconv.ParseBool`, and use that everywhere.

wilsonjord (commented):
Just a couple of small comments. Another question, will this be published on s3, like the protobufs?

meta/sqlite/sql.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (q QueryOpt) V() any {
+	_, v := q(0)
+	return v
+}
+
+func Option(k, v any) QueryOpt {
+	return func(n int) (string, any) {
+		return fmt.Sprintf(" %s = $%d", k, n+1), v
```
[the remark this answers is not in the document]
ozym: Probably me just looking at the examples and seeing they all start at $1.

I did a quick test and was getting this error:

> failed to execute query missing named argument "0"

so it may not be universal within the go driver to use $0.

meta/sqlite/sql.go
ozym: good point, I'll have a think.  I'd like to use a standard true/false but we have a couple of times where it's true/false/maybe type settings.

ozym: Have gone for the extend the function as above, to make it the same as the strconv version (i.e. add 0,1,t,f etc).

meta/sqlite/sql.go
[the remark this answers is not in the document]
wilsonjord: Ah ok thanks for checking (sqlite3 cli allows `.param set $0 foo`). No problem, starting at `$1` is good (I've always used `:key` type notation, so wasn't clear on the numbered notation, assume it has significance when binding values all in one call).

wilsonjord (commented):
Two concerns, one doesn't really matter, the other one might.

The one that doesn't matter: I'm not sure decrease in readability to produce a "pretty" schema string is worth it...some clients can give you that functionality (for `sqlite3`, you can use `.schema --indent`), some clients ignore it (I sometimes use DB Browser for SQLite, and it just inlines everything). Another option (aside from "don't bother"), is to use a sql formatting lib to transform the final string.

The one that might matter, is the treatment of the `Start` and `End` fields.
In the schema, these are `DATETYPE`, but I believe that's simply treated as a user-defined field, and doesn't involve any type checking - they are just text fields.
So for comparisons involve these fields, what's being using is a straight text comparison:
```sql
                AND "Start" <= NEW."End"
                AND "End" >  NEW."Start"
```
With such comparisons, `'1999-09-27T00:00:00'` does not equal `'1999-09-27'`, whereas `datetime('1999-09-27T00:00:00')` does equal `datetime('1999-09-27')`.

One simply thing would be to enforce the time format on schema, something like:
```
Start DATETYPE CHECK(test for correct date format)
```
Then text comparisons become totally fine.

ozym:
Updated the tabbing to use spaces, adjusted line breaks and removed the quotes which so far are not needed.

```
sqlite>
sqlite> .schema Timing
CREATE TABLE Timing(
  Station TEXT,
  Location TEXT,
  Correction TEXT,
  Start DATETIME,
  End DATETIME,
  PRIMARY KEY(Station,Location,Start),
  FOREIGN KEY(Station,Location) REFERENCES Site (Station,Location)
);
CREATE TRIGGER NoOverlapOnTiming BEFORE INSERT ON Timing WHEN EXISTS (
  SELECT * FROM Timing
    WHERE Station == NEW.Station
    AND Location == NEW.Location
    AND Start <= NEW.End
    AND End >  NEW.Start
)

BEGIN
SELECT RAISE(FAIL, "Overlapping Intervals on Timing");
END;
```

vs

```
sqlite> .schema -indent Timing
CREATE TABLE Timing(
  Station TEXT,
  Location TEXT,
  Correction TEXT,
  Start DATETIME,
  End DATETIME,
  PRIMARY KEY(Station,Location,Start),
  FOREIGN KEY(Station,Location) REFERENCES Site(Station,Location)
);
CREATE TRIGGER NoOverlapOnTiming BEFORE INSERT ON Timing WHEN EXISTS (
  SELECT * FROM Timing
    WHERE Station == NEW.Station
    AND Location == NEW.Location
    AND Start <= NEW.End
    AND End >  NEW.Start
)

BEGIN
SELECT RAISE(FAIL, "Overlapping Intervals on Timing");
END;

```

ozym:
Have now updated the datetime settings, e.g.

```
sqlite> .schema Timing
CREATE TABLE Timing(
  Station TEXT,
  Location TEXT,
  Correction TEXT,
  Start DATETIME CHECK (Start IS strftime('%Y-%m-%dT%H:%M:%SZ', Start)),
  End DATETIME CHECK (End IS strftime('%Y-%m-%dT%H:%M:%SZ', End)),
  PRIMARY KEY(Station,Location,Start),
  FOREIGN KEY(Station,Location) REFERENCES Site (Station,Location)
);
CREATE TRIGGER NoOverlapOnTiming BEFORE INSERT ON Timing WHEN EXISTS (
  SELECT * FROM Timing
    WHERE Station == NEW.Station
    AND Location == NEW.Location
    AND datetime(Start) <= datetime(NEW.End)
    AND datetime(End) >  datetime(NEW.Start)
)

BEGIN
SELECT RAISE(FAIL, "Overlapping Intervals on Timing");
END;
```

wilsonjord (approved):
Thanks for the changes - I find the schema creation easier to read (well, as easy as dynamically constructed SQL can be), and the use of `datetime` is more rigorous.

ozym:
Once this is merged, I have some updates I'd like to do to add the response files into the database.  I may also look at any other side-loaded data that comes to mind.

elidana:
This new delta feature looks quite cool, thanks @ozym !!

A small comment from me:
would be possible to please add a README.md in `cmd/deltadb` with instructions for dummies like me on how to use it?
I found the instructions in https://github.com/GeoNet/delta/pull/2275#issue-2631051019 very useful to give it a first try

ozym:
@elidana have added a README for the deltadb command.

elidana (approved):
very very very nice , and thanks for adding the README!!!

jajera (approved):
lets give it a go

PR #2284, opened 2024-11-08
feat: add ability to list response files

ozym:
This is a set of helper functions to allow listing of the response file snippets available


elidana (approved):
looks good!

PR #2286, opened 2024-11-10
add transactions and a key value table for deltadb

ozym:
This is needed to extend initialisation of the deltadb to include other tables such as responses.

This rework shifts most of the database code away from the Table definitions, it now merely is a means of listing the various tables and their associated data.  This can then be used as needed by the actual database implementations.

This depends on #2275

### DIFF
--- a/meta/table.go
+++ b/meta/table.go
@@ -1,5 +1,9 @@
 package meta
 
+import (
+	"sort"
+)
+
 // Table holds internal settings suitable for automatic code generation.
 type Table struct {
 	name    string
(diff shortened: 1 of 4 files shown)


elidana (dismissed):

elidana (dismissed):
have been trying this locally but a bit in the dark as a sqlite newbie, so happy for this to be merged

elidana (approved):
looks good, and error when using the API when converting some null depth values is now fixed

wilsonjord:
Something I missed in my previous review, I notice
```
CREATE TABLE %s%s
INSERT INTO %s%s
```

Sqlite does not support schemas (à la `postgres`) , unless it's related to attached databases (https://www.sqlite.org/lang_createtable.html)

I assume your trying to get around this fact by "simulating" a schema, producing a concatenated schema/tablename, but I don't know how useful this is in practice - at the very least, SCHEMANAME_TABLENAME is probably more readable.

ozym:
Yeah, I wondered about that. I think I'll simplify things by removing that concept.

ozym:
@wilsonjord updated now, makes life simpler.

meta/table.go
```diff
@@ -118,78 +122,97 @@ func (t Table) Encode(list ListEncoder) [][]string {
...
-		if v == "" {
-			continue
-		}
-		res = append(res, v)
+// KeyValue contains key value pairs and their labels.
+type KeyValue struct {
+	label   string
+	desc    string
+	entries map[string]string
+}
```
wilsonjord, line 136: question: (more for my own learning) is this to enable easier creation of a Table directly from a (key-value) CSV source, do you don't need to create it's own `network.go / mark.go etc` type thing?

wilsonjord (approved):
Looks good, just one question for me out of curiosity

meta/table.go
ozym: The next PR should illustrate an example of this usage. But it's aimed more at data/information that is stored within delta but not as csv files.  So, the example is pulling in little XML "snippets" which are used to describe instrument responses without the complexity of having to model that information. These tend to end up being more akin to a key value store which I felt was a good fit for the name.

PR #2287, opened 2024-11-10
Add response table to deltadb

ozym:
This uses the key-value ad-hoc table to include response files into the database.

This depends on #2286


elidana (dismissed):
looks good, just need to check if all file changes that are currently showing here are legitimate, once https://github.com/GeoNet/delta/pull/2287/files is merged

elidana (approved):

elidana (commented):
looks good!
i can initialise the db and query station and network API from this branch

elidana (approved):

wilsonjord (approved):

PR #2294, opened 2024-11-18
feat: update consistency checks

ozym:
The checks that fail have been commented out, these will be addresses in follow up updates.

This should bring the checks closer to what's listed in the `meta/set.go` file.


wilsonjord (commented):
This PR adds more tests, reorders tests in a presumably more sensible grouping, and adds commented out **new** tests, intended for future implementation?

elidana (approved):
from what I see here all tests have been reordered and those that are commented out here were not included/run in main 

(I need to understand this better, as I thought that some of those automatic checks were running, perhaps with a different mechanism, but clearly not this consistency test modified here)

will be good to align everything and add tests for the more recent tables/data domains included in delta, thanks @ozym !!

ozym:
> This PR adds more tests, reorders tests in a presumably more sensible grouping, and adds commented out **new** tests, intended for future implementation?

In part yes, I've added and re-ordered the files to be tested.  Some were missed after updates to the code, where these now fail the tests I have commented them out.  No new tests will need to be written, but some files will need to be updated. In most cases it is simply alphabetical re-ordering. In one case it's a stray space at the end of a line, and for classes it's the order of citation names, which also need to be sorted alphabetically.

In one case, in classes.csv there's an entry for a station that doesn't exist. I don't know any more details about it but it will fail the checks unless either it's added to the stations csv file, or removed (my preferred option as this is a work in progress table).

wilsonjord (approved):

ozym:
This consistency test simply reads in a file and writes out the file and compares the two, if they don't match then it raises a flag.

PR #2296, opened 2024-11-18
feat: update doases consistency checks

ozym:
This turns on the doases file check. The only issue was the order of the entries when sorted by serial number


elidana (approved):
looks good, thanks Mark!

PR #2297, opened 2024-11-18
feat: updated channel and component consistency checks

ozym:
This comes down to sorting the two files

I have done a simple sort of the files in main, and the files here which show no differences;

```
diff channels.csv.main channels.csv.update
diff components.csv.main  components.csv.update
```


wilsonjord (commented):
is the sorting a requirement of the test, or a higher requirement? For non-timeseries, it would be an unusual requirement for input data to be sorted, no?

ozym:
> is the sorting a requirement of the test, or a higher requirement? For non-timeseries, it would be an unusual requirement for input data to be sorted, no?

This was something that was designed into the system from the beginning, although in this case it was missed.  The idea is to try and keep multiple commits a little easier to merge if there's a standard sorting in the file.

wilsonjord (approved):

elidana (approved):
looks good, thanks @ozym !

PR #2298, opened 2024-11-18
feat: update antenna and session consistency checks

ozym:
This involves sorting the files.

I've checked that the original files and the updated files still sort to the same thing (i.e. checking that there's nothing missing or double ups).

```
diff antennas.csv.main antennas.csv.update
diff sessions.csv.main sessions.csv.update
```


wilsonjord (approved):

ozym:
I had to update the files after changes were made since the original PR.

Prior to merging this will likely need another check, basically the csv files in main and in this PR should match when they are sorted.

jajera (dismissed):
lets give it a go. looks good other than the conflict.

ozym:
@aleks-spes this is something that should have been done in the beginning but was missed out for a number of tables.

The main benefit is that the order is predictable, e.g. you can easily compare two different file copies and they should either match exactly or not match at all (i.e. no more they are the same, but in a different order).

The other benefit is that when doing a git merge, if things are grouped together they tend to merge better, or at least differences tend to be close together.

PR #2301, opened 2024-11-20
feat: update site end times to match station end times

ozym:
This is based on:

```
WI227,EN,White Island Fumarole 13,-37.52188,177.189942,9,,WGS84,2018-09-13T00:00:00Z,2023-12-01T00:00:00Z
```

from stations.csv, e.g. the site shouldn't live after the station has been closed.


rumachan (approved):
Thanks for picking up the WI227 issue.

elidana (approved):

PR #2302, opened 2024-11-20
white_is doas info

rumachan:
Add info for Whakaari/White Island DOAS sensors that operated up to a little while after the 2019 eruption. Older instruments than currently operating at Ruapehu. Start/end times etc from data in FITS. Instrument data provided by Agnes.
Using station names consistent with other DOAS sites.
This will enable these data to be migrated from FITS to Tilde.


ozym (approved):
looks good to me

PR #2304, opened 2024-11-27
TRVZ install new infrasound sensor

Thomas-Benson:
Adding new Chaparral Physics M64Vx2 (413282) 1t TRVZ
Sensor is at a new location code 31
Assuming the same coordinates as the other infrasound sensor it is installed in parallel with:
-39.298816259, 175.547821602, 2062


install/streams.csv
salichon, line 1: @Thomas-Benson We re missing the LDF 1hz  i reckon as it is streamed :)

salichon (changes requested):
one more LD stream :)

salichon (approved):
sweet thanks!

CBurton90 (approved):

PR #2307, opened 2024-11-28
WLCZ fix cascadia sensor serial number

Thomas-Benson:
Change Cascadia Trillium and Cascadia Titan serial number from `3515` to `3510`


gva6500 (approved):
Thanks for fixing it 👍

salichon (approved):

PR #2315, opened 2024-12-11
MRHZ Lennartz upgrade

Thomas-Benson:
- Removing LE-3DliteMkII (F-0373) from MRHZ LC 10
 - Replacing with LE-3DliteMkIII (J-0936) at same location


salichon (approved):
nice

CBurton90 (approved):

PR #2322, opened 2025-01-15
feat: update golang to version 1.22

ozym:
This includes upgrading associated packages


ardrigh (approved):
All the deps are belong to Go

PR #2325, opened 2025-01-17
fix: adjust mount and station times

ozym:
A few small updates to allow sensors and cameras to be installed inside the operational times of the sites/mounts


rumachan (approved):
These are making changes to things I entered, probably when I was green behind the ears and made lots of mistakes. So, its good to get them sorted.

PR #2326, opened 2025-01-17
fix: adjust camera view and installation times

ozym:
This simply expands the time windows and adjusts a view code.  I notices WHOH was likely using the wrong view code based on the times.


rumachan (approved):
This is getting embarassing!

PR #2327, opened 2025-01-17
fix: update the table building

ozym:
This fixes some misconfiguration that resulted in some empty tables, it also cleans up the calling mechanism to build the database


meta/sqlite/sensor.go
```diff
@@ -125,7 +128,7 @@ CREATE TABLE IF NOT EXISTS preamp (
   preamp_id INTEGER PRIMARY KEY NOT NULL,
   site_id INTEGER NOT NULL,
   subsource TEXT NULL,
-  scale_factor REAL DEFAULT 1.0 NOT NULL,
+  scale_factor REAL NOT NULL ON CONFLICT REPLACE DEFAULT 1.0,
```
wilsonjord, line 131: question: I'm curious as to how this works; normally would expect to see this on the insert/upsert side.

ozym: sqlite seems to be a bit odd about how it handles defaults and nulls.  Normally if you want the default value you simply don't send anything through for that entry, however this becomes painful for large prepared statements (you would need one for every combination of missing data), the above coding can handle the clash of the not null setting and being sent a null value.

jajera (approved):
lets give it a try

elidana (changes requested):
i am trying to review from the height of my ignorance.
I can build the db and interrogate, but can you please update the README with the updated instructions, since some of the `./deltadb command` have now dropped?

ozym:
@elidana would be possible to merge this PR and then work  on updating the documentation. Otherwise it's likely to sit here for another six months, I need to get my head back into the space it was six months ago.

elidana:
of course!! i'll update my review

elidana (approved):
approving with closed eyes (was trying steps highlighted in https://github.com/GeoNet/delta/pull/2275 to test but they don't work and I need some sqlite food to run locally).

PR #2332, opened 2025-01-30
to accommodate data from FITS

rumachan:
**Sorry this is using a personal clone, but I had most of these edits underway before SOD moved to push direct to GeoNet, and didn't want (was scared) to try to fiddle**

Delta changes to accommodate transferring data from FITS to Tilde. Problems around not having all the required metadata, so using "unknown" in various ways.

Some small modifications of existing entries required, plus additions for older streams and datalogger/sensors not previously in Delta.
One new stream type is "overflow-flowrate" as data in FITS calculated that way, and no practical benefit in converting back to "overflow-height".
Older data for "lake-height"  was measured and then calculated relative to overflow level so does not match current data which are height above sensor. Using aspect of "relative to overflow" to indicate that in older data.


ozym (approved):
looks good, still open to updates in the future if any of the unknowns are resolved.

rumachan:
Thanks Mark. Yes, we can always fill in if we get more information.

PR #2349, opened 2025-02-14
nfc: FTPS CUSP3D to EQR120 SM recorder upgrade

Thomas-Benson:
Removing CUSP3D (42232) from FTPS LC 20
Replacing with EQR120 (46883) at same location
EQR120 Firmware version 2.4.5
Updating Streams to continuous from 2020-02-09


CBurton90 (approved):

install/recorders.csv
salichon, line 1: oreintation at 0.5 degrees precision  is  almost too much :))

salichon (approved):

PR #2351, opened 2025-02-16
Make HORZ back home

salichon:
no GeoNet sites in far pacific north yet


Thomas-Benson:
@salichon we'll need to change network/stations.csv too

salichon:
@Thomas-Benson would you bless it if your good with it ? 
thanks!

Thomas-Benson (approved):

PR #2354, opened 2025-02-17
nfc: CPWZ replace lennartz

Thomas-Benson:
- Removing Lennarz LE-3DliteMkII (K-09997) from CPWZ lc 10
 - Replacing with Lennartz Le-3DliteMkIII (F-0371) at same location


salichon (approved):
perfetto

PR #1740, opened 2023-06-07
update model name for Kinemetrics group

elidana:
Hello @ozym ,

starting now with renaming models.
Can you please check this one if ok (also in terms of response files).

/cc @CBurton90 @mnaguit

PR #2355, opened 2025-02-18
feat: stationxml generation, update Kinemetrics model names

elidana:
To continue the `stationxml` rework and name self-consistency within delta (and alignment, when possible, to the SAGE Nominal response library, this pr is renaming Kinemetrics dataloggers 

- `BASALT 8X DATALOGGER` to `Basalt/8`
- `Obsidian 4X Datalogger` to `Obsidian/4`

- @ozym can you please check this is all that needs changing? that will help with follow up pull request
- @salichon or @Thomas-Benson can you please double check that the response file looks ok? 

### Additional info
For the Kinemetrics group, remain to do ETNA and ETNA2 models (will be in follow up pull request)

replace https://github.com/GeoNet/delta/pull/1740/

To check this change, have locally build a response file and looks ok to me (by a first, coarse look).
`cd cmd/stationxml-build/; go build; ./stationxml-build -station RHBS`

```xml
       <DataLogger resourceId="Datalogger#Basalt/8X:1626">
          <Type>Strong Motion Recorder</Type>
          <Description>Kinemetrics Basalt Strong Motion Recorder</Description>
          <Manufacturer>Kinemetrics</Manufacturer>
          <Model>Basalt/8X</Model>
          <SerialNumber>1626</SerialNumber>
          <InstallationDate>2013-01-28T00:00:01</InstallationDate>
          <RemovalDate>2013-06-04T01:00:00</RemovalDate>
        </DataLogger>
        <Response>
          <InstrumentSensitivity>
            <Value>1.70934447112192e+06</Value>
            <Frequency>1</Frequency>
            <InputUnits>
              <Name>m/s**2</Name>
            </InputUnits>
            <OutputUnits>
              <Name>count</Name>
            </OutputUnits>
          </InstrumentSensitivity>
          <Stage number="1">
            <PolesZeros resourceId="PolesZeros#FBA-ES-T" name="RHBS.20.HN1.2013.030.stage_1">
```

/cc @CBurton90 and @gva6500 FYI


Thomas-Benson (approved):
LGTM! My local build of the stationxml show the name change in all the correct places, and without any impacts on the rest of the response

salichon (dismissed):

ozym:
What was the thinking behind using Basalt/8X vs using Basalt/8 ?   (for example we use Q330/3 and Q330/6 to represent the same thing).   The original names were to distinguish between the Basalts (etc) that had internal sensors vs the ones that had external sensors.

elidana:
Thanks @ozym ! No specific thinking at all, I thought the `X` was part of the "offical" model name, and I saw that used in the kinemetrics website (https://kinemetrics.com/legacy_products/basalt-basalt4x-basalt8x/) . 
Happy to use whatever is best. 

In the SAGE NRL I see a "Basalt_PG8" model...is that the same? 

Whatever name fits best to the seismo-expert, I'm happy to update this pr accordingly

ozym:
It think it's all short hand, the basic models are the same, i think that having the same style of coding is better.

If you really want to do the complete thing then -> https://service.iris.edu/irisws/nrl/docs/1/help/ and look for instconfig table.

"Basalt_PG8..." from above is referencing the gain and not the number of channels (see PG),  From a response point of view it doesn't matter how many channels it has, but from a building the response side of things it does as you need to know how to join the various sensors together especially if different channels on the datalogger have different gain settings.

elidana:
thanks @ozym , that makes sense! ok, let's go with Basalt/8 then, and I'll try to keep that in mind when reworking on the other models.

elidana:
now updated to use `Basalt/8` and `Obsidian/4' (removed trailing X)

@ozym , @Thomas-Benson can I please have another round of review?
@ozym , now that the naming bit is addressed (hopefully) , does this one looks ok to update the model name? Is there anything else that would need changing? 

If this one is complete I can then migrate another pending pr that I have from an old delta fork that I'd like to delete. Thanks!

Thomas-Benson (approved):
xml checked and everything lgtm!

ozym (commented):
looks good.

ozym (approved):
👍

PR #2358, opened 2025-02-20
nfc: CMWZ replace SM sensor + datalogger

Thomas-Benson:
- Removing Guralp Fortis (TF1114) From CMWZ LC 20
 - Removing Q330S/6 (6065) 

- Replacing datalogger with Centaur CTR4-6S (11923)
- Replacing sensor with TitanSMA (002416)


CBurton90 (approved):

salichon (approved):
sweet as thanks!

PR #2360, opened 2025-02-23
nfc: WHHZ SP sensor swap

Thomas-Benson:
Replacing Le-3DliteMkII (I-0639) at WHHZ LC 10 with Le-3DliteMkIII (K-1002) at same location


gva6500 (approved):

salichon (approved):
👍

CBurton90 (approved):

PR #2375, opened 2025-03-03
nfc: add scandoas domain and references

elidana:
Hi @rumachan ,

can you please review this? It is
- adding `gnss` domain description (as in Tilde api)
- add `scandoas` domain (was missing) and related citation

Thanks!


rumachan (changes requested):
In domains.csv "scandoas,Continuous sulphur dioxide gas emission rates and supporting data". Remove "and supporting data". Description was written when we considered including wind, quality, etc supporting data, but we decided not to, and it was never changed.

In datasets.csv, problem with "scandoas,EN,Gns2022b,true" is that EN is described as "GeoNet environmental data using low rate data collection platform network" in network/networks.csv, which isn't the case for scanDOAS. Think we need to generalise network/networks.csv to "GeoNet environmental data using various automatic low rate data collection networks"as that will include scanDOAS and any future "low rate" automatic collection system we might use.

elidana:
Thanks @rumachan !

- domain description: here I was just capturing the description used in Tilde (https://tilde.geonet.org.nz/v4/dataSummary/scandoas) - shall this one change to? My plan, if possible, would be to finalise this delta table and then manage Tilde domain description through this one (and developmnent work and perhaps some proper review of those tables is needed before this can happen)
- for the `EN` i wasn't sure either. So you confirm is correct to use it as network code for scandoas as used in delta? Or is there another one? I had a look around and was only able to find `EN` (but I might well be wrong and did big mistakes). Happy to change the network description for `EN` if that's the best option

will wait to hear back from you and then apply the changes

rumachan:
Yes, the description used in Tilde for scanDOAS needs to be changed as its just SO2 (maybe CH4 in the future), but no supporting data. I just never got round to it.

I checked also. scanDOAS stations are assigned to EN network (by me). The EN network description is quite specific (to LRDCP) but to use is for scanDOAS we need to make the description a little more general.

elidana:
@rumachan thanks for your review. I have now:
- updated scandoas description to make it consistent with what is in tilde (in the table and in the tilde configuration code)
- changed the description of the `EN` network to be more generic (and cover LRDCP and scandoas)

can you please have another look and hopefully if all looks ok approve so that it can be merged?

can someone from the @GeoNet/ssd-team-developer group please review the `scandoas.go` as well?
We only need to update the description of the `scandoas` domain, but I hope I captured all the bits where a change is required.

Thanks!

rumachan (approved):
Looks good

wilsonjord (approved):
What little code changes there is looks fine, most of the other changes look like data changes.

elidana:
thanks both!

PR #2377, opened 2025-03-04
TO006 and NA002 multigas sensor changes

rumachan:
Got a bit complicated as there had been undocumented sensor changes at NA002 and that needed to be done to facilitate the TO006 changes. Two sensors were required to be added to assets.

No logger or configuration or any other changes required.


ozym (approved):
yes, complicated.

elidana (approved):
gosh this one was complicated!!!

@rumachan lools good for what I can tell. Only question is: do the `gains.csv` file need to be updated? 
The comment in the internal ticket that triggered my question is the one about

> Yesterdays automated calibration shows new calibration values of 38.7ppm H2S (in response to 25ppm H2S cal gas) and 14.0ppm SO2 (in response to 10ppm SO2 cal gas). 

(though I guess since the sensor model is the same, the gain will be the same?)

rumachan:
Good question @elidana . No, gains file does not have to change. The regular calibration is required as these sensors tend to have a variable response, largely as they are electro-chemical and it is believed the sensors "age" due to expose to the chemical they are sensing. I would have explained a lot more about this in some forum or another but haven't been able to get the required person to focus on it yet. The calbiration is weekly, and the response to that is different every week!

PR #2384, opened 2025-03-06
nfc: HBHS Basalt to TitanSMA SM recorder swap

Thomas-Benson:
Replacing SM recorder at HBHS LC 20
 - Removing Basalt (1770)
 - Replacing with TitanSMA (002419)


salichon (approved):
sweet

CBurton90 (approved):

PR #2387, opened 2025-03-07
nfc: Alloy Firmware 6.23 upgrades - North Island Geonet

EvanCGNS:
Sites Changed: KAPT, OKOH, PGNE, MWOR, ORRG, WIAH, NLS3, NLS4


rumachan (dismissed):
Lots of back and forth checking times vs serial numb and stations. Guess that is the only way. Can't find any inconsistencies.

rumachan (approved):
Same as last time after conflicts sorted so approving

PR #2390, opened 2025-03-11
nfc: WTAZ upgrade BB BH sensor

Thomas-Benson:
Replacing borehole sensor at WTAZ LC 13
 - Removing TC120 PH2 (9205)
 - Replacing with T120-BH2 (003536)
 - Slight change in depth from 90 to 73m
 - Orientation to be confirmed at a later date


salichon (approved):
👍  sweet

gva6500 (approved):

PR #2394, opened 2025-03-13
nfc: INSS Basalt to Titan SM recorder swap

Thomas-Benson:
Removing Basalt (2157) from INSS LC 20
Replacing with TitanSMA (002454) at same location


salichon (approved):
looks good to me thanks !

CBurton90 (approved):

PR #2397, opened 2025-03-17
nfc: OTVZ add new infrasound sensor

Thomas-Benson:
Adding new infrasound sensor to OTVZ LC 31
New sensor is Chaparral M64Vx2 (852V2)
Using same coordinates as existing infrasound sensor


CBurton90 (approved):

salichon (approved):
ta

PR #2381, opened 2025-03-06
nfc: HSPZ open new site and add broadband sensor

Thomas-Benson:

PR #2401, opened 2025-03-17
nfc: IFMZ open new BB + SM site

Thomas-Benson:
Opening new site IFMZ - Mount Inframeta
Sensor: Trillium Compact 120PH-2 (10199)
Sensor (SM): Titan TACCL-N1 (3530)
Datalogger: Centaur CTR4-6S (11936)

Site coordinates: -43.023815, 171.060776, 1339m


CBurton90 (dismissed):

salichon (changes requested):
@Thomas-Benson @CBurton90 
  when we set a station BB with HH and LH is it usually  considered as a national network as opposed to regional ones?  
  OR we need consistency in streams SI regional network ... 
 see https://github.com/GeoNet/delta/pull/2381 HSPZ

install/streams.csv
```diff
@@ -1077,6 +1077,8 @@ IF07,10,E,H,100,false,false,false,2008-10-07T04:31:00Z,2009-02-11T22:00:00Z
 IF10,10,E,H,100,false,false,false,2008-10-01T03:14:00Z,2009-02-11T23:00:00Z
 IF12,10,E,H,100,false,false,false,2008-10-03T01:46:00Z,2009-02-11T23:30:00Z
 IF14,10,E,H,100,false,false,false,2008-10-08T23:31:00Z,2009-02-13T01:30:00Z
+IFMZ,10,H,H,100,false,false,false,2025-03-16T22:26:00Z,9999-01-01T00:00:00Z
```
salichon, line 1080: IFMZ HH is axial right ? (as in the data queried)
Orienation to set in mind :)

salichon (changes requested):

Thomas-Benson:
@salichon changed the stream to axial and added the LH stream I missed the first time around

CBurton90:
> @Thomas-Benson @CBurton90 when we set a station BB with HH and LH is it usually considered as a national network as opposed to regional ones? OR we would need consistency in streams SI regional network ... see #2381 HSPZ

@salichon AFAIA we always stream HH and LH no matter whether the site is a regional or national one. This ticket should indicate if the site was to be national and it does not https://github.com/GeoNet/tickets/issues/14746.

cc @Thomas-Benson @ozym

install/streams.csv
CBurton90: Yes 10m borehole = axial

salichon (approved):

ozym:
Yes, when we install a broadband we should also be configuring the 1Hz streams (this is used to calculate the long period products without the overhead of the high frequencies) . It shouldn't matter what type of site it is as otherwise why install a broadband.

PR #2406, opened 2025-03-20
SLTZ add SM sensor

Thomas-Benson:
Add SM sensor to SLTZ LC 20
 - Titan TACCL-N1 (3398) starting from `2025-03-20T20:05:00Z`


CBurton90 (dismissed):

install/connections.csv
```diff
@@ -840,6 +840,7 @@ SECF,10,Secretary Island,Short Period,,2003-08-23T04:53:00Z,2004-05-12T17:31:00Z
 SIZ,10,Horseshoe Bay,Short Period,,1990-12-14T11:46:00Z,2001-10-23T22:04:00Z
 SKTD,10,Skate,Portable,,2010-09-05T04:01:00Z,2010-09-28T23:59:00Z
 SLTZ,10,Solution Range,,,2025-02-28T12:00:00Z,9999-01-01T00:00:00Z
+SLTZ,20,Solution Range,,,2025-03-20T20:05:00Z,9999-01-01T00:00:00Z
```
salichon, line 843: @Thomas-Benson @CBurton90 
is it better if we put the pin numbers (0,3) asw we have a strong motyion and a bb on the same datalogger  ?

salichon (changes requested):
question about pins number

salichon (approved):
pin number shopudl be allrigt  - let see nex t monday :)

install/connections.csv
CBurton90: Yeah probably worth putting the SM on pin 3.

Thomas-Benson:
Added pins

salichon (approved):
in awe!

PR #2409, opened 2025-03-24
DAVS Cusp3D to EQR120 SM recorder swap

Thomas-Benson:
- Removing CUSP3D (42108) from DAVS LC 20
 - Replacing with EQR120 (469190 at same location
   - using firmware version 2.4.5
 - updating streams to continuous from `2020-02-09`


salichon (approved):
thanks all good

CBurton90 (approved):

PR #2417, opened 2025-03-28
nfc: NEZ swap BH BB sensor + datalogger + move infrasound

Thomas-Benson:
NEZ - relocation + replacement
 - Replacing datalogger Centaur CTR4-6S (11443) with Centaur CTR4-6S (9590)
  (Noting that while another datalogger + sensor were attempted here in the interim, no data was streamed and no delta updates were made)
 - Replacing TC120PH-2 (10258) with TC120PH-2 (10201)
 - Moving infrasound sensor from LC 30 to LC 31
  Estimating location of new infrasound sensor to be -39.270491, 174.096300

Also updating some connections to make the previous install more consistent. It is slightly messy since the sensor banks were swapped on 2023-02-28


CBurton90 (approved):

salichon (approved):
pfiuuu,.. all good  will need azimuth for posthole

PR #2418, opened 2025-03-30
BHW Q330 to centaur datalogger swap

Thomas-Benson:
Removing Q330S/6 (6853) from BHW | Baring Head
Replacing with Centaur (9715) at same place
Using Centaur firmware version 4.8.5


CBurton90 (approved):

salichon (approved):
sweet

PR #2422, opened 2025-03-31
nfc: PAWZ Q330 to Centaur Datalogger swap

Thomas-Benson:
Removing Q330 (6717) from PAWZ - Paruwai Farm
Replacing with Centaur (9288) at same location
Using Centaur firmware version 4.8.5


CBurton90 (approved):

salichon (approved):
super

PR #2423, opened 2025-04-01
nfc: NWFS Basalt to TitanSMA recorder swap

Thomas-Benson:
Removing Basalt (2154) from NWFS LC 20
Replacing with TitanSMA (002346) at same location
Using TitanSMA firmware version 4.9.2
Updating streams to continuous from `2020-02-09`


salichon (approved):
ty very much

PR #2426, opened 2025-04-03
manual overflow/height

rumachan:
Adding some features entries to facilitate some manualcollect Ruapehu Crater Lake data that hasn't yet been added to Tilde.


elidana (approved):
nice, thanks @rumachan , and I like that the terminology aligns with what is already in use for continous envirosensors!

PR #2427, opened 2025-04-04
Taranaki Maunga camera changes

rumachan:
Changes in front line files, that impact text on images and other files with Taranaki/Egmont. Changes only for webcam material in Delta.


elidana (approved):
changes look good, only thing is that I'm not entirely sure if this need to be changed elsewhere as well for the process that publish images to pick it up. Thanks @rumachan !

ozym (approved):
All good.

rumachan:
Mark indicated the change is now driven from Delta, so lets see if it updates, possibly overnight

PR #2429, opened 2025-04-06
fix: update sensor config

ozym:
This makes sure the camera install is within the view, which is within the mount


wilsonjord (approved):
LGTM

jajera (approved):

PR #2433, opened 2025-04-11
nfc: WKMT - Whakamoenga - New Tsunami Gauge site

EvanCGNS:
Installed: 
 - 2x Druck ptx-1830 series (SN: 12970861, 11920992)
 - Centaur CTR4-6S (SN: 11435)

Installation Timestamp: 2025-03-27T02:15:18Z

Set the timestamp for sites.csv, stations.csv, and connections.csv to the start of the day, as that's standard practice in the GNSS world. If that's not standard for TG's, let me know!


install/connections.csv
salichon, line 1: Why are  the connections  opened before the dataloggers ?
Having same Time helps to be consistent with the other instruments and the metadata 
as told about :) @EvanCGNS

install/sensors.csv
salichon, line 1: The WKMT is a lake sensor hence in  freshwater so the scale factor shoudl be 1 
as for HRMT for instance  @EvanCGNS

salichon (changes requested):

salichon:
Starting site and station on the day is  ok as long as itis before :),
though in practice  we tend to set timestamps all with the same start  times. 

only the connections time shall be set consiswtent with the instrument time stamps

salichon:
@elidana  ;) 
i ll do this would but it needs some change

install/connections.csv
EvanCGNS: Good point about the connections bit - is it ok to open the site before the others? With GPS stuff we often open the mark and the monument before the equipment at the site

ozym: From a technical point of view it is fine, it's basically the same as saying the plugs and cables have been installed prior to the actual equipment.  But, I guess, from a management point of view it may add some confusion at a later date (perhaps) but these settings tend to live past equipment changes so long as the changes are similar (e.g. the essentially the same location codes are used).

install/sensors.csv
EvanCGNS: Fixed!

EvanCGNS:
@salichon @CBurton90 @gva6500 ready for a re-review!

salichon (approved):
cool

salichon:
@CBurton90 @ozym hello!  a little reminder for that PR in Delta 
WKMT tsunami gauge in freshwater (taupo) 
I don t think we require more in delta for that one.
cheers!

PR #2437, opened 2025-05-01
feat: add location field to constituents table

ozym:
The DART analysis is evolving and is now site location specific, that means there are two sets of constituents for the same site but over different deployments. The idea is to allow the application to choose the more specific model as required, it also allows adding a Z0 offset as this is deployment specific


jajera (approved):
lgtm

meta/sqlite/gauge.go
```diff
@@ -61,11 +62,11 @@ END;
 var constituent = Table{
 	Create: constituentCreate,
 	Insert: func() string {
-		return fmt.Sprintf("INSERT INTO constituent (gauge, number, constituent, amplitude, lag, start_date, end_date) VALUES ((%s), ?, ?, ?, ?, ?, ?);",
+		return fmt.Sprintf("INSERT INTO constituent (gauge_id, location, number, constituent, amplitude, lag, start_date, end_date) VALUES ((%s), ?, ?, ?, ?, ?, ?, ?);",
 			gauge.Select(),
 		)
 	},
-	Fields: []string{"Gauge", "Identification Number", "Analysis Time Zone", "Analysis Latitude", "Analysis Longitude", "Crex Tag", "Start Date", "End Date"},
+	Fields: []string{"Gauge", "Location", "Number", "Constituent", "Amplitude", "Lag", "Start Date", "End Date"},
```
junghao, line 69: Though the new field names looks more like what we've been using, not sure where the old one applied and why changed now?

junghao (approved):
Looks good to me

salichon (approved):
sounds good to me

meta/sqlite/gauge.go
ozym: This basically gives a list of csv "column" names to load into the sqlite database when inserting from the raw files.

PR #2438, opened 2025-05-01
nfc: ORRZ open new BB BH site

Thomas-Benson:
Opening new site ORRZ | Mangere
Sensor: Trillium 20PH-1 (10782), borehole depth 295m
Datalogger: Centaur (11470), fw version 4.8.5
Site location: -36.97266193, 174.770494, 44m elevation.

Orientation tbd


Thomas-Benson:
Fixed sensor model, should be TC20PH-2 instead of TC120PH-2

CBurton90 (dismissed):

Thomas-Benson:
Moved Serial number, though it's worth noting that the assets file is not very consistent, and at some point we may just want to go through the whole file(s) and organise everything nicely.

salichon (approved):

PR #2444, opened 2025-05-13
feat: add coastal-config app

ozym:
This adds a simple application to output a json configuration file which is suitable for coastal gauge processing, much like that generated for dart processing


raki-git (approved):
LGTM

PR #2447, opened 2025-05-14
new ru001 site

rumachan:
Big new installation at Ruapehu Crater Lake, RU001.

All of this stuff is required documentation, and to get data ingested in Tilde.

PR #2449, opened 2025-05-14
Update firmware.csv

ricardojohansson:
added new CR6 now installed at RU001 with FW 12.02


elidana:
ha, thanks @ricardojohansson ! I've already  asked @rumachan to add this in https://github.com/GeoNet/delta/pull/2447

might be cleaner to directly add it there (as this one would also fail as the datalogger does not yet exist).

Let's keep this one open, but ideally the firmware entry is directly added in the big one and this pr can be closed.

Great work on RU001!!!!! really impressive

elidana:
fimrware included in https://github.com/GeoNet/delta/pull/2447, closing this as not needed (but thanks Richard!!)

elidana:
(and https://github.com/GeoNet/delta/pull/2450)

PR #2450, opened 2025-05-14
fix: add firmware and role for RU001 envirosensor

elidana:
Hi @rumachan , 

not sure what I did, but when I rebased https://github.com/GeoNet/delta/pull/2447 the additional changes got lost.
Here's a pr to fix this (use of `role` to disambiguate the two dataloggers at RU001 and inclusion of firmware info)


ricardojohansson (approved):

rumachan (approved):
That is the same as I tried to do

PR #2456, opened 2025-05-21
TO003 datalogger+sensor change

rumachan:
An old datalogger and sensor change that got missed.


ozym (approved):
looks good from what I can see.

PR #2459, opened 2025-05-28
nfc: rename guralp sensors

ozym:
This swaps to using 40T-30S, 40T-60S to use 40T/30S and 40T/60S to indicate variations on a common sensor, rather than different sensor models per se.


MCorfy (approved):
Seems reasonable to me

elidana:
replaces part of https://github.com/GeoNet/delta/pull/1741

PR #2462, opened 2025-05-30
nfc: Closing NZJ site 41

EvanCGNS:
Closure timestamp: 2025-06-01T18:00:00Z UTC (6am, Sunday 1st NZST).

NOTE: All closing times will be updated to the exact time after deployment (in the re-opening ticket). This pre-deployment merge is to block Tilde data ingestion only.


ozym (approved):
That's fine, we can come back round with actual times when we know them.  This will hopefully stop some data being processed internally.

JonoHanson:
@EvanCGNS I might be getting confused, but I think the UTC - NZT conversion isn't right.

I think 2025-06-01T18:00:00Z UTC (6am, Sunday **2nd** NZST).

EvanCGNS:
Thank you @JonoHanson!!!! You're absolutely right, the timestamp should be for the 31st UTC if we want to close it on the 1st our time, got them mixed up

PR #2465, opened 2025-05-30
fix: update gauge sqlite table

ozym:
The gauge table should reference the station directly rather than the location, the presence of the network code is related to a later join and not for the direct table building


wilsonjord (approved):
code appears to do what is on the tin in the PR description, although it's not clear to me if it's consistent with the CSVs and/or `meta` struts or not.

PR #2469, opened 2025-06-03
nfc: Closing NZL site 41

EvanCGNS:
Closure timestamp: 2025-06-04T18:00:00Z UTC (6am, Thursday 5th  NZST).


salichon (approved):
I ll log a firmware for NZL update later  as it is required  for other DART systems
 cheers
j

PR #2472, opened 2025-06-08
nfc: Closing NZK site 41

EvanCGNS:
Closure timestamp: 2025-06-08T18:00:00Z UTC (6am, Monday 9th NZST).


ozym (dismissed):

salichon (changes requested):
@EvanCGNS 
can you also consider the firmware: 
- https://github.com/GeoNet/delta/blob/5e1efb06b50700762840518d02a1eff6102cc084/install/firmware.csv#L936

so that it s consitent (and we don t forget) 
thanks

ozym (approved):
better to get it in, then fix problems later, as this will stop data that needs cleaning up from getting places it shouldn't.

PR #2475, opened 2025-06-11
nfc: KBAZ + ORRZ add 1Hz streams

Thomas-Benson:
Adding 1Hz streams for TH20 instruments, since it has been determined that these streams are still useful for these sensors.


salichon:
merged as it is a correction

PR #2476, opened 2025-06-12
nfc: Closing NZD site 41

EvanCGNS:
Closure timestamp: 2025-06-14T18:00:00Z UTC (6am, Monday 15th NZST).

Applied this change to gauges, recorders, streams, firmware, and sites files.


salichon (approved):
ta!

PR #2479, opened 2025-06-27
Fix ASBS wrong Longitude

salichon:
Aberration longitude ..
ASBS Station coordinate missed a point separator


elidana:
thanks and good catch @salichon !!!
Can you please check the coordinates though?

that's what i get if I insert them in google map
image

salichon:
@elidana okay lat was mixed up too ... 
corrected!

elidana:
nice, thanks!! 

all good now

image

PR #2481, opened 2025-06-30
APPS recorder swap for TitanSMA

salichon:
APPS 
- titanSMA 002437  start  2025Jun19th 0014 
- Sofw version is 4.9.2
- Basalt close 2025Jun18th 23:15


Thomas-Benson (changes requested):
Also need to add the TitanSMA to `install/firmware.csv`

PR #2487, opened 2025-07-02
fix: updated second version

ozym:
Due to the dual nature of naming things, there are multiple versions floating around of response files

The same as #2486 but different file.


elidana (approved):
aaah good catch!!! we need to continue the work to tidy those response file names...
but for the time being this will do, thanks @ozym !

PR #2496, opened 2025-07-09
WHOH Axis camera added

rumachan:
Three file changes necessary to setup new Axis camera at Whakatane.

@CBurton90 or @gva6500 would you please do the RIM team review as its a new site. Thank you.


rumachan:
Changed install date to be datetime of first image. Suggested by @samgeonet and also caused a consistency error nit picked up by `test.sh`.

CBurton90 (approved):

staylorofford (approved):
LGTM.

With the test.sh requirements, we could move the install times to the time of deployment as long as the view start is the same / before the install time. I am happy with going off first data in archive though.

PR #2498, opened 2025-07-09
nfc: WPRZ SP sensor swap

Thomas-Benson:
Swapping SP sensor at WPRZ - Whakapapataringa
Removing LE-3DliteMkII (I-0643)
Installing LE-3DliteMkIII (K-0998)


salichon (approved):
sweet

PR #2502, opened 2025-07-13
nfc: MAVZ close failed borehole sensor

Thomas-Benson:
Closing entries associated with failed borehole sensor on LC 10
Using a closing time of 2024-11-19T19:23:00Z
Leaving the `network/sites.csv` entry for LC 10 open, as this borehole is likely to be reoccupied


salichon (approved):
ok

PR #2503, opened 2025-07-14
fix: avoid multiple volcam images per endpoint

ozym:
Having two cameras with the same id will generate multiple update issues. This is a simple check to only process one, it relies on the order being correct as to which one it allows through.


jajera (approved):
lets give it a go

PR #2505, opened 2025-07-15
nfc: DALS CUSP3D to EQR120 Sm recorder swap

Thomas-Benson:
Removing Cusp3D (42295) from DALS LC 20
Replacing with EQR120 (46899) at same location and orientation
Using EQR120 firmware version 2.4.5
Updating streams to continuous from 2020-02-09


salichon (approved):
perfetto
ta

CBurton90 (approved):

PR #2506, opened 2025-07-17
nfc: VGPK receiver swap - NetR9 to Alloy

EvanCGNS:
files changed:
install/firmware.csv
install/receivers.csv
install/sessions.csv


elidana (approved):
looks good, thanks @EvanCGNS and sorry for taking so long

PR #2514, opened 2025-07-25
nfc: APZ SM recorder swap

Thomas-Benson:
Replacing SM recorder at APZ | The Paps
 - Removing TitanSMA (001705)
 - Replacing with TitanSMA (002425) at same location and orientation
 - Firmware version 4.9.2


salichon (approved):
perfect

PR #2517, opened 2025-07-29
nfc: PARW Receiver swap - Alloy for Alloy

EvanCGNS:
Changes to:
assets/receivers.csv
install/firmware.csv
install/receivers.csv

No changes to install/sessions.csv as PARW already had a multi-gnss Alloy at the site


elidana (approved):
looks good! 
(I did also check that previous receiver was already set to full-GNSS , so no need to also update sessions.csv file)

CBurton90 (approved):

PR #2518, opened 2025-07-30
DART instrument Response Decimation Swapped in stages 

salichon:
Adapt decimation stages for stationxml validation


salichon:
Hi there !
as a start 
something to grind on about the response of the DART 
to make them validated by the IRIS /sage/gage/earthscope  stationxml validator 
https://github.com/iris-edu/stationxml-validator
some suggestion of Mark in the related ticket :) 

@ozym @Thomas-Benson

ozym:
Okay, Plan B. Make the input sample rate for the second stage the same as tbe first (e.g. 0....) with a factor of 1.  That is, no decimation.

ozym:
You could also try -15 rather than 0.06666666666666667

salichon:
Kia Ora @Thomas-Benson @ozym 
Could we have a last check on that PR for DART REsponse consistency ?
 (no that this is used intensevely though)

PR #2519, opened 2025-07-30
nfc: HUKS Cusp3D to TitanSMA SM recorder swap

Thomas-Benson:
Replacing SM recorder at HUKS | Huiakama School
 - Removing Cusp3D (42291)
 - Replacing with TitanSMA (002440) at same location and orientation
 - Using firmware version 4.9.2


salichon (approved):
sweet thnks

PR #2520, opened 2025-07-31
nfc: WPWS Basalt to TitanSMA SM recorder swap

Thomas-Benson:
Replacing SM recorder at WPWS | Waipawa
 - Removing Basalt (2413) from LC 20
 - Replacing with TitanSMA (002547) at same location, new orientation `284`
 - Firmware version 4.9.2
 - Updating streams to continuous from `2020-02-09`


salichon (approved):
ka pai!

PR #2526, opened 2025-08-10
Close WHOH view 02 (old camera) to allow new camera (view 03) to appear on website

staylorofford:
As per https://github.com/GeoNet/tickets/issues/18459#issuecomment-3161892182

Close the old camera view.

We will need to update the view close time to the camera uninstall time when the old camera is removed.

Until then the old camera images will continue to be archived but not appear on the website.


staylorofford:
Have updated to reflect data management approach of deleting old camera files post-new camera install and renaming new camera files to view 02.

PR #2528, opened 2025-08-13
Update dataloggers.csv

ricardojohansson:
New datalogger installed at TOUEM today SN 20472


elidana (approved):
looks good. this is only updating the assets file, so the installation will need to be reflected in the other metadata tables

PR #2532, opened 2025-08-14
nfc: WVAS Basalt to TitanSMA SM recorder swap

Thomas-Benson:
Replacing SM recorder at WVAS LC 20
 - Removing Basalt (2139)
 - Replacing with TitanSMA (002546) at same location and orientation
 - Firmware version 4.9.2
 - Updating streams to continuous from `2020-02-09`


salichon (approved):
great

PR #2534, opened 2025-08-19
nfc: WHSB upgrade building array to EQR120s

Thomas-Benson:


salichon:
yo @Thomas-Benson - shall we finlaize Welli hospital that building one ?
cheers

Thomas-Benson:
I think this is consistent with what we do with other building arrays?
16.4m elevation for the station/site entries is the ground level of the building, and the depths account for the position within the building.
Putting the site elevations at zero wouldn't work unless the elevation was added to the depths, which would be less obvious for users.
the only reasonable way to simplify it would be to remove the depths and adjsut the elevations, but this would be going against the  conventions used for the other building arrays

salichon (approved):
ta

CBurton90 (dismissed):

Thomas-Benson:
Had a very painful clash with the mass EQR120 firmware updates in https://github.com/GeoNet/delta/pull/2637

salichon (approved):
pfiuuu 
thanks!
See later  for merging aye!

salichon:
kiao ra @Thomas-Benson  is it still blocked ? 
cheers
jerome

PR #2536, opened 2025-08-21
fix Markdown for fluxgate calibration table

applecuckoo:
One of the Markdown fluxgate calibration tables was previously not rendering properly - this is a quick fix for that.


ozym (approved):
Thanks for the fix.

PR #2538, opened 2025-08-25
TO006 sulphur sensor change

rumachan:
Routine change of H2S and SO2 sensors at TO006.


CBurton90 (approved):

elidana (approved):
looks good, did check serial numbers and other metadata against field logs and timestamp agree with raw data in the archive.

PR #2541, opened 2025-08-26
nfc: NTVZ replace broadband, remove old infrasound

Thomas-Benson:
Replacing Broadband sensor at NTVZ LC 10
 - Removing Guralp CMG-3ESPC (T36194)
 - Replacing with TC 120PH-2 (10290)

Removing Setra from LC 30


CBurton90 (approved):

Thomas-Benson:
Delta group has confirmed the Chaparral will remain on LC 30

salichon (approved):
NTVZ is a little twisted a bit of work of a station ))
well managed though

chapparal indeed
https://github.com/GeoNet/delta/blob/d874041b99c44e2bfe4fe3dd14db66323b249597/install/sensors.csv#L51



looks good

PR #2543, opened 2025-08-26
nfc: BSWZ SM sensor swap

Thomas-Benson:
Replacing SM sensor at BSWZ
 - Removing FBA-ES-T-ISO (3498) from LC 20
 - Replacing with TitanTACCL-N1 (3549) at same location and orientation


CBurton90 (approved):

salichon (approved):
sweet

PR #2545, opened 2025-08-27
nfc: MOVZ L4C-3D to TC120PH-2 SP to BB upgrade

Thomas-Benson:
Replacing sensor at MOVZ
 - Removing SP L4C-3D (2825) from LC 10
 - Replacing with TC120PH-2 (10201) at same location
 - Updating stream codes since this is a SP to BB swap


salichon (approved):
👍 thnks!

PR #2559, opened 2025-09-10
nfc: IFPS Cusp3D to TitanSMA Sm recorder swap main

Thomas-Benson:
Replacing SM recorder at MGCS LC 20

-     Removing Cusp3D (42169)
-     Replacing with TitanSMA (002550), with orientation 248, FW 4.9.2


rumachan (approved):
Triggered = false means continuous data, I assume. That was the only one I scratched my head over.

gva6500 (approved):
lgtm

PR #2560, opened 2025-09-10
chore: remove metadata for SM site CRGZ as no data are available

elidana:
CRGZ was a temporary site installed between 2016 and 2017 to support some science experiments in the East Cape region.

As GeoNet, we have gathered metadata of the equipment that was used, but associated data were not collected for unknown reasons. 

This pull request remove metadata entries from the GeoNet metadata database.

An extract of the StationXML file is attached here for future reference

[crgz.xml.gz](https://github.com/user-attachments/files/22246867/crgz.xml.gz)

/cc @KTJacobs  or @segburg  in case you have CRGZ data stored in the GNS temporary deployment archive

and /cc @ozym fyi


Thomas-Benson (approved):
No data for this site in the archive as far as I can tell, so LGTM

PR #2561, opened 2025-09-10
nfc: CMWS add missing streams and remove incorrect connection

Thomas-Benson:
Adding missing stream entry of CMWS
Removing 5 second connection entry on CMWZ which looks like a typo


elidana (approved):
oh nice, thanks a lot @Thomas-Benson !!

PR #2567, opened 2025-09-17
nfc: PAPS BASALT to TitanSMA SM recorder swap

Thomas-Benson:
- Removing Basalt 2158 from PAPS LC 20
- Replacing with TitanSMA (002548) at same location, orientation 357.
  -   FW version 4.9.2
- Updating streams to continuous from 2020-02-09


rumachan (approved):
Looks good

PR #2582, opened 2025-09-29
fix: add a band option for dart-config

ozym:
This fixes the problem that the config is generated for each configured band, this is the simplest fix


salichon:
Lo @wilsonjord  ! 
the duplication of NZF(41) is legit to me as the dates start/end are consistent.
NZF has some history .... got its BPR recorder/buoy system  on the loose at tha time and got it replaced 
one deployment 41 lasted 1 month only   and this is shown  in that file :)

PR #2584, opened 2025-09-30
add Trillium Posthole 120 T120-PH3 StationXML response

CBurton90:
Added from the EarthScope NRL:

https://service.iris.edu/irisws/nrl/1/combine?instconfig=sensor_Nanometrics_Trillium120PH_EG3_SG1203_LP120_STgroundVel&format=stationxml-resp

Docs reference:

image


Thomas-Benson:
The response looks good as far as I can tell, but we also need an entry in `install/components.csv` to make it usable

ozym (changes requested):
Suspect it's not 100% needed but given it's from NRL, we should probably add the top level entry, ie.

```
<Response xmlns="http://www.fdsn.org/xml/station/1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.fdsn.org/xml/station/1 http://ds.iris.edu/files/xml/station/fdsn-station-response-1.1.xsd">
```

salichon (approved):
ok thanks for the pitch @CBurton90

PR #2591, opened 2025-10-03
nfc: LOTT Tsunami sensor swap

Thomas-Benson:
Like-for-like tsunami sensor swap at LOTT | Lottin Point
 LC 40:
 - Remove Druck PTX-1830,10837198
 - Replace with Druck PTX-1830 (11982419)

LC 41:
 - Remove Druck PTX-1830,10837199
 - Replace with Druck PTX-1830 (11912321)

Question, I've left the dips at 0 the same as the previosu install, but most of the recent tsunami sensor use a dip of 90. What is the convention here currently?


salichon (approved):
thanks for catching up

PR #2665, opened 2026-01-23
nfc: AVAB Building array upgrade

Thomas-Benson:
Building array upgrade that didn't get put into delta
Upgrade looks like it occured `2021-05-17T12:18:00Z` - `2021-06-21T04:05:00Z`
The sensors at LC 24 and 2A aren't working properly, so I've skipped them for now, but it looks like those location codes might have some junk data that needs to be cleaned up


Thomas-Benson:
Updated start time for sensor 2A to 2022-03-29T20:52:00Z, since this sensor was offline/producing junk data

salichon (approved):
yay! thanks!

PR #2689, opened 2026-03-02
nfc: UPKZ add surface sensor

Thomas-Benson:
Adding surface sensor to UPKZ LC 11
 - Installing LE-3DliteMkII (I-0646)
 - Using same coordinates as the borehole site


salichon (approved):
kapai
thanks

gva6500 (approved):

PR #2725, opened 2026-04-22
nfc: Updated MKTR coordinates to match NZGD 2000

EvanCGNS:


aleks-spes (commented):
Thanks @EvanCGNS ! I believe the coordinate system should be `NZGD2000`.
Also, I think we should start using `Start` and `End` timestamps more deliberately. Could you please create a new line with updated coordinates and proper starting and closing timestamps?

aleks-spes (approved):

PR #2728, opened 2026-04-28
OT001 depth sensor replace

rumachan:
Replace depth sensor. Two files need changes.


elidana (approved):
hi @rumachan , this looks good.
Is the gain value still the same as in the current gains.csv?

```
gains.csv:OT001,03,01,,0.70325,,,2022-07-28T01:59:00Z,9999-01-01T00:00:00Z
```

I'm assuming yes, so approving this, but if not please feel free to update and ask me to review again!

PR #2734, opened 2026-04-29
nfc: MATS Basalt to TitanSMA SM recorder swap

Thomas-Benson:
Replacing SM recorder at HAVS LC 20
 - Removing BASALT (623)
 - Replacing with TitanSMA (002642) at same location
   - New orientation 0 (True North)
   - Titan FW version 4.9.2
 - Updated streams to continuous from 2020-02-09, and updated channel codes to non-axial due to the change in orientation


gva6500 (approved):

salichon (approved):
tick

PR #2739, opened 2026-05-01
KMTP dips

rumachan:
Adding dips not previously available


staylorofford (approved):
Assuming positive dips indicate up rotation from the horizontal for cameras, LGTM

PR #2748, opened 2026-05-19
gaschemistry

rumachan:
Adding required entries to `environment/features.csv` to permit working with volcanic gaschemistry data. Those data streams follow to proposed new stream naming conventions for manualcollect data.


ozym (approved):
can't comment on the actual entries but the form looks fine

PR #2749, opened 2026-05-21
feat: add multigas code

ozym:
This adds code to handle multigas calibration data


rumachan (approved):
Based on what I understand, which is mostly not the go code, it look sensible.

jajera (approved):
lets give it a go

PR #2753, opened 2026-05-27
nfc: MWOZ,NBEZ,UPKZ close temporary surface sensor installations

Thomas-Benson:
Closing temporary surface sensor installations at:
 - MWOZ.11
 - NBEZ.11
 - UPKZ.11
(All 3 sensors were installed on the secondary banks of existing dataloggers, so there is only the sensor related entries to close)


gva6500 (approved):
lgtm

salichon (approved):
nice

PR #2766, opened 2026-06-04
fix: update the external network for the XX code

ozym:
This will allow building stationxml files for temporary sites


staylorofford (commented):
Nice, and I trust you are across all the downstream implications of this.

staylorofford (approved):

PR #2755, opened 2026-05-28
nfc: NZH - close site 42

EvanCGNS:
~Closing timestamp: 2026-06-11T18:00:00Z~

_Updated_ closing timestamp: 2026-06-14T18:00:00Z

PR #2772, opened 2026-06-15
Revert "nfc: NZH - close site 42"

EvanCGNS:
Reverts GeoNet/delta#2755

PR to reclose the site will be create when we have more of an idea when work will happen at the site.


JonoHanson (approved):
LGTM

PR #2775, opened 2026-06-17
Add FOXMC Camera to delta

staylorofford:
Currently a draft PR while time of first image is being sought.


rumachan (commented):
That looks fine.

I don't know how we deal with wrong internal timing on camera. Wonder by how much. We will need to note that information somewhere, though I'm not sure where.

PR #2786, opened 2026-06-28
temporarily remove temperature sensor

rumachan:
It will later be replaced by a different sensor


elidana (approved):
looks good , nothing here that there's no data for this station in the archive between 
`2026-06-18 21:50` and `2026-06-19 01:50` UTC

PR #2792, opened 2026-07-06
nfc: FWVZ remove setra

Thomas-Benson:
Removing Setra infrasound from FWVZ LC 30
Moving Chaparral infrasound from LC 31 to the newly vacated LC 30
Closing all entries for LC 31


salichon (commented):
@Thomas-Benson 
yo
i have some seismic sensors conprised in that package - Is it legit there ?
ta !

Thomas-Benson:
It looks like it got mixed up with the last PR I made, it should be separated now though...

install/streams.csv
salichon, line 1: ~@Thomas-Benson~
~as it is a loccode swap should it be legit to add a  30 LDF stream as for the former loccode 31?~  
nope it is here

salichon (approved):
cool thks!

gva6500 (approved):

PR #2797, opened 2026-07-21
nfc: Updated DART timestamps from BPR data

EvanCGNS:
Updated location code 43 on NZF, and 42 on NZG, NZH, and NZI to match the data from the BPRs.


salichon (approved):
kapai thanks !

PR #2803, opened 2026-07-24
nfc: TWGZ LE-3Dlite to TC120PH-2 sensor swap

Thomas-Benson:
Replacing sensor at TWGZ LC 10
Removing short-period LE-3DliteMkII (I-0648)
Replacing with broadband TC120PH-2 (9284)
Updating stream codes to match new sensor


gva6500 (approved):
lgtm

salichon (approved):
ka pai

PR #2811, opened 2026-08-10
nfc: update JTGT,NTGT,SSFT,TBWT to NZGD2000 coordinates

elidana:
Those newly installed GNSS at tide gauges now have official NZGD2000 coordinates calculated by LINZ


EvanCGNS (approved):
LGTM!

PR #2816, opened 2026-08-18
nfc: PWZ Obsidian to TitanSMA SM recorder swap

Thomas-Benson:
Replacing SM recorder at PWZ LC 21
Removing Obsidian (584)
Replacing with TitanSMA (002656) at same location and orientation
 - FW version 4.9.2


salichon (approved):
sweet ta

gva6500 (approved):
👍

PR #2823, opened 2026-08-24
nfc: OPZ Obsidian to TitanSMA SM recorder swap

Thomas-Benson:
Replacing SM recorder at OPZ LC 21
Removing Obsidian (619)
Replacing with TitanSMA (002654) at same location and orientation
-  FW version 4.9.2
-  Updated streams to continuous from 2020-02-09


salichon (commented):
@Thomas-Benson 
- We might need to work on the former  AND the new  recorder orientation   from 14 deg (?) to the new  289deg  

- this also  implies in that case  the we _may_ reconsider the naming from HN ENZ to 12Z ?  say at least from ....  **!! 2017 !! Loads of files** ....  (~9x2*365 ~10 k files) 
Does it worth it - in addition to check for the years before .... ? 
the components standards allow ENZ +/- 5deg.

-  In any case we streams,.csv need to be set to "true" for the axial info now. .

salichon (approved):
\good thx!

gva6500 (approved):

install/streams.csv
```diff
@@ -1953,7 +1953,9 @@ OPZ,10,H,H,100,false,false,false,2008-06-27T01:30:00Z,9999-01-01T00:00:00Z
 OPZ,20,B,N,50,false,false,false,2008-06-27T01:30:00Z,2017-06-26T02:00:00Z
 OPZ,20,H,N,200,false,false,true,2008-06-27T01:30:00Z,2017-06-26T02:00:00Z
 OPZ,21,B,N,50,false,false,false,2017-06-26T03:00:00Z,2020-06-16T00:00:00Z
-OPZ,21,H,N,200,false,false,true,2017-06-26T03:00:00Z,9999-01-01T00:00:00Z
+OPZ,21,H,N,200,false,false,true,2017-06-26T03:00:00Z,2020-02-08T23:59:59Z
+OPZ,21,H,N,200,false,false,false,2020-02-09T00:00:00Z,2026-08-20T01:04:00Z
```
salichon, line 1958: Axial Question to be worked later

salichon:
Merging as @Thomas-Benson is   having fun in the Waiotapu sulfur steam

PR #2826, opened 2026-08-31
close FOXKP datalogger

rumachan:
Closing datalogger as this was missed when site was destroyed.


elidana (approved):
oh nice catch! all other metadata have a close timestamp already
thanks @rumachan !

PR #2828, opened 2026-09-02
update multigas calibration details

rumachan:
Change from 400 to 600 ppm calibration for CO2, at RU001 and TO003


elidana (approved):
looks good, thanks @rumachan !

PR #2836, opened 2026-09-10
feat: update volcano names

bpeng:
- update volcano names: `Whakaari/White Island`, `Taranaki Maunga`, `Rangitāhua/Kermadec Islands`,`Rangitāhua/Raoul Island`


ozym:
@bpeng this isn't going to do what you think it does.  This merely associates what's in the camera view table entries with a config for building captions.  The caption comes from the this delta table, however having found a view it will update the config to attach this to a volcano (which is defined here).  So what is likely to happen is the camera doesn't get configured so no images will be captioned for those cameras.

ozym:
e.g. you also need to update the "network/views.csv" to update the caption, and once that is updated then this code can match that to an internal list of keywords which will then attach it to a volcano.  The volcano itself isn't configured in delta anywhere so it needs to do the linking here.

sue-h-gns (dismissed):
Need to get Steve's approval

ozym:
If this gets merged then camera images for Raoul will stop being updated.

ozym:
The tests don't need to change, they are just tests.  They don't need to be geographically accurate, this will lessen the size of these changes.

rumachan (commented):
I don't pretend to understand how everything gets sorted behind the scenes, and defer to Mark and SysDev for that. In the past I have made only changes required to adjust the label on webcam images on the website, e.g. White Island became Whakaari/White Island, and a couple of years ago Taranaki became Taranaki Maunga. That change is in network/views.csv. The required change for Raoul Island is missing from this PR `RIMK,01,Raoul Island,0,,0,Images looking into Green Lake on Rangitāhua/Raoul Island from the volcano camera situated on Mount Moumoukai.,2009-05-18T00:00:02Z,9999-01-01T00:00:00Z`

JonoHanson:
please hold for now.

elidana:
@bpeng ,
after talking with @rumachan and @ozym , I have created a new pull request to update the watermark on camera images on the GeoNet website: https://github.com/GeoNet/delta/pull/2842.
Hopefully it will be sufficient to cover what you were hoping to do with this one. If not, we can address other changes bit by bit after checking the wider implications on other downstream processes.
if you're happy with that, we can close this one and capture any remaining work separately?

bpeng:
Close as it is replaced by another PR

PR #2848, opened 2026-09-18
nfc: TMDS Cusp3D to EQR120 SM recorder swap

Thomas-Benson:
Upgrading SM recorder at TMLDS LC 20
 - Removing Cusp3D (42231)
 - Replacing with EQR120 (47081) at same location and orientation
   - FW version 2.4.5


gva6500 (approved):

salichon (approved):
danke

